PR #3, opened 2019-01-15
Specify network ID in first position of byte array

IkerAlus:
A minor update to: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0009.md

### DIFF
--- a/proposals/lip-0009.md
+++ b/proposals/lip-0009.md
@@ -55,7 +55,7 @@ Note that the sum symbol “+” in the formula above stands for the string conc
 
 It is important to point out that the version and community identifier fields should be well documented and defined in the Lisk ecosystem (for example as in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0009.md
```diff
@@ -55,7 +55,7 @@ Note that the sum symbol “+” in the formula above stands for the string conc
 
 It is important to point out that the version and community identifier fields should be well documented and defined in the Lisk ecosystem (for example as in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).
 
-In order to implement the proposed change in the Lisk ecosystem (Lisk Core, Lisk Elements), the logic in the following functions needs to be updated to include the network identifier:
+In order to implement the proposed change in the Lisk ecosystem (Lisk Core, Lisk Elements), the network identifier has to be included in first position of the byte array used as the input of the transaction signing process. This will impact the following functions according to the current implementation:
```
janhack: The getBytes function is used both for signing and generating the transaction ID. Therefore, I think it would be better to make clear that the byte array for generating transactions remains unchanged and does not include the network identifier.

IkerAlus: added clarification in the new commit below

PR #17, opened 2019-05-28
Add clarification to LIP and sample code

janhack:
- Clarify the definition of the active set of delegates in the first 2 rounds.
- Add clarification that forging nodes must not forge while changing to a different chain.
- Change of terminology from "conflicting blockheaders" to "contradicting blockheaders" to avoid confusion with the term "conflicting blocks".

### DIFF
--- a/proposals/lip-0014.md
+++ b/proposals/lip-0014.md
@@ -104,7 +104,7 @@ The Lisk-BFT protocol is formally defined and analyzed in [8]. It defines that b
 The whole section is organized as follows:
 
 *   The section "Additional Block Header Properties" defines the new properties `heightPrevious` and `heightPrevoted` for the block headers and the section "Validating Additional Block Header Properties" describes how these are validated.
-*   The section "Detecting Conflicting Block Headers" describes the conditions for when block headers are conflicting and the corresponding delegate violated the protocol.
+*   The section "Detecting Contradicting Block Headers" describes the conditions for when block headers are contradicting and the corresponding delegate violated the protocol.
 *   The section "Computing Prevotes and Precommits" describes how the `prevotes` and `precommits` implied by a block header can be computed.
 *   The formal specification of the different cases in the diagram above is given in the section "Applying Blocks According to Fork Choice Rule".
 *   The section "Timing of Block Forging" describes how the timing of block forging needs to be adjusted to be in line with the fork choice rule.
@@ -135,9 +135,9 @@ For a new child block `B`, which is supposed to be added to the current chain co
 
 In a separate LIP, we further propose a Proof-of-Misbehaviour transaction that can contain such an invalid block header as evidence for a violation of the protocol rules. This would allow to punish a delegate for providing incorrect information.
 
-### Detecting Conflicting Block Headers
+### Detecting Contradicting Block Headers
 
-In order to enforce the Lisk-BFT protocol rules, we also need to detect if a set of valid block headers by the same delegate violate the protocol rules. The properties that the valid block headers by one delegate need to satisfy are outlined in [8, Lemma 4.2]. We now present a function that checks these properties. We assume that every block header contains the following properties:
+In order to ensure that the implied `prevotes` and `precommits` satisfy the consensus framework [8, Definition 3.1], we also need to detect if a set of valid block headers by the same delegate are contradicting. The properties that the valid block headers by one delegate need to satisfy are outlined in [8, Lemma 4.2]. We now present a function that checks these properties. We assume that every block header contains the following properties:
 
 *   `height`: the height associated with the block
 *   `heightPrevious`: largest height of any block previously forged, as described above
@@ -149,7 +149,7 @@ The following code checks whether there is a conflict between two block headers,
 a violation of the Lisk-BFT protocol, by using the properties shown in [8, Lemma 4.2]:
 
 ```js
-checkHeadersConflicting(blockHeader1,blockHeader2) {
+checkHeadersContradicting(blockHeader1,blockHeader2) {
    // Order the two block headers such that b1 must be forged first
    let b1=blockHeader1;
    let b2=blockHeader2;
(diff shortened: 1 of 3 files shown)


proposals/lip-0014.md
```diff
@@ -471,6 +477,9 @@ Assume the mechanism described in this section is triggered according to the con
 (A.height<B.height && A.heightPrevoted==B.heightPrevoted)
 ```
 
+It is again essential that a node does not forge any blocks while performing the following steps for switching to a different chain. Otherwise, the node may forge a block contradicting the fork choice rule. This means that forging must be deactivated before starting the following steps.
+After exiting the "Block Synchronization Mechanism" (because it is aborted or successfully completed) the block at the tip of the chain satisfies the condition above and forging can be activated again.
```
michielmulders: It mentions `Block Sync Mechanism` but this is also valid for `Fast Chain Switching`, right? Whenever the FCS finishes successfully, forging is activated again. If it fails, I assume `block synchronization` in.

proposals/lip-0014.md
```diff
@@ -471,6 +477,9 @@ Assume the mechanism described in this section is triggered according to the con
 (A.height<B.height && A.heightPrevoted==B.heightPrevoted)
 ```
 
+It is again essential that a node does not forge any blocks while performing the following steps for switching to a different chain. Otherwise, the node may forge a block contradicting the fork choice rule. This means that forging must be deactivated before starting the following steps.
```
IkerAlus: Is this `forging off` requirement not better in place as soon as Case 5 occurred and moving to a different chain starts? Perhaps it is clearer to set it as part of the 3 steps of that subsection. Also, this way we avoid repeating this requirement twice in this and previous section and maybe just a short reminder here is enough.

proposals/lip-0014.md
```diff
@@ -149,7 +149,7 @@ The following code checks whether there is a conflict between two block headers,
 a violation of the Lisk-BFT protocol, by using the properties shown in [8, Lemma 4.2]:
```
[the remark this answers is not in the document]
janhack: Yep, I will fix it.

proposals/lip-0014.md
```diff
@@ -135,9 +135,9 @@ For a new child block `B`, which is supposed to be added to the current chain co
 
 In a separate LIP, we further propose a Proof-of-Misbehaviour transaction that can contain such an invalid block header as evidence for a violation of the protocol rules. This would allow to punish a delegate for providing incorrect information.
 
-### Detecting Conflicting Block Headers
+### Detecting Contradicting Block Headers
 
-In order to enforce the Lisk-BFT protocol rules, we also need to detect if a set of valid block headers by the same delegate violate the protocol rules. The properties that the valid block headers by one delegate need to satisfy are outlined in [8, Lemma 4.2]. We now present a function that checks these properties. We assume that every block header contains the following properties:
+In order to ensure that the implied `prevotes` and `precommits` satisfy the consensus framework [8, Definition 3.1], we also need to detect if a set of valid block headers by the same delegate are contradicting. The properties that the valid block headers by one delegate need to satisfy are outlined in [8, Lemma 4.2]. We now present a function that checks these properties. We assume that every block header contains the following properties:
```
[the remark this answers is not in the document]
janhack: I will add ` and the respective delegate follows the fork choice rule`.

proposals/lip-0014.md
```diff
@@ -364,7 +365,7 @@ If any of the validations fail, we discard the block. We further ban the peer se
 **2. Step Perform Lisk-BFT related validations and verifications**
 
 *   Check if the `B.heightPrevoted` value in the block header is correct. This is outlined in the section "Validating Additional Block Header Properties" above.
-*   Check if there is any conflict between the block header and any block by the same delegate in the current chain. For performance reasons, for a block `B` that is supposed to be added to the tip of the chain, we only check the blocks at heights `B.height-303` up to `B.height-1`. This verification step is formally defined in the section "Detecting Conflicting  Block Headers" above.
+*   Check if there is any contradiction between the block header and any block by the same delegate in the current chain. For performance reasons, for a block `B` that is supposed to be added to the tip of the chain, we only check the blocks at heights `B.height-303` up to `B.height-1`. This verification step is formally defined in the section "Detecting Contradicting Block Headers" above.
```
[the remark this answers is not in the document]
janhack: I mentioned it and added the reference to the paper.

proposals/lip-0014.md
janhack: I initially thought so, too. But then I intentionally put it after all the checks in the section `Moving to a Different Chain` are done so that forging does not stop because a delegate receives a block with incorrect information from a peer (e.g.,  invalid `heightPrevoted`).

proposals/lip-0014.md
janhack: An analogous sentence is contained in the `Fast Chain Switching Mechanis` section. I don't understand the last sentence.

proposals/lip-0014/LiskBFT.js
```diff
@@ -177,14 +177,14 @@ class LiskBFT {
   }
 
   /**
-   * Check whether blockheader b is conflicting with any block at height [maxHeightStored-heightOffset, maxHeightStored].
+   * Check whether blockheader b is contradicting any block at height [maxHeightStored-heightOffset, maxHeightStored].
```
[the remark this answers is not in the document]
janhack: Additionally mentioned this.

proposals/lip-0014.md
```diff
@@ -366,6 +366,8 @@ If any of the validations fail, we discard the block. We further ban the peer se
 
 *   Check if the `B.heightPrevoted` value in the block header is correct. This is outlined in the section "Validating Additional Block Header Properties" above.
 *   Check if there is any contradiction between the block header and any block by the same delegate in the current chain. For performance reasons, for a block `B` that is supposed to be added to the tip of the chain, we only check the blocks at heights `B.height-303` up to `B.height-1`. This verification step is formally defined in the section "Detecting Contradicting Block Headers" above.
+Note that if there is no block by the same delegate at heights `B.height-303` up to `B.height-1`, there is no contradiction.
```
AndreasKendziorra: As discussed, the formulation `there is not contradiction` should be changed to something like "there is no contraction with any non-finalized block"

janhack: Changed to  `there is no contradiction with any block at these heights.`.

PR #24, opened 2019-09-23
Add LIP for address system using base32 encoding of long hash of pubKey plus checksum

AndreasKendziorra:


proposals/use_base32_encoding_of_long_hash_of_pubkey_plus_checksum_for_address.md
```diff
@@ -0,0 +1,494 @@
...
+
+##### Reclaim Transaction
+
+###### Type
+
+The type of the reclaim transaction is the lowest positive integer that is not used for any other transaction type on the Lisk blockchain (8 at the time of writing).
+
+###### Schema
+
+The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _string_. The value of `asset.amount` must represent a decimal number and this number must be equal to the balance associated to the address of the current format in Beddows.
```
MaximeGagnebin: I find it a bit weird to specify that "amount" must be of type _string_. Also, in the serialisation below, we say that it should be serialised as a number. Maybe this is an implementation detail?
@shuse2 could you give your take on this?

PR #34, opened 2019-10-28
Add clarifications and minor specification changes to LIP 0014

janhack:
This PR add some clarifications to LIP 0014 and contains minor specification changes as a result of the implementation work on this LIP.


proposals/lip-0014.md
```diff
@@ -506,9 +509,9 @@ The node chooses one peer `P` at random from the set `bestPeers` and request its
 
 **3. Step: Revert to last common block**
 
-  a)  The node requests the last common block `C` from `P`.
+  a)  The node requests the last common block `C` from `P`. If `P` does not provide a common block `C` after several tries, then `P` is banned and the block synchronization is restarted from Step 1.
```
IkerAlus: Should the number of tries be specified? or is it up to the node operator to choose the value?

MaximeGagnebin: Good point. Our implementation will definitely have a set number for the amount of tries, but it would not break the protocol to use another number. Maybe we can make a non-mandatory recommandation ?

janhack: I changed `several` to `three`, but this value can be changed in the implementation (if three is to harsh, for instance).

janhack:
I also removed the Javascript prototype code. We can still use it for the specification tests, but in the LIP repository it does not add much value as all relevant parts are already specified as code in the LIP.

janhack:
I added the information that a peer is banned if it provides an invalid block in the Fast Chain Switching Mechanism.

PR #38, opened 2020-01-15
Update Multisignature LIP

MaximeGagnebin:
Update to multisignature concept:
- the concept of "second sig accounts" is removed from the Lisk protocol. The same security can be obtained by making the converting the account to a multisignature account with 2 required keys. Therefore, we will disable the "second signature registration" transaction.
- the way `signatures` is stored and verified is improved.
- a multisignature account can now register just 1 public key (previous minimum was 2).

### DIFF
--- a/proposals/lip-0017.md
+++ b/proposals/lip-0017.md
@@ -190,12 +190,12 @@ The asset property of a Multisignature Registration Transaction has to be serial
 ```
 serialize_asset_for_multisignature_registration(asset):
    let byteBuffer be an empty byte buffer
-   let lm be the length of the array asset.mandatoryKeys
-   append lm as an 8-bit unsigned integer to byteBuffer
+   let numMandatoryKeys be the length of the array asset.mandatoryKeys
+   append numMandatoryKeys as an 8-bit unsigned integer to byteBuffer
    for each key in asset.mandatoryKeys:
        append the big endian encoding of key to byteBuffer
-   let lo be the length of the array asset.optionalKeys
-   append lo as an 8-bit unsigned integer to byteBuffer
+   let numOptionalKeys be the length of the array asset.optionalKeys
+   append numOptionalKeys as an 8-bit unsigned integer to byteBuffer
    for each key in asset.optionalKeys:
        append the big endian encoding of key to byteBuffer
    append asset.numberOfSignatures as an 8-bit unsigned integer to byteBuffer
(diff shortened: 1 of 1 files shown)


proposals/lip-0017.md
```diff
@@ -89,105 +93,133 @@ There are, however, still some limitations that disallow arbitrarily large numbe
...
-    - For every public key in `mandatoryKeys` or `optionalKeys,` there must be an entry with a valid signature and matching index.
-    - If the sender public key is also contained in one of the two arrays, then there must be a index-signature entry for this public key as well.
+   - The value must be an integer between 1 and the total number of public keys of the account (sum of optional and mandatory keys).
+   - The value must not be smaller than the length of the value of `mandatoryKeys`.
+   - The value specifies how many signatures are needed for every outgoing transaction from the registered multisignature account, where signatures for mandatory and optional keys are counted.
+5. The `signatures` property of the transaction object is appended with signatures corresponding to the public keys contained in `mandatoryKeys` and `optionalKeys`. All public keys must have a corresponding signature. The signatures corresponding to `mandatoryKeys` are appended first, followed by those corresponding to `optionalKeys`, where the order of signatures is the same as the order of public keys in the respective array.
```
pablitovicente: In the case of `optionalKeys` the order might be the same but gaps can occur so it would not be possible to pair signatures to keys. For example if we have as `optionalKeys` -> `{ a, b, c }` a transaction where with signatures from `{a, c}` - signatures { 0xab..., 0xcd...} - might be sent and they might be ordered but how do we check that `0xcd` corresponds to key `c` instead of `b`?

proposals/lip-0017.md
```diff
@@ -89,105 +93,133 @@ There are, however, still some limitations that disallow arbitrarily large numbe
...
-        append the big endian encoding of key to byteBuffer
-    let lo be the length of the array asset.optionalKeys
-    append lo as an 8-bit unsigned integer to byteBuffer
-    for each key in asset.optionalKeys:
-        append the big endian encoding of key to byteBuffer
-    append asset.numberOfSignatures as an 8-bit unsigned integer to byteBuffer
-    return byteBuffer
+serialize_asset_for_multisignature_registration(asset):
+   let byteBuffer be an empty byte buffer
+   let lm be the length of the array asset.mandatoryKeys
```
pablitovicente: as we're using some variables in this pseudo code with full words I would do the same for `lm` and `lo` that probably mean `lengthMandatory` and `lengthOptional`

proposals/lip-0017.md
AndreasKendziorra: For the registration transaction, there needs to be a signature for every public key. So, no gaps in this case.

proposals/lip-0017.md
MaximeGagnebin: changed

PR #42, opened 2020-01-30
Remove the specification that the blockID has to be validated

AndreasKendziorra:
The `id` property is a non-required property. Therefore, it should not be validated during block validation. Instead, the blockID should be computed (and overridden if already existing) during block validation, as it is currently done.

### DIFF
--- a/proposals/lip-0020.md
+++ b/proposals/lip-0020.md
@@ -47,7 +47,7 @@ Although the success probability is very low, we want to increase the resistance
 
 ## Backwards Compatibility
 
-The change introduces a hard fork, because of the following: Blocks forged by nodes following the proposed protocol get rejected by nodes following the current protocol and vice versa since the format of the `id` and the `previousBlock` property change.
+The change introduces a hard fork, because of the following: Blocks forged by nodes following the proposed protocol get rejected by nodes following the current protocol and vice versa since the format of the `previousBlock` property changes.
 
 ## Reference Implementation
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0020.md
```diff
@@ -35,11 +35,7 @@ If `H` is the height from which on the proposed protocol is used, then the block
 
 ### BlockIDs in JSON Objects
 
-In a JSON object representing a block, all properties that contain a blockID have to be represented as a hexadecimal string. The properties of a block that contain a blockID are the `previousBlock` and the `id` property.
-
-### BlockID Verification
-
-During the verification of a block, the blockID must be verified as correct. If the blockID verification fails, the block has to be rejected. If `H` is the height from which on the proposed protocol is used, then a block `B` with `B.height < H` must have a valid blockID according to the current format (such a verification may happen, for example, during synchronisation).
+In a JSON object representing a block, all properties that contain a blockID have to be represented as a hexadecimal string. The properties of a block that contain a blockID are the `previousBlock` and the `id` property, whereby only the `previousBlock` property is required.
```
ricott1: I would not even mention `id` as a property of a block JSON object. Maybe something like: 
```suggestion
In a JSON object representing a block, the property `previousBlock`, containing a blockID, has to be represented as a hexadecimal string.
```

Also, going through the LIP I noticed a possible typo:
```suggestion
In the proposed protocol, the blockID of a transaction B is computed as follows:
```

Finally, I would change the very last sentence in the backwards compatibility section:
```suggestion
...since the format of the `previousBlock` property change.
```

AndreasKendziorra: all changes applied

PR #44, opened 2020-02-11
Update LIP-0003

IkerAlus:
- Use delegate address instead of delegate public key
- Add the dependency of LIP-0022 to generate the round's unique identifier
- Update functions and links
- Re-organize content


proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
 
-- The delegate's unique identifier can be the public key or the ID of the delegate. Arguably, it is better to use the public key since the ID system might be changed in the future. However, this is something to be discussed for the reference implementation. Both options are equal in terms of generating randomness.
-- The round's unique identifier can be the height of the block in which the new round starts or the height of the round itself. As an example, it can be calculated in the same way it is calculated currently, using the height of the round:
+- The delegate's unique identifier is the binary address of the delegate. Note that this proposal does not depend on the implementation of [LIP-0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md). The only requirement here is that the addresses have to be unique for each forging delegate which is ensured by the protocol in any case.
+- The round's unique identifier is one of the two random seed generated by the specifications in [LIP-0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation).
```
janhack:
```suggestion
- The round's unique identifier is one of the two random seed generated according to the specifications in [LIP 0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation).
```

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
-2. One hash is generated per delegate using `seedSource` and the delegate public key (or ID). For simplicity, we propose to use the hashing function already utilised in the existing code, although the final choice here is left to the reference implementation. Example given in pseudo-code:
+2. One hash is generated per delegate using `seedSource` and the delegate address. For simplicity, we propose to use SHA-256, which is the hashing function already utilised in the existing function. Following the current implementation, an example in pseudo-code will be:
 
 	```
-	forEach delegate
-		delegate.pubKeyHash = createHash(seedSource + delegate.publicKey)
+	forEach delegate in delegateList
+		delegate.AddressHash = hash(seedSource + delegate.address)
 	```
+	Note that `+` stands for the string concantenation operator.
```
janhack: From the rationale, I assume `delegate.address` is a binary value, a Buffer object in JavaScript (8 bytes for the current address system, 20 bytes for the proposed address system). So it should be rather concatenation of bytes here, right?

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
+The changes affect the `shuffleDelegateListForRound` function mentioned above. These changes consist of removing the nested `for` loops and generating a new list of “delegate hashes” to be re-calculated and ordered every round. In particular, the proposed fix is as follows:
 
-1. The variable `seedSource` is generated in the same way as now:
+1. The variable `seedSource` is generated as follows:
 
 	```js
-	var seedSource = modules.rounds.calc(height).toString();
+	const seedSource = randomSeed1;
 	```
+	where `randomSeed1` is the variable specified in [LIP-0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation).
```
janhack:
```suggestion
	where `randomSeed1` is computed as specified in [LIP 0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation). 
```

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
+By having a unique hash per delegate for every round, we will ensure that in every round the list of delegates will be re-ordered based on a uniform distribution. Moreover, these changes not only fix the uniformity issue (caused by the two issues raised in the [Motivation](#motivation) section), but it will also allow simpler and more human-readable code in any implementation.
 
 ## Specification
 
-The changes will only affect the `Delegates.prototype.generateDelegateList` function mentioned above. These changes consist of removing the nested `for` loops and generating a new list of “delegate hashes” to be re-calculated and ordered every round. In particular, the proposed fix implements a new algorithm that goes as follows:
+The changes affect the `shuffleDelegateListForRound` function mentioned above. These changes consist of removing the nested `for` loops and generating a new list of “delegate hashes” to be re-calculated and ordered every round. In particular, the proposed fix is as follows:
 
-1. The variable `seedSource` is generated in the same way as now:
+1. The variable `seedSource` is generated as follows:
```
janhack:
```suggestion
1. The variable `seedSource` is defined as follows:
```

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
 
-3. The delegates are reordered according to the hashes generated in step 2. This can be done according to any criteria (in an ascending order, for example) as long as the list is sorted in a deterministic and consistent way. Hence, the sorting algorithm must be [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) to generate the same list of delegates on every node on the network, even in the unlikely event of a hash collision in the second point. As an example, the `sort()` method in JavaScript can be used for this purpose.
-
-This will not only fix the uniformity issue (caused by the two issues raised in the [Motivation](#motivation) section), but it will also reduce complexity, allowing simpler, more human-readable code in any implementation.
+3. The delegates are reordered in lexicographical order according to the hashes generated in step 2. This sorting must be deterministic and consistent. This means it has to be [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) to generate the same list of delegates on every node on the network, even in the unlikely event of a hash collision in the second point.
```
janhack: We would need to define the order of the delegate list before calling the sorting *stable* algorithm so that tie-breaking in case of a hash collision is defined. Originally, the delegate list was always given by rank. With standby delegates, we may want to assume lexicographic order by address before the sorting.

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
+	where `randomSeed1` is the variable specified in [LIP-0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation). 
 
-2. One hash is generated per delegate using `seedSource` and the delegate public key (or ID). For simplicity, we propose to use the hashing function already utilised in the existing code, although the final choice here is left to the reference implementation. Example given in pseudo-code:
+2. One hash is generated per delegate using `seedSource` and the delegate address. For simplicity, we propose to use SHA-256, which is the hashing function already utilised in the existing function. Following the current implementation, an example in pseudo-code will be:
 
 	```
-	forEach delegate
-		delegate.pubKeyHash = createHash(seedSource + delegate.publicKey)
+	forEach delegate in delegateList
+		delegate.AddressHash = hash(seedSource + delegate.address)
```
janhack: Maybe `delegate.roundHash` would be a more suitable name, as the hash does not only depend on the address. But just a suggestion.

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
 
 The proposed fix is based on hashing each delegate’s unique identifier together with a unique identifier for the round:
 
-- The delegate's unique identifier can be the public key or the ID of the delegate. Arguably, it is better to use the public key since the ID system might be changed in the future. However, this is something to be discussed for the reference implementation. Both options are equal in terms of generating randomness.
-- The round's unique identifier can be the height of the block in which the new round starts or the height of the round itself. As an example, it can be calculated in the same way it is calculated currently, using the height of the round:
+- The delegate's unique identifier is the binary address of the delegate. Note that this proposal does not depend on the implementation of [LIP-0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md). The only requirement here is that the addresses have to be unique for each forging delegate which is ensured by the protocol in any case.
```
MaximeGagnebin:
```suggestion
- The delegate's unique identifier is the binary address of the delegate. Note that this proposal does not depend on the implementation of [LIP 0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md). The only requirement here is that the addresses have to be unique for each forging delegate which is ensured by the protocol in any case.
```

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
+1. The variable `seedSource` is generated as follows:
 
 	```js
-	var seedSource = modules.rounds.calc(height).toString();
+	const seedSource = randomSeed1;
 	```
+	where `randomSeed1` is the variable specified in [LIP-0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation). 
 
-2. One hash is generated per delegate using `seedSource` and the delegate public key (or ID). For simplicity, we propose to use the hashing function already utilised in the existing code, although the final choice here is left to the reference implementation. Example given in pseudo-code:
+2. One hash is generated per delegate using `seedSource` and the delegate address. For simplicity, we propose to use SHA-256, which is the hashing function already utilised in the existing function. Following the current implementation, an example in pseudo-code will be:
```
ricott1:
```suggestion
2. One hash per delegate is generated using `seedSource` and the delegate address. For simplicity, we propose to use SHA-256, which is the hashing function already utilised in the existing function. Following the current implementation, an example in pseudo-code will be:
```

proposals/lip-0003.md
ricott1: also I would not capitalize the variable name `AddressHash --> addressHash`

proposals/lip-0003.md
ricott1: We could use | then (its what I use in the pseudocode in the Merkle tree LIPs

proposals/lip-0003.md
ricott1: but then wouldn't we be assuming no collision in the address (which would imply collision in the hash if I understand correctly)? So that would not solve the problem. Thinking about it, the problem is not hash collision of 32 bytes, but of 20 bytes (maybe this was clear)

ricott1: disregard my previous comments, addresses are unique identifiers

proposals/lip-0003.md
IkerAlus: changed to Jan's suggestion

proposals/lip-0003.md
IkerAlus: change to Alessandro's notation

proposals/lip-0003.md
IkerAlus: I think it is an overkill to impose a certain order in the input or to do an extra sorting before the actual ordering for this reason. Here we would need to have a hash collision AND a different input list order in the nodes (in practice this input list should be the same for every node if they run our software). I think we are safe by just assuming that SHA-256 is collision resistant (as it is done normally).  Therefore I am going to remove the last sentences of this bullet point, that by the way they are more rationale than specification.

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
-```
 
-By having a unique hash per delegate for every round, we will ensure that in every round the list of delegates will be re-ordered based on a uniform distribution.
+By having a unique hash per delegate for every round, we will ensure that in every round the list of delegates will be re-ordered based on a uniform distribution. Moreover, these changes not only fix the uniformity issue (caused by the two issues raised in the [Motivation](#motivation) section), but it will also allow simpler and more human-readable code in any implementation.
 
 ## Specification
 
-The changes will only affect the `Delegates.prototype.generateDelegateList` function mentioned above. These changes consist of removing the nested `for` loops and generating a new list of “delegate hashes” to be re-calculated and ordered every round. In particular, the proposed fix implements a new algorithm that goes as follows:
+The changes affect the `shuffleDelegateListForRound` function mentioned above. These changes consist of removing the nested `for` loops and generating a new list of “delegate hashes” to be re-calculated and ordered every round. In particular, the proposed fix is as follows:
```
MaximeGagnebin: This is maybe a bit to "implementation" specific. I liked having the reference to the functions in the rationale/motivation but I think the Specifications should not really mention the function name, or the `for` loops. Maybe something more generic: "This section specifies how the delegate set has to be ordered for each round."

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
 
-1. The variable `seedSource` is generated in the same way as now:
+1. The variable `seedSource` is defined as follows:
 
 	```js
-	var seedSource = modules.rounds.calc(height).toString();
+	const seedSource = randomSeed1;
 	```
+	where `randomSeed1` is computed as specified in [LIP 0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation).
```
MaximeGagnebin: I think that we don't really need this step 1. In step 2, we could say:
delegate.roundHash = hash(randomSeed1 | delegate.address)
where randomSeed1 is the random seed for round N (we should probably define which one) as  specified in [LIP 0022]f

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
-		delegate.pubKeyHash = createHash(seedSource + delegate.publicKey)
+	forEach delegate in delegateList
+		delegate.roundHash = hash(seedSource | delegate.address)
 	```
+	where `|` indicates bytes concatenation.
 
-3. The delegates are reordered according to the hashes generated in step 2. This can be done according to any criteria (in an ascending order, for example) as long as the list is sorted in a deterministic and consistent way. Hence, the sorting algorithm must be [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) to generate the same list of delegates on every node on the network, even in the unlikely event of a hash collision in the second point. As an example, the `sort()` method in JavaScript can be used for this purpose.
-
-This will not only fix the uniformity issue (caused by the two issues raised in the [Motivation](#motivation) section), but it will also reduce complexity, allowing simpler, more human-readable code in any implementation.
+3. The delegates are reordered in lexicographical order according to the hashes generated in step 2.
```
MaximeGagnebin: This last point could specify "In the unlikely event that several delegates have the same hash, they are ordered lexicographically by delegate address."
This will never be necessary, but for me it feels a bit better if all possibilities are covered.

proposals/lip-0003.md
MaximeGagnebin: Another possibility would be that the "ordering" mechanism just receives a set of delegates AND a random seed, it then returns an ordered list of delegates.

The responsability of selecting the random seed is then left to LIP 0022.

proposals/lip-0003.md
janhack: This is a nice idea and uniquely defines the ordering. It should also be a simple change of defining the comparator function used for sorting as something like
` (delegate1.roundHash< delegate2.roundHash) ||  (delegate1.roundHash == delegate2.roundHash &&  (delegate1.address< delegate2.address))``.

proposals/lip-0003.md
janhack: A good point as it is not clear in this LIP from which round `randomSeed1` is used. LIP 0022 defines basically a function `randomSeed1(r)` where `r` is a round. I think then the `forger list` for round `r` (using the vocabulary from the documentation) is computed from the set of 103 delegates and the value `randomSeed1(r-1)`.

proposals/lip-0003.md
janhack: I agree that it would be better to make this part more timeless and implementation independent.

proposals/lip-0003.md
```diff
@@ -75,35 +78,32 @@ Given the study of the issue presented in the previous section, this LIP propose
...
 	```
+	where `randomSeed1` is computed as specified in [LIP 0022](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation). 
 
-2. One hash is generated per delegate using `seedSource` and the delegate public key (or ID). For simplicity, we propose to use the hashing function already utilised in the existing code, although the final choice here is left to the reference implementation. Example given in pseudo-code:
+2. One hash per delegate is generated using `seedSource` and the delegate address. For simplicity, we propose to use SHA-256, which is the hashing function already utilised in the existing function. Following the current implementation, an example in pseudo-code will be:
 
 	```
-	forEach delegate
-		delegate.pubKeyHash = createHash(seedSource + delegate.publicKey)
+	forEach delegate in delegateList
```
janhack: To make the specification self-contained and implementation-independent, it would be nice to shortly state what `delegateList` is.

proposals/lip-0003.md
IkerAlus: This LIP was one of the first ever written and it is not . Also, the motivation is very specific regarding the _bug_ in the said function. I re-phrased it and simplified it.

proposals/lip-0003.md
IkerAlus: thanks for the good point, added!

proposals/lip-0003.md
IkerAlus: good point for simplicity and clarification of the round, added, thanks!

PR #45, opened 2020-02-11
Update LIP-0022

IkerAlus:
- Use delegate address instead of delegate public keys in the algorithm and delegate lists.
- Remove the changes to LIP-0003 from the specifications
- Update reference to functions and links


proposals/lip-0022.md
```diff
@@ -10,7 +10,7 @@ Updated: 2019-10-30
 
 ## Abstract
 
-This LIP proposes to include two standby delegates in the forging delegate list of every round to incentivize a greater number of online nodes in the Lisk network. The selection of these standby delegates is done by a Randao-based weighted random selection scheme to ensure the fairness and unpredictability of the process. This scheme is also used to reorder the forging delegates list in every round.
+This LIP proposes to include two standby delegates in the forging delegate list of every round to incentivize a greater number of online nodes in the Lisk network. The selection of these standby delegates is done by a Randao-based weighted random selection scheme to ensure the fairness and unpredictability of the process. This scheme is also used to reorder the forging delegates list in every round (see [LIP-0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md)).
```
MaximeGagnebin:
```suggestion
This LIP proposes to include two standby delegates in the forging delegate list of every round to incentivize a greater number of online nodes in the Lisk network. The selection of these standby delegates is done by a Randao-based weighted random selection scheme to ensure the fairness and unpredictability of the process. This scheme is also used to reorder the forging delegates list in every round (see [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md)).
```

proposals/lip-0022.md
```diff
@@ -177,25 +177,19 @@ Internally, the algorithm works as follows:
...
 
-*   The round length is now 103 block slots. Currently, this is defined by the constant  <code>[ACTIVE_DELEGATES](https://github.com/LiskHQ/lisk-sdk/blob/310283408102ff9f891830225afe8124e833a608/config/default/constants.js#L55)</code>.  For clarity, this constant should be replaced for three constants defining the number of the top 101 delegates by `delegateWeight`, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants).
+*   The round length is now 103 block slots. Currently, this is defined by the constant  <code>[ACTIVE_DELEGATES](https://github.com/LiskHQ/lisk-sdk/blob/310283408102ff9f891830225afe8124e833a608/config/default/constants.js#L55)</code>.  For clarity, this constant can be replaced for three constants defining the number of the top 101 delegates by `delegateWeight`, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants).
```
MaximeGagnebin: I find this sentence pretty confusing, maybe something like:
"For clarity, this constant can be replaced by three constants: the number of top delegates by `delegateWeight`, i.e.,101, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants)."

proposals/lip-0022.md
```diff
@@ -177,25 +177,19 @@ Internally, the algorithm works as follows:
...
+*   `delegate1` and `delegate2` have to be included in the delegate list<sup>3</sup> before it is re-ordered for `round` + 1. Currently, the function <code>[createRoundDelegateList](https://github.com/LiskHQ/lisk-sdk/blob/5997714fde5181701fa0508eebd7f3c595cbf92c/elements/lisk-dpos/src/delegates_list.ts#L148)</code> computes this list before being re-ordered.
 
-Bear in mind that this proposal does not change any other point in the round processing once the list with the 103 forging delegates for the next round is computed.  
+This LIP also proposes to use the 20 bytes adddress of a delegate as defined in [LIP-0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md) for the [delegate lists](https://github.com/LiskHQ/lisk-sdk/blob/5997714fde5181701fa0508eebd7f3c595cbf92c/elements/lisk-dpos/src/delegates_list.ts#L124). This also implies two things:
```
janhack: This line and the two next line rather refer to the implementation than the protocol if I understand correctly. I don't understand why here the order in the delegate list is important because the forging order is defined in LIP 0003.

MaximeGagnebin:
```suggestion
This LIP also proposes to use the 20 bytes adddress of a delegate as defined in [LIP 0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md) for the [delegate lists](https://github.com/LiskHQ/lisk-sdk/blob/5997714fde5181701fa0508eebd7f3c595cbf92c/elements/lisk-dpos/src/delegates_list.ts#L124). This implies two things:
```

proposals/lip-0022.md
```diff
@@ -177,25 +177,19 @@ Internally, the algorithm works as follows:
...
+* The delegate address is used for tie break instead of the public key when two delegates have the same `delegateWeight`.
 
-This proposal also updates the proposed changes in [LIP-0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md). The variable `seedSource`, defined in the first bullet point of the Specification section in that document, is redefined as:
-
-```js
-var seedSource = randomSeed1;
-```
-
-The remaining changes of LIP-0003 remain the same and stay valid.
+* The delegate address has to be derived from the block property `generatorPublicKey` in order to verify the correct block slot (this is currently implemented in the function [verifyBlockForger](https://github.com/LiskHQ/lisk-sdk/blob/5997714fde5181701fa0508eebd7f3c595cbf92c/elements/lisk-dpos/src/delegates_list.ts#L227)).
```
MaximeGagnebin: Isn't that how we do it anyways ?

proposals/lip-0022.md
```diff
@@ -177,25 +177,19 @@ Internally, the algorithm works as follows:
...
 
-*   The round length is now 103 block slots. Currently, this is defined by the constant  <code>[ACTIVE_DELEGATES](https://github.com/LiskHQ/lisk-sdk/blob/310283408102ff9f891830225afe8124e833a608/config/default/constants.js#L55)</code>.  For clarity, this constant should be replaced for three constants defining the number of the top 101 delegates by `delegateWeight`, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants).
+*   The round length is now 103 block slots. Currently, this is defined by the constant  <code>[ACTIVE_DELEGATES](https://github.com/LiskHQ/lisk-sdk/blob/310283408102ff9f891830225afe8124e833a608/config/default/constants.js#L55)</code>.  For clarity, this constant can be replaced for three constants defining the number of the top 101 delegates by `delegateWeight`, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants).
 *   The change in the length of the round affects the logic to check when a round is finished which is 103 block slots in total.
```
MaximeGagnebin: I think this sentence is rather an implementation detail and could be removed.

proposals/lip-0022.md
IkerAlus: not really. Now the delegate lists contains the delegates' public keys and this can be directly verified against the property `generatorPublicKey` of the corresponding block. Once these changes are active, one will need to derive an address (as it is done now or as it is specified in LIP 0018) from `generatorPublicKey` in order to validate that the delegate who forged the block was the correct one for the corresponding slot.

proposals/lip-0022.md
IkerAlus: With the first bulletpoint I wanted to cover every case in which there is a tie by `delegateWeight`. Bear in mind that this is not about the ordered output of the function in LIP 0003 but about the input of said function. As said in line 186 here, a list of 103  delegates have to be created were 101 of them are the top 101 delegates by `delegateWeight`. The bulletpoint here covers this situation saying that if for example delegate 101 and delegate 102 have the same `delegateWeight`, the one with the "smaller" address will make it to the next forging round. This is now done by the function `createRoundDelegateList` but it is done with the delegate public keys. Perhaps it is clearer if I join both points in the same bullet point at line 186?

IkerAlus: Regarding the second bulletpoint at line 190, I think it is in the limit of being considered a protocol change. I preferred to state it here to avoid future issues.

proposals/lip-0022.md
IkerAlus: changed as proposed

proposals/lip-0022.md
IkerAlus: I agree. Removed.

proposals/lip-0022.md
janhack: I think this bullet point should be removed. The constants are part of the protocol and these implementation details should rather be decided at time of implementation.

janhack: Note that the statement "The round length is now 103 block slots." is incorrect as block slots can also be missed.

proposals/lip-0022.md
janhack: Good point regarding the tie-breaking. Maybe the "Round Computation and Delegate List" section could be re-structures a bit to make this more clear. One way would be to define the delegate list from scratch (or forger list as used in the documentation) and how it is created (add 101 active delegate by weight with tie-breaking by address plus 2 randomly selected standby delegates). Currently, it rather defines how an object (delegate_list class) in the code needs to be modified.

proposals/lip-0022.md
IkerAlus: good point about the round length. The actual protocol change is that a round has 103 blocks now. In the code currently this is specified by using the `ACTIVE_DELEGATES` but this has to change. I just tried to make that clear but I agree these details should be removed from the LIP.

proposals/lip-0022.md
```diff
@@ -177,25 +177,17 @@ Internally, the algorithm works as follows:
...
+A round is composed of 103 blocks now.
 
-*   The round length is now 103 block slots. Currently, this is defined by the constant  <code>[ACTIVE_DELEGATES](https://github.com/LiskHQ/lisk-sdk/blob/310283408102ff9f891830225afe8124e833a608/config/default/constants.js#L55)</code>.  For clarity, this constant should be replaced for three constants defining the number of the top 101 delegates by `delegateWeight`, the number of standby delegates, i.e., 2, and the length of the round (sum of the two previous constants).
-*   The change in the length of the round affects the logic to check when a round is finished which is 103 block slots in total.
-*   `delegate1` and `delegate2` have to be included in the delegate list<sup>3</sup> before it is re-ordered for `round` + 1. Currently, the function <code>[getKeysSortByVote](https://github.com/LiskHQ/lisk-sdk/blob/7c31c9d4b12a72f18a35ead79198b2ba6d4bd091/framework/src/modules/chain/submodules/delegates.js#L32)</code> computes this list before being re-ordered.
+Before being [shuffled](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md), the forger list<sup>3</sup> for `round` + 1 is constructed as follows:
```
janhack:
```suggestion
Before being ordered according to [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md), the set of 103 forgers for `round + 1` is constructed as follows:
```

janhack: For me it would be clearer if there is only one `forger list` for a round, i.e., the forgers in the order of assigned block slots. What is defined here is rather a set, as the order does not matter for the protocol (only the order in LIP 0003 matters). I find it a bit confusing to call this set forger list and also impose some arbitrary order on it.

janhack: I also find the foot note confusing as here you specify how this list is computed and it is unclear what computing with a delay of 2 rounds means.

proposals/lip-0022.md
```diff
@@ -177,25 +177,17 @@ Internally, the algorithm works as follows:
...
 
-Bear in mind that this proposal does not change any other point in the round processing once the list with the 103 forging delegates for the next round is computed.  
+* The addresses of the top 101 delegates at the end of `round` - 1 by lexicographic order according to the combined key `(delegateWeight, address)` are included in the forging list.
+* The addresses of `delegate1` and `delegate2`, defined in the previous section, are included at the end of the forger list in this order respectively.
```
janhack:
```suggestion
* The addresses of `delegate1` and `delegate2`, defined in the previous section, are added to the set of forgers.
```

proposals/lip-0022.md
janhack: You could also define the creation of the forger list here and make the 3rd step, order the delegates according to LIP 0003.

proposals/lip-0022.md
```diff
@@ -177,25 +177,17 @@ Internally, the algorithm works as follows:
...
-*   The change in the length of the round affects the logic to check when a round is finished which is 103 block slots in total.
-*   `delegate1` and `delegate2` have to be included in the delegate list<sup>3</sup> before it is re-ordered for `round` + 1. Currently, the function <code>[getKeysSortByVote](https://github.com/LiskHQ/lisk-sdk/blob/7c31c9d4b12a72f18a35ead79198b2ba6d4bd091/framework/src/modules/chain/submodules/delegates.js#L32)</code> computes this list before being re-ordered.
+Before being [shuffled](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md), the forger list<sup>3</sup> for `round` + 1 is constructed as follows:
 
-Bear in mind that this proposal does not change any other point in the round processing once the list with the 103 forging delegates for the next round is computed.  
+* The addresses of the top 101 delegates at the end of `round` - 1 by lexicographic order according to the combined key `(delegateWeight, address)` are included in the forging list.
```
janhack:
```suggestion
* The addresses of the top 101 delegates by delegate weight at the end of `round - 2` are added to the forger set. Ties are broken in favor of smaller address. 
```

janhack: - I think it should be the end of `round-2`. The same weights that are used for the standby delegate selection above.
- Sorting lexicographically by combined key `(delegateWeight, address)` is a bit confusing, because the delegate with smallest weight come first and it is unclear what top means then. Did you want to do the tie-breaking by larger address?

IkerAlus: True, the round was wrong. Also modified the bullet point to make it less confusing.

proposals/lip-0022.md
IkerAlus: Modified the first part as suggested. Also changed the footnote formulation to make it clearer, I dont want to remove it to have a quick link to check where this 2 rounds delay comes from.

proposals/lip-0022.md
IkerAlus: changed

proposals/lip-0022.md
IkerAlus: I also removed every occurrence of forging list as the structure being specified here.

proposals/lip-0022.md
```diff
@@ -177,25 +177,17 @@ Internally, the algorithm works as follows:
...
-*   The change in the length of the round affects the logic to check when a round is finished which is 103 block slots in total.
-*   `delegate1` and `delegate2` have to be included in the delegate list<sup>3</sup> before it is re-ordered for `round` + 1. Currently, the function <code>[getKeysSortByVote](https://github.com/LiskHQ/lisk-sdk/blob/7c31c9d4b12a72f18a35ead79198b2ba6d4bd091/framework/src/modules/chain/submodules/delegates.js#L32)</code> computes this list before being re-ordered.
+Before being ordered according to [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md) to generate the forging list, the set of forging delegates for `round` + 1 is constructed as follows:
 
-Bear in mind that this proposal does not change any other point in the round processing once the list with the 103 forging delegates for the next round is computed.  
+* The addresses of the top 101 delegates by `delegateWeight` at the end of `round` - 2 are added to the set<sup>3</sup>. Ties are broken in favor of smaller address.
```
janhack: To me now the sentence says the same things as the footnote at the end, which I find confusing.

PR #46, opened 2020-02-12
Distinguish between protocol level and UI level in LIP 0018

AndreasKendziorra:
LIP 0018 is updated such that there is a clear distinction between addresses on the protocol level (20 bytes) and the encoding for the UI level (adding a checksum and using a base32 encoding).
Before, it was specified that the checksum and Base32 encoding has to be validated during transaction validation. This is however meaningless since neither the checksum nor any other encoding details are signed or hashed. Only the 20 bytes of the public key hash are signed and hashed and therefore relevant for the protocol.


proposals/lip-0018.md
```diff
@@ -26,29 +26,57 @@ Secondly, the addresses do not have a checksum or any other error detection mech
...
+This section covers the definition of the address on the protocol level and further protocol related specifications. These addresses are also referred to as **binary addresses**.
 
-The output of step 5 is the address for the given account. An example of an address computation can be found in the [appendix](#address-computation). In the following subsections, some details of the steps above get clarified.
+#### Address Computation
+
+The binary address for an account with public key, `pubkey`, is generated by computing `SHA-256(pubkey)` and truncating the output to its first 160 bits.
+
+#### Usage in the Protocol
+
+The 160-bit binary addresses have to be used for the `recipientID` property of balance transactions. Moreover, if [LIP-0023](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) becomes active, then these addresses have to be used for
```
ManuGowda: @AndreasKendziorra do we plan to change the convention `recipientID` to `recipientAddress`? if its which LIP is address them? if they are addressed can we update here to `recipientAddress`?

proposals/lip-0018.md
```diff
@@ -26,29 +26,57 @@ Secondly, the addresses do not have a checksum or any other error detection mech
...
+#### Usage in the Protocol
+
+The 160-bit binary addresses have to be used for the `recipientID` property of balance transactions. Moreover, if [LIP-0023](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) becomes active, then these addresses have to be used for 
+
+- the `delegateAddress` property of votes, i.e., the entries in `asset.votes` of vote transactions, and
+- the `delegateAddress` property of unlock objects, i.e., the entries of `asset.unlockObjects` in unlock transactions.
+
+#### Address Length Validation in Transaction Validation
+
+When a transaction object contains a binary address, the length has to be validated. That means, the binary address must consist of 20 bytes. Otherwise, the transaction has to be rejected.
```
ManuGowda: is it good to keep the representation consistent? 160 bits instead of 20 bytes? it's not a big thing, but just for consistency. Or we can update the previous reference to be 20 bytes.

proposals/lip-0018.md
```diff
@@ -26,29 +26,57 @@ Secondly, the addresses do not have a checksum or any other error detection mech
...
 
-##### Creating checksum
+###### Creating Checksum
 
-The checksum can be computed by the function `createChecksum` below. For this, the output of step 1 has to be split into 5-bit strings, and each bit string has to be converted into an integer. Hence, the input for `createChecksum` is a list of 32 integers between 0 and 31. The output of `createChecksum` is a sequence of 6 integers between 0 and 31, where each integer represents a 5-bit string. Thus, the checksum consists of 30 bits.
+The checksum can be computed by the function `createChecksum` below. For this, the 160 bits of `addr` have to be split into 5-bit strings, and each bit string has to be converted into an integer. Hence, the input for `createChecksum` is a list of 32 integers between 0 and 31. The output of `createChecksum` is a sequence of 6 integers between 0 and 31, where each integer represents a 5-bit string. Thus, the checksum consists of 30 bits.
```
ManuGowda: same here, 160 bits

proposals/lip-0018.md
AndreasKendziorra: Yes, it is planned to use this wording. It will be established with this one: https://research.lisk.io/t/define-schema-and-use-generic-serialization-for-transactions/211?u=andreas.kendziorra
I prefer to have it changed there because once this change it active changing a transaction property name is no protocol change anymore. Whereas in this LIP, it would imply another (not really needed) protocol change.

proposals/lip-0018.md
AndreasKendziorra: In some places I find the bit representation more useful. For example below, where it is about splitting a bitstring into 5-bit substrings. And in some places I prefer to specify the length of the actual byte sequence that is used to represent an address.

Note, that if we used 155-bit hashes for the addresses, I would also sometimes say "155 bits" and sometimes "20 bytes" because representing 155 bits requires 20 bytes as well. Do you see what I mean?

proposals/lip-0018.md
AndreasKendziorra: same as above

PR #49, opened 2020-03-09
Add LIP defining a generic serialization for transactions

IkerAlus:
Add a new LIP defining a generic serialization method for transactions.


AndreasKendziorra:
> Looks all good to me, maybe the only thing is to remove the leading 0x from the signatures and in general bytes values in the bottom example. I think we had some discussion with @AndreasKendziorra and @MaximeGagnebin for the generic serialization LIP, and we decided to use 0x only for numbers. Maybe @AndreasKendziorra can confirm this?

We don't have to be strict about it. I think in this case there is no confusion. The case where I found the usage of "0x" confusing was the little endian varint encoding of an integer which could be false misinterpreted as  the regular hex representation of the integer.

PR #51, opened 2020-03-09
Add LIP "Define schema and use generic serialization for account state"

ricott1:
Add a new LIP defining a generic serialization method for account state.


proposals/lip-serialization-accounts
```diff
@@ -0,0 +1,415 @@
...
+
+
+### Handling Legacy Addresses
+
+After [LIP 0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md) "Use base32 encoding of long hash of public key plus checksum for address" is implemented, the address format in the Lisk protocol will change. In particular, the current 8 bytes addresses will be replaced by 20 bytes addresses. Accounts that have issued no transactions at the time of the address conversion will have no public key set, and as such the two address formats can not be linked (and the old format can not be updated). In this case, the `address`  property will contain an address in the old format and users will have to move their balance from an old to a new account by issuing a [Reclaim Transaction](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md#reclaim-transaction) from the new account. If the public key is present instead, the `address` property will be updated to the new format.
```
ricott1:
```suggestion
After [LIP 0018 "Use base32 encoding of long hash of public key plus checksum for address"](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md) is implemented, the address format in the Lisk protocol will change. In particular, the current 8 bytes addresses will be replaced by 20 bytes addresses. Accounts that have issued no transactions at the time of the address conversion will have no public key set, and as such the two address formats can not be linked (and the old format can not be updated). In this case, the `address`  property will contain an address in the old format and users will have to move their balance from an old to a new account by issuing a [Reclaim Transaction](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md#reclaim-transaction) from the new account. If the public key is present instead, the `address` property will be updated to the new format.
```

ricott1:
@IkerAlus What do you mean by format of the document?

PR #57, opened 2020-04-16
Add round computation case for less than 102 delegates

IkerAlus:


proposals/lip-0022.md
```diff
@@ -149,7 +149,7 @@ A round is composed of 103 blocks now.
 
 Before being ordered according to [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md) to generate the forging list, the set of forging delegates for `round` + 1 is constructed as follows:
 
-* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address.
+* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In the hipothetical case that 101 or less delegates were available, the addresses of these available delegates would be selected and the round computation would end here.
```
ricott1, line 152:
```suggestion
* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In the hypothetical case that 101 or less delegates are available, the addresses of the available delegates are selected and the round computation ends.
```

ricott1: or even remove hypothetical

proposals/lip-0022.md
```diff
@@ -149,7 +149,7 @@ A round is composed of 103 blocks now.
 
 Before being ordered according to [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md) to generate the forging list, the set of forging delegates for `round` + 1 is constructed as follows:
 
-* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address.
+* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In the case that 101 or less delegates are available, the addresses of the available delegates are selected and the round computation ends here.
```
janhack: I would prefer to use an enumerated list here because these are actually two steps that have to be done one after the other.

proposals/lip-0022.md
```diff
@@ -149,7 +149,7 @@ A round is composed of 103 blocks now.
 
 Before being ordered according to [LIP 0003](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md) to generate the forging list, the set of forging delegates for `round` + 1 is constructed as follows:
 
-* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address.
+* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In the case that 103 or less delegates are available, the addresses of the available delegates are selected and the round computation ends here.
```
janhack, line 152: I would prefer to use an enumerated list here because these are actually two steps that have to be done one after the other.

janhack: How about something like the following:
> From the set of all non-banned delegates at the end of `round` - 2, the addresses of the top 101 delegates by delegate weight are added to the forger set. Ties are broken in favor of the smaller address. In the case there are 103 or less delegates to select from,  all non-banned delegates at the end of `round` - 2 are added to the forger set and the following step of adding standby delegates is omitted.

janhack: Otherwise, it is not so clear what "available" means.

janhack: What is maybe also important to mention is that delegate 102 and 103 (if those exist) are considered standby delegates for the BFT consensus (i.e., their consensus votes are ignored).

IkerAlus: thanks for the suggestion. I changed the text as suggested. Also added a clarification for the BFT rules.

proposals/lip-0022.md
```diff
@@ -149,8 +149,8 @@ A round is composed of 103 blocks now.
...
 
-* The addresses of the top 101 delegates by delegate weight at the end of `round` - 2 are added to the set (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address.
-* The addresses of `delegate1` and `delegate2` are included to the set of forging delegates. These two delegates are the output of the selection algorithm specified in the next section.
+1. From the set of all non-banned delegates at the end of `round` - 2, the addresses of the top 101 delegates by delegate weight are added to set of forging delegates (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In case there are 103 or less delegates to select from, all non-banned delegates at the end of `round` - 2 are added to this set and the following step of adding 2 extra delegates is omitted.
```
ricott1, line 152:
```suggestion
1. The addresses of the top 101 delegates by delegate weight are added to the set of forging delegates from the set of all non-banned delegates at the end of `round` - 2 (the 2 rounds delay is specified in [Change of Delegates](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#change-of-delegates) subsection of LIP 0014). Ties are broken in favor of the smaller address. In case there are 103 or less delegates to select from, all non-banned delegates at the end of `round` - 2 are added to this set and the following step of adding 2 extra delegates is omitted.
```

proposals/lip-0022.md
```diff
@@ -145,12 +145,12 @@ As in the previous subsection, the random seed computation can be performed effi
 
 ### Round Computation
 
-A round is composed of 103 blocks now.
+A round is composed of 103 blocks now. In the case of having less than 103 delegates in the forging list, the forging list is looped again to assign the remaining slots (i. e., as it is done when delegates miss their forging slot).
```
MaximeGagnebin, line 148: Maybe here, the information could be more general. Something like: "The block slots in the round are assigned to the delegates by looping over the delegate list." And if we want some additional info, maybe "This implies that a delegate may forge multiple blocks in a round if some delegates missed their forging slot, or if the forging list is shorter than the round length."

ricott1: I like Maxime's formulation too, just a minor thing, I'd say: "The block slots in the round are assigned to the delegates by looping over the FORGING list FROM THE START. This implies that a delegate may forge multiple blocks in a round if some delegates missed their forging slot, or if the forging list is shorter than the round length."

janhack: I like the suggestion by @ricott1 as I think it is more clear than the current one.

IkerAlus: thanks for the suggestion, updated as @ricott1  said.

PR #58, opened 2020-04-17
Remove transactionID uniqueness check in LIP 0019

AndreasKendziorra:


proposals/lip-0019.md
```diff
@@ -44,10 +48,14 @@ transactionID = SHA-256(transactionBytes(tx))
 
 ### Usage of the New Format
 
-The new format has to be applied to every transaction, including the transactions included in the blockchain in the past, and has to be used in every aspect of the protocol, i.e., for validating the transactionID uniqueness.
+The new format has to be applied to every transaction, including the transactions included in the blockchain in the past, and has to be used in every aspect of the protocol.
```
janhack, line 51: As the transaction ID is not an explicit transaction property, I don't understand what "has to be applied to every transaction" means. Moreover, if we migrate Lisk mainnet as proposed in the LIP "Define decentralized snapshot and hardfork process" (https://research.lisk.io/t/define-decentralized-snapshot-and-hardfork-process/222), then there is no transaction history that needs to be migrated.

janhack: Maybe it is sufficient to write something like
> The new 256 bit transaction ID has to be used in every aspect of the protocol.

AndreasKendziorra: done

PR #66, opened 2020-07-13
Update LIPs to reflect public key removal from accounts

janhack:

### DIFF
--- a/proposals/lip-0035.md
+++ b/proposals/lip-0035.md
@@ -52,7 +52,6 @@ The state of an account on the Lisk mainnet after applying the block at height `
 
 * The value of the `address` property of an account is the 20-byte value specified in [LIP 0018](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md) if the account has a registered public key. Otherwise, it is the 8-byte value of the legacy address.
 * The value of the `balance` property is the account balance in Beddows.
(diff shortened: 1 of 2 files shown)


proposals/lip-0035.md
```diff
@@ -67,9 +66,9 @@ The state of an account on the Lisk mainnet after applying the block at height `
 
 In order to ensure the uniqueness of the account state, we perform the following additional checks:
 
-* Ensure that an account only has an associated public key if and only if there is at least one outgoing transaction from that account.
+* Ensure that an account has a 20-byte address if there is at least one outgoing transaction from that account and an 8-byte address otherwise.
```
ricott1, line 69: Just to be sure: there is no way that someone received a transaction in a 20-bytes address because the new address system comes after the regenesis, right?

also, consider changing 20-byte --> 20-bytes and 8-byte --> 8-bytes address

ricott1, line 69: Here and in several other places, should it be, e.g., 20-byte address or 20-bytes address?

Also, just to be sure, there is no way someone received a transaction in a 20-byte address because the new address system comes after regenesis, right?

janhack: I think 20-byte value is correct, similar to 5-day week or 32-bit value.

Yep, there are no transactions to a 20-byte address because the new address system is introduced after regenesis.

PR #67, opened 2020-08-13
Update LIP0018 specifications of reclaim transaction

MaximeGagnebin:
Updated the reclaim transaction in a way that can be extended from the base transaction, i.e., the fee is always paid by the account sending the transaction, and cannot be deduced from the reclaimed amount anymore.
The rationale for including `amount` in the reclaim transaction was also updated.

### DIFF
--- a/proposals/lip-0018.md
+++ b/proposals/lip-0018.md
@@ -190,7 +190,7 @@ The type of the reclaim transaction is the lowest positive integer that is not u
 
 ###### Schema
 
-The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must represent a number and this number must be equal to the balance associated to the address of the current format in Beddows.
+The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must be equal to the balance associated to the address of the current format in Beddows.
 
 A reclaim transaction is of the following form:
 
--- a/proposals/lip-0018.md
+++ b/proposals/lip-0018.md
@@ -190,7 +190,7 @@ The type of the reclaim transaction is the lowest positive integer that is not u
 
 ###### Schema
 
-The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _string_. The value of `asset.amount` must represent a decimal number and this number must be equal to the balance associated to the address of the current format in Beddows.
+The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must represent a number and this number must be equal to the balance associated to the address of the current format in Beddows.
 
 A reclaim transaction is of the following form:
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0018.md
```diff
@@ -190,7 +190,7 @@ The type of the reclaim transaction is the lowest positive integer that is not u
 
 ###### Schema
 
-The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _string_. The value of `asset.amount` must represent a decimal number and this number must be equal to the balance associated to the address of the current format in Beddows.
+The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must represent a number and this number must be equal to the balance associated to the address of the current format in Beddows.
```
ricott1, line 193:
```suggestion
The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must be equal to the balance associated to the address of the current format in Beddows.
```

I think that specifying that it must represent a number is redundant now (that is type uint64)

proposals/lip-0018.md
```diff
@@ -209,18 +209,17 @@ A reclaim transaction is of the following form:
 
 A reclaim transaction, `tx`, must fulfil the following points to be valid:
 
-- Let `old_addr` be the address according to the old format for `tx.senderPublicKey`. Then, `old_addr` must be contained in `unregisteredAddresses`.
+- Let `old_addr` be the address according to the old format for `tx.senderPublicKey`. `old_addr` must be contained in `unregisteredAddresses`.
 - Let `old_balance` be the sum of all funds send to `old_addr` in Beddows. Then `tx.asset` must contain the property `amount`, and the number represented by `tx.asset.amount` must be equal to `old_balance`.
```
ricott1, line 213:
```suggestion
- Let `old_balance` be the sum of all funds send to `old_addr` in Beddows. Then `tx.asset` must contain the property `amount`, and `tx.asset.amount` must be equal to `old_balance`.
```

proposals/lip-0018.md
```diff
@@ -209,18 +209,17 @@ A reclaim transaction is of the following form:
...
-- Let `old_balance` be the sum of all funds send to `old_addr` in Beddows. Then `tx.asset` must contain the property `amount`, and the number represented by `tx.asset.amount` must be equal to `old_balance`.
-- Let `new_addr` be the address of the new format for `tx.senderPublicKey` and let `new_balance` be the balance of `new_addr`. If there exists no account for `new_addr`, then `new_balance` is 0. Then `tx.fee` must be smaller or equal than `old_balance + new_balance`.
+- Let `old_addr` be the address according to the old format for `tx.senderPublicKey`. `old_addr` must be contained in `unregisteredAddresses`.
+- Let `old_balance` be the sum of all funds send to `old_addr` in Beddows. Then `tx.asset` must contain the property `amount`, and `tx.asset.amount` must be equal to `old_balance`.
+
+Notice that the base properties of the transaction also have to be valid. In particular, this implies that the transaction fee is paid by the account corresponding to `tx.senderPublicKey`. In some cases, users might need to credit accounts specifically to reclaim the balance sent to "unregistered" addresses.
```
IkerAlus, line 215: Suggestion:
```
Notice that the base properties of the transaction also have to be valid. In particular, this implies that the transaction fee is paid by the account corresponding to `tx.senderPublicKey`. Hence this account needs to exist beforehand and have enough balance to pay for the fee.
```

janhack: I like Iker's suggestion. The current formulation of the last sentence would rather be suited for the rationale section. Also, isn't it rather "always" than "in some cases"?

proposals/lip-0018.md
```diff
@@ -190,7 +190,7 @@ The type of the reclaim transaction is the lowest positive integer that is not u
 
 ###### Schema
 
-The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _string_. The value of `asset.amount` must represent a decimal number and this number must be equal to the balance associated to the address of the current format in Beddows.
+The `asset` property of a reclaim transaction needs to contain the `amount` property which is of type _uint64_. The value of `asset.amount` must be equal to the balance associated to the address of the current format in Beddows.
```
janhack, line 193: This LIP was written when we did not have protobuf serialization and Lisk amounts in JSON were strings. I don't think it adds much clarification to change it here and rather makes the LIP inconsistent. Later in the serialization section we write " interpreted as a 64-bit signed integer" for example. LIP 0028 already defines all the serializations, also for the reclaim transaction.

proposals/lip-0018.md
MaxGagnebin: Changed to Iker's suggestion.

proposals/lip-0018.md
MaxGagnebin: Changed the serialization to say "unsigned integer". This way, the types are consistent with future LIPs, which I think is good if anyone checks this LIP to have info on the reclaim transaction.

proposals/lip-0018.md
MaximeGagnebin: Changed for Iker's suggestion.

proposals/lip-0018.md
MaximeGagnebin: Changed the serialization section to say "unsigned integer". I understand your point. I think that here it is worth changing the type of `amount` to be more coherent with future LIPs.

janhack: That sounds also good.

karmacoma (approved):

PR #70, opened 2020-08-27
LIP28: Update due to new on-chain architecture

IkerAlus:
removed `type` property and added `moduleID` and `assetID` to the main transaction schema


proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
+Assuming that the JSON schemas for the `transactionSchema` and for the `asset` properties are provided according to the specifications in the [LIP 0027](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md), the serialization and deserialization of a transaction object are done as follows.
 
 ### Serialization
 
 Consider a data structure `trsData` representing a valid (signed or unsigned) transaction object to be serialized. The serialization procedure is done in 3 steps:
 
-1. The `trsData.asset` property is serialized to bytes according to the corresponding  `asset` property schema specified by the `trsData.type` property (see [below](#baseTransaction-schema)).
+1. The `trsData.asset` property is serialized to bytes according to the corresponding  `asset` property schema specified by the combination of `trsData.moduleID` and `trs.assetID` properties (see [below](#transactionSchema-schema)).
```
MaximeGagnebin, line 48: In a few places, you use `trsData.moduleID` and `trs.assetID`. Why is it not `trsData.assetID` ?

proposals/lip-0028.md
```diff
@@ -81,51 +81,57 @@ Given a data structure, `signedTrsData`, representing a transaction object with
...
 
-1. `type`: An integer identifying the type of the transaction. It specifies the JSON schema to be used to serialize-deserialize the `asset` property of the transaction.
-2. <code>[nonce](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0015.md)</code>: An integer which is unique for each transaction from the account corresponding to the <code>senderPublicKey.</code>
-3. <code>fee</code>: An integer that specifies the fee in Beddows to be spent by the transaction.
-4. <code>senderPublicKey</code>: The public key of the account issuing the transaction. A valid public key is 32 bytes long.
-5. <code>asset</code>: The serialized asset.
-6. <code>signatures</code>: An array with the signatures of the transaction. Note that this property is an array which implies that in case of <code>signatures</code> being initialized to its default value, it does not appear in the binary message. The elements of this array are bytes, either of length 64, or of length 0.
+1. `moduleID`: An integer identifying the module the transaction belongs in. Together with `assetID` specifies the JSON schema to be used to serialize-deserialize the `asset` property of the transaction.
```
MaximeGagnebin, line 88:
```suggestion
1. `moduleID`: An integer identifying the module the transaction belongs to. Together with `assetID`, this specifies the JSON schema to be used to serialize-deserialize the `asset` property of the transaction.
```

proposals/lip-0028.md
```diff
@@ -81,51 +81,57 @@ Given a data structure, `signedTrsData`, representing a transaction object with
...
+2. `assetID`: An integer identifying the specific asset for the transaction in the module. Together with `moduleID` specifies the JSON schema to be used to serialize-deserialize the `asset` property of the transaction.
```
MaximeGagnebin, line 89:
```suggestion
2. `assetID`: An integer identifying the specific asset for the transaction in the module. Together with `moduleID`, this specifies the JSON schema to be used to serialize-deserialize the `asset` property of the transaction.
```

proposals/lip-0028.md
```diff
@@ -371,7 +377,8 @@ This proposal introduces a hard fork in the network. Since the transaction seria
 
 ```js
 myTrs = {
```
MaximeGagnebin, line 379: I can only comment here. I think that the binary of `myTrs` below should also be updated.

proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
 
-As it is shown in the diagram below, every transaction object in the Lisk protocol, regardless of the type, has a common set of 6 properties. The data related to the specific transaction type is given in the `asset` property. This architecture is also expected to be used for custom transactions created with the Lisk SDK. For simplicity and consistency, custom transactions should have these 6 properties, while all the new logic and data is defined with respect to the custom `asset` property.
+As it is shown in the diagram below, every transaction object in the Lisk protocol, regardless of the type, has a common set of 7 properties. The data related to the specific transaction type is given in the `asset` property. This architecture is also expected to be used for custom transactions created with the Lisk SDK. For simplicity and consistency, custom transactions should have these 7 properties, while all the new logic and data is defined with respect to the custom `asset` property.
 
 ![Transaction Serialization Diagram](lip-0028/transaction_serialization_diagram.png)
```
MaximeGagnebin, line 36: In this diagram, we still mention "Base Transaction", but I think you removed this term from the rest of the LIP. I think it could also be removed here.

MaximeGagnebin (changes requested):
Only few changes, mainly wording. Also, I think the binary example should be checked. Nice work!

proposals/lip-0028.md
ricott1: when you say combination, you mean literally the concatenation of the values? Because in principle this sounds as we could do any operation taking these 2 values as input. Could we maybe explicitly state how this combination is taken?

ricott1 (changes requested):
Just a minor thing, and I'd rename transactionSchema to something else (baseTransaction made a lot of sense to me)

proposals/lip-0028.md
IkerAlus: fixed

proposals/lip-0028.md
IkerAlus: those are mistakes @MaximeGagnebin  bad copy paste habits.

proposals/lip-0028.md
IkerAlus: fixed

proposals/lip-0028.md
IkerAlus: fixed

proposals/lip-0028.md
IkerAlus: This is specified in the new  LIP 0033. I added the link to the specs of LIP 0033 that hopefully should be updated not much later we update this one (assuming it is not a new LIP). @ricott1

ricott1 (approved):

MaximeGagnebin (approved):

proposals/lip-0028.md
```diff
@@ -3,15 +3,15 @@ LIP: 0028
 Title: Define schema and use generic serialization for transactions
 Author: Iker Alustiza <iker@lightcurve.io>
 Discussions-To: https://research.lisk.io/t/define-schema-and-use-generic-serialization-for-transactions/211
-Type: Informational
+Type: Standards Track
 Created: 2020-02-18
 Updated: 2020-07-21
 Requires: 0027
```
janhack, line 9: The new LIP proposing to replace the `type` property by the `moduleID` and `assetID` property should also be added as requirement.

proposals/lip-0028.md
```diff
@@ -3,15 +3,15 @@ LIP: 0028
...
 ```
 
 ## Abstract
 
-This LIP defines a generic serialization for transactions by specifying the appropriate JSON schema. The proposed serialization method is applied in two steps, first to the asset property, where type-specific properties are defined, and then to the base transaction object. It will be used for signing but also for storing and transmitting transactions in the Lisk protocol. Further, this shall be the default serialization method for custom transactions created via the Lisk SDK.
+This LIP defines a generic serialization for transactions by specifying the appropriate JSON schema. The proposed serialization method is applied in two steps, first to the asset property, where type-specific properties are defined, and then to the transaction object. It will be used for signing but also for storing and transmitting transactions in the Lisk protocol. Further, this shall be the default serialization method for custom transactions created via the Lisk SDK.
```
janhack, line 14: Consider the following suggestion for the second sentence. I tried to avoid writing "type-specific" as there is no more transaction type. Moreover, "transaction base object" may be more clear as the transaction object is rather the whole object including the nested asset.

> The proposed serialization method is applied in two steps, first to the transaction asset, which contains the custom properties specific to the transaction, and then to the transaction base object, which is the same for all transactions.

proposals/lip-0028.md
janhack: The image is shown very large and looks quite ugly this way. Maybe we could change to using HTML, e.g.,
`drawing`
 See https://stackoverflow.com/questions/14675913/changing-image-size-in-markdown

proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
 
-As it is shown in the diagram below, every transaction object in the Lisk protocol, regardless of the type, has a common set of 6 properties. The data related to the specific transaction type is given in the `asset` property. This architecture is also expected to be used for custom transactions created with the Lisk SDK. For simplicity and consistency, custom transactions should have these 6 properties, while all the new logic and data is defined with respect to the custom `asset` property.
+As it is shown in the diagram below, every transaction object in the Lisk protocol, regardless of the type, has a common set of 7 properties. The data related to the specific transaction type is given in the `asset` property. This architecture is also expected to be used for custom transactions created with the Lisk SDK. For simplicity and consistency, custom transactions should have these 7 properties, while all the new logic and data is defined with respect to the custom `asset` property.
```
janhack, line 34:
```suggestion
As it is shown in the diagram below, all transaction objects in the Lisk protocol have a common set of 7 properties. The data related to the specific transaction type is given in the `asset` property. This architecture is also expected to be used for custom transactions created with the Lisk SDK. For simplicity and consistency, custom transactions should have these 7 properties, while all new data is defined as part of the custom `asset` property.
```

janhack: I suggested to remove some mentions of "type" where it was not necessary. As this LIP only talks about the transaction data, I would remove "all the new logic". Also note that custom transaction logic can still uitlize base properties such as the sender.

proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
+For this reason, this LIP proposes to divide the serialization-deserialization process. First it defines a serialization-deserialization process for the whole transaction object with the `transactionSchema` JSON schema where the `asset` property is assumed to be of type `bytes`. Secondly, it defines the serialization-deserialization process for the `asset` property with the corresponding `asset` property JSON schema depending on the type of the transaction (specified by the unique combination of `moduleID` and `assetID` properties). There are 7 different `asset` property schemas, one for each of the transaction types in the Lisk protocol. In order to simplify the creation of custom transactions with the SDK and to make the validation process less error prone, every property defined inside a given `asset` schema is required.
```
janhack, line 38: > specified by the unique combination of `moduleID` and `assetID` properties

I would say "given by the values of  `moduleID` and `assetID`".

proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
+Assuming that the JSON schemas for the `transactionSchema` and for the `asset` properties are provided according to the specifications in the [LIP 0027](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md), the serialization and deserialization of a transaction object are done as follows.
 
 ### Serialization
 
 Consider a data structure `trsData` representing a valid (signed or unsigned) transaction object to be serialized. The serialization procedure is done in 3 steps:
 
-1. The `trsData.asset` property is serialized to bytes according to the corresponding  `asset` property schema specified by the `trsData.type` property (see [below](#baseTransaction-schema)).
+1. The `trsData.asset` property is serialized to bytes according to the corresponding  `asset` property schema specified by the combination of `trsData.moduleID` and `trsData.assetID` properties (see [below](#transactionSchema-schema)) as specified in [LIP 0033](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0033.md#specification).
```
janhack, line 48:
```suggestion
1. The `trsData.asset` property is serialized to bytes according to the corresponding  `asset` property schema given by the values of `trsData.moduleID` and `trsData.assetID`. The correspondence between `asset` property schema and values of `trsData.moduleID` and `trsData.assetID` is defined in [LIP 0036](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0036.md#specification).
```

janhack: Both "see below" and "as specified in" together are a bit confusing without saying what is defined where. I already changed the link although it is currently not working.

proposals/lip-0028.md
```diff
@@ -31,37 +31,37 @@ When a transaction object is serialized, also the asset property has to be seria
...
+3. The transaction object from step 2 is serialized according to the `transactionSchema` schema.
 
 ### Deserialization
 
 Consider a binary message `trsMsg` to be deserialized. The deserialization procedure is done in 3 steps:
 
-1. The serialized bytes of `trsMsg` are deserialized according to the `baseTransaction` schema to obtain `trsData`.
-2. The bytes value of the `trsData.asset` property is deserialized according to the corresponding `asset` property schema specified by the `trsData.type` property.
+1. The serialized bytes of `trsMsg` are deserialized according to the `transactionSchema` schema to obtain `trsData`.
+2. The bytes value of the `trsData.asset` property is deserialized according to the corresponding `asset` property schema specified by the combination of `trsData.moduleID` and `trsData.assetID` properties as specified in [LIP 0033](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0033.md#specification).
```
janhack, line 57:
```suggestion
2. The bytes value of the `trsData.asset` property is deserialized according to the corresponding `asset` property schema given by the values of `trsData.moduleID` and `trsData.assetID`.
```

janhack: Maybe the references once above in the serialization are sufficient.

proposals/lip-0028.md
janhack: Was the binary updated and reviewed?

proposals/lip-0028.md
```diff
@@ -136,7 +142,7 @@ baseTransaction = {
 
 ### `asset` property schema
 
-There is a unique `asset` property schema per transaction type, implied by the value of the `type` property. All the properties defined in any `asset` schema are required.
+There is a unique `asset` property schema per transaction type, implied by the unique combination of the values in `moduleID` and `assetID` properties. All the properties defined in any `asset` schema are required.
```
janhack, line 145:
```suggestion
The `asset` property schema of a transaction has to be uniquely implied by the values `moduleID` and `assetID`. All  properties defined in any `asset` schema are required.
```

janhack (changes requested):
I added some comments. We also need to update the links to correctly reference the LIP introducing the `moduleID` and `assetID`.

proposals/lip-0028.md
IkerAlus: Yes, as per the last commit.

proposals/lip-0028.md
IkerAlus: thanks for the nice suggestion, I changed the size to something nicer, however centering the image is not so easy. It still looks fine.

proposals/lip-0028.md
IkerAlus: changed, thanks

proposals/lip-0028.md
IkerAlus: changed

proposals/lip-0028.md
IkerAlus: fixed

proposals/lip-0028.md
IkerAlus: fixed

proposals/lip-0028.md
IkerAlus: LIP 0036 added as a requirement

shuse2 (approved):

janhack (approved):
Looks all good.

karmacoma (approved):

PR #72, opened 2020-09-16
Update LIP 35 to clarify account migration

janhack:
This PR adds some details about the exact account migration due to accounts with a non-standard string representation in Lisk Core 2.0. The PR also changes the tie-breaking in the initDelegates array to simplify the implementation.


nazarhussain (approved):

shuse2 (approved):

proposals/lip-0035.md
```diff
@@ -64,25 +64,27 @@ The state of an account on the Lisk mainnet after applying the block at height `
...
+    * If there is an account `acc_standard` with standard string representation corresponding to the same `legacyAddress`, then the balance of `acc_non_standard` is added to the balance of the migrated account `acc_standard`. Note that the migrated account `acc_standard` may have an 8-byte address or a 20-byte address.
+    * If there no other account with standard string representation corresponding to `legacyAddress`, then the balance of `acc_non_standard` is added to migrated account with address `legacyAddress` and all properties set to default. The migration should ensure that there is at most one account for every 8-byte address, as there can be multiple non-standard accounts that need to be merged to the same account (e.g., `04L` and `88888888888888888888L`).
```
IkerAlus, line 75:
```suggestion
    * If there is no other account with standard string representation corresponding to `legacyAddress`, then the balance of `acc_non_standard` is added to migrated account with address `legacyAddress` and all properties set to default. The migration should ensure that there is at most one account for every 8-byte address, as there can be multiple non-standard accounts that need to be merged to the same account (e.g., `04L` and `88888888888888888888L`).
```

IkerAlus: also, shouldn't the last sentence of this bullet point be outside the bullet point since it is a general comment that applies to the entire migration of accounts?

proposals/lip-0035.md
```diff
@@ -64,25 +64,27 @@ The state of an account on the Lisk mainnet after applying the block at height `
...
+* Discard the so-called *Genesis Account* with public key `d121d3abf5425fdc0f161d9ddb32f89b7750b4bdb0bff7d18b191d4b4bafa6d4` (Lisk address `6566229458323231555L` in the old format), which has negative balance.
+* Migrate every account, except the Genesis Account and any account without incoming transaction, as follows.
+  * If the string representation of an account in Lisk Core 2.0 is standard (The string is `0L` or does not start with `0`. The decimal number is between `0` and `2^64-1` inclusive), then the account is migrated as defined above. If the account has at least one outgoing transaction, then the migrated account has a 20-byte address and otherwise it has an 8-byte address.
+  * If the string representation of an account in Lisk Core 2.0 is non-standard (It is not `OL`, but starts with `0`. Or the decimal number is larger than `2^64-1`), the account corresponds to one unique 8-byte address, denoted by `legacyAddress`, which are the 8 bytes signed in the transaction creating the account. Note that the account cannot have any outgoing transaction as the string representation is non-standard. The non-standard account `acc_non_standard` is migrated as follows:
```
IkerAlus, line 73: I find this point a bit confusing: 
Regarding the first sentence: 
a) there can be more than one non-standard accounts with the same 8-byte address, as it is said in the last sentence of the line 75, right?
b) did not the transaction creating the account sign the decimal + L string and that is why we have addresses over the 8 bytes range?

IkerAlus: I  believe the specs are right, but the way it is written confuses me. the Rationale section clarified it afterwards.

IkerAlus (changes requested):
Please consider my comments and the little typo fix

proposals/lip-0035.md
janhack: In the Lisk Core 2.0 serialization, the string representation (`123L`) is serialized as 8-bytes, see https://github.com/LiskHQ/lisk-sdk/blob/v2/elements/lisk-transactions/src/utils/get_transaction_bytes.ts.

proposals/lip-0035.md
janhack: I fixed the type. And I added " to ensure that all accounts with the same legacyAddress value are migrated to the same account" to whole point about migrating non-standard accounts. You are right that this is the goal of the whole migration of non-standard accounts (for only standard accounts it is easily satisfied).

IkerAlus (changes requested):
The new changes LGTM.

Two details:
The link to LIP22 at line 132 is broken because of bad formatting.
Also, I suggest to remove the explicit title of every mention to LIP 0034 for uniformity with the rest.

janhack:
I fixed the link and changed the reference style for LIP 0034. Please have another look.

IkerAlus (approved):
LGTM

karmacoma (approved):

PR #85, opened 2021-05-20
Update LIP31 by including two new appendixes and other minor changes

ricott1:
We update LIP31 to:

 - Include an inclusion proof protocol for leaf nodes only
 - define an update protocol for leaf nodes
 - change the variable name path to siblingHashes for consistency with the SMT LIP
 - change the variable name datalength to size
 - add the protocol for ‘right verification’, i.e. using the appendPath and a right witness to recompute the root of the tree
 - add a protocol for inclusion proof for leaf node values
 - change the scope of queryData in the inclusion proof protocol to be an array of hashes queryHashes

### DIFF
--- a/proposals/lip-0031.md
+++ b/proposals/lip-0031.md
@@ -6,7 +6,7 @@ Discussions-To: https://research.lisk.io/t/introduce-merkle-trees-and-inclusion-
 Status: Draft
 Type: Informational
 Created: 2020-02-19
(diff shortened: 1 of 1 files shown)


MaximeGagnebin (approved):
All good from my side.

IkerAlus (approved):
LGTM, great work.

karmacoma (approved):

PR #92, opened 2021-08-11
Add LIP: "Introduce Auth module"

ricott1:


proposals/lip-introduce_the_auth_module.md
```diff
@@ -0,0 +1,220 @@
...
+
+```python
+isValidNonce(trs):
+    let senderAddress be the address corresponding to trs.senderPublicKey
+    return trs.nonce == authAccount(senderAddress).nonce
+```
+
+Notice that a transaction in the transaction pool with a nonce greater than the account nonce is considered "pending" rather than invalid, since it could become valid in the future.
+
+* Check the signatures property of the transaction [as explained in LIP 0017](https://github.com/LiskHQ/lips-staging/blob/update-LIP31/proposals/lip-0017.md#signatures-replaces-signature) using `authAccount(address)`, where `address` is the address corresponding to `trs.senderPublicKey`.
```
IkerAlus, line 205: this link is pointing to the wrong URL.

ricott1: good catch, thanks

IkerAlus (approved):

IkerAlus (commented):
A comment regarding the entries creation needs clarification before approval.

proposals/lip-introduce_the_auth_module.md
```diff
@@ -0,0 +1,235 @@
...
+
+As part of the verification of a transaction `trs`, the following checks are applied. If the checks fail the transaction is invalid and the whole block is discarded. 
+
+* Check that the transaction nonce is equal to the account nonce by returning `isValidNonce(trs)`.
+
+```python
+isValidNonce(trs):
+    let senderAddress be the address corresponding to trs.senderPublicKey
+    if there is no entry in the in the auth data substore with storeKey == senderAddress:
+        return True
```
IkerAlus, line 201: I think this still should check that `trs.nonce == 0`.

IkerAlus: it is true it is not a security issue not to enforce it, but I think it makes sense

IkerAlus (changes requested):
One comment regarding the nonce check.

proposals/lip-introduce_the_auth_module.md
ricott1: thanks!

IkerAlus (approved):
LGTM

MaximeGagnebin (approved):
LGTM

shuse2 (approved):

ishantiw (approved):

karmacoma (approved):

PR #99, opened 2021-08-26
Update status and reference implementation

shuse2:
### Update status
- lip-0005
- lip-0006
- lip-0011

Update status of above lips to `Obsolete`

### Update reference implementation
- lip-0013
- lip-0014
- lip-0015
- lip-0016
- lip-0017
- lip-0021
- lip-0022
- lip-0023
- lip-0024
- lip-0025
- lip-0026
- lip-0035
- lip-0036

Update Reference implementation of above LIPs


proposals/lip-0021.md
```diff
@@ -131,4 +131,4 @@ The changes will introduce a hard fork as the vote transactions would change and
 
 ## Reference Implementation
```
janhack, line 132: This LIP is withdrawn and was never implemented.

proposals/lip-0024.md
```diff
@@ -207,4 +207,4 @@ If the implementation of this proposal is done at height `h` then we additionall
 
 ## Reference Implementation
 
-TBD
+[Add fail safe delegate banning mechanism](https://github.com/LiskHQ/lisk-sdk/issues/5058)
```
janhack, line 210: I did not have a closer look, but it seems like the implementation you link is mainly about productivity banning, which was already part of LIP 0023, and not about the PoM transaction defined in this LIP.

proposals/lip-0013.md
```diff
@@ -363,3 +363,7 @@ This appendix covers the impact of burning the minimum fee on the token supply d
 ![Reward 2](lip-0013/reward_2.png)
 
 ![Reward 1](lip-0013/reward_1.png)
+
+## Reference Implementation
```
janhack, line 367: There is already a section "Reference Implementation" in this LIP.

janhack (changes requested):
I left some minor comments. I think it would be nice if you could also add the "Reference Implementation" section to the "template.md" file in the root directory. This way we also have this consistent structure in the template.

janhack (approved):
Looks good.

karmacoma (approved):

PR #100, opened 2021-08-31
Add LIP "Introduce PoA module"

IkerAlus:


proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,591 @@
+```
+LIP:
+Title: Introduce PoA module
+Author: Iker Alustiza <iker@lightcurve.io>
```
ManuGowda, line 4: add ishan bitte

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,591 @@
...
+        *   The `aggregationBits` argument is the byte array given in `trs.params.aggregationBits`.
+        *   The `signature` argument is the aggregate signature given in `trs.params.signature`.
+        *   The `m` argument is the output bytes of the serialization, as specified in [LIP 0027][lip27], of `trs.params.validatorAddresses`, `trs.params.weights`, `trs.params.validatorsUpdateNonce`, and `trs.params.threshold` properties according to the following schema:
+ 
+        ``` java 
+        validatorSignatureMessage = {
+            "type": "object",
+            "properties": {
+               "validatorAddresses": {
+               "type": "array",
```
ManuGowda, line 444: format seems odd

ManuGowda (approved):

proposals/lip-Introduce_PoA_module.md
IkerAlus: true, added

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+#### Snapshot Substore
+
+This substore contains the snapshot of the active authorities for the current round, next round and in two rounds.
+
+##### Store Prefix, Store Key, and Store Value
+
+* The store prefix is set to `STORE_PREFIX_SNAPSHOT`.
+* Each store key is `uint32be(roundNumber)`, where `roundNumber` can be 0, 1 or 2 corresponding to the current round, the next round and in two rounds respectively.
+* Each store value is the serialization of an object following `snapshotStoreSchema`.
+* Notation: Let `snapshotStore(roundNumber)` be the store entry with prefix STORE_PREFIX_SNAPSHOT and key `uint32be(roundNumber)`.
```
janhack, line 170:
```suggestion
* Notation: Let `snapshotStore(roundNumber)` be the store entry with prefix `STORE_PREFIX_SNAPSHOT` and key `uint32be(roundNumber)`.
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+                      "dataType": "bytes",
+                      "fieldNumber": 2
+                  }
+	      }
+	   }
+        },		    
+        "threshold": {
+           "dataType": "uint64",
+           "fieldNumber": 2,
+        },
```
janhack, line 199:
```suggestion
        }
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+   "type": "object",
+   "properties": {
+       "validators: {
+          "type": "array",
+          "fieldNumber": 1,    
+          "items": {
+              "type": "object",
+              "required": ["address", "weight"],
+              "properties": {
+	          "address": {
```
janhack, line 185: I think here tabs and spaces are mixed and the indentation is incorrect in the source.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+    },
+    "required": ["validators", "threshold"]
+ }
+ ```
+
+##### Properties and Default Values
+
+The properties of this schema are as follows:
+
+  * Each element in the `validators` array corresponds to a validator and stores its address and BFT weight property. The elements in the array must be sorted lexicographically by `address` property.
```
janhack, line 209: Maybe I would simply call it `weight` property as you did before. Because the weight in PoA is used both as BFT weights as well as weight for the authority update command.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+
+##### Verification 
+ 
+Let `trs` be a transaction with module ID `MODULE_ID_POA` and command ID `COMMAND_ID_REGISTRATION_AUTHORITY` to be verified. 
+The list of verification conditions for `trs.params` is as follows:
+
+* The `trs.params.name` property has to contain only characters from the set `[a-z0-9!@$&_.]`, must not be empty and has to be at most `MAX_LENGTH_NAME` characters long.
+* Let `address` be the 20-byte address derived from `trs.senderPublicKey`. Then `address` must not already be registered as a validator. This is, the validator substore has no entry with store key `address`.
+* The value of `trs.params.name` must not already be registered as a validator name. This is, the name substore has no entry with store key `trs.params.name`.
+* `isKeyRegistered(trs.params.blsKey)` must return `true`, where the function is defined in the [Validators module][validatorsModule].
```
janhack, line 299: Shouldn't `isKeyRegistered(trs.params.blsKey)` return `False`?

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+* `PopVerify(trs.params.blsKey, trs.params.proofOfPossession)` must return `VALID`, where [`PopVerify`][bls-specs-v4-popverify] is part of the [BLS signature scheme][lip-bls-public-key-registration].
+* `tx.params.generatorKey` must have length 32.
```
janhack, line 301: Do we not have similar size constraints for the `blsKey` and `proofOfPossession`?

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+   },
+   "required": [
+      "newValidators",
+      "threshold",
+      "validatorsUpdateNonce",
+      "signature",
+      "aggregationBits"
+   ]
+}
+
```
janhack, line 408: Also above the a spaces and tabs mixed that is why the indentation in the source seems incorrect.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+               },
+            },
+            "required": [
+                "newValidators",
+                "threshold",
+                "validatorsUpdateNonce"
+            ]
+        }
+        ```
+        *   The `weights` argument is set to the `weights` property of `snapshotStore(0).validators` array.
```
janhack, line 480: The array itself does not have a `weights` property, but only its elements. Maybe we could say
> The `weights` argument is set to the array containing the `weight` property of the elements in `snapshotStore(0).validators` in the same order.

Or explicit and short

> > The `weights` argument is set to `[ validator.weight for validator in snapshotStore(0).validators]`.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+```python
+chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresses)
+
+# Pass the required information to the BFT module
+BFTThreshold = snapshotStore(0).threshold
+setBFTParameters(BFTThreshold, BFTThreshold, snapshotStore(0).validators)
+
+# Pass the list of validators to the Validators module
+for i in range(len(snapshotStore(0).validators)):
+    addresses =  snapshotStore(0).address[i]
```
janhack, line 533:
```suggestion
    addresses = snapshotStore(0).address[i]
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+        or currentBFTParameters.certificateThreshold != BFT_THRESHOLD):
+        # Pass the required information to the BFT module
+        setBFTParameters(BFTThreshold, BFTThreshold, snapshotStore(0).validators)
+    
+    # Reshuffle the list of validators and pass it to the Validators module
+    roundStartHeight = chainProperties.roundEndHeight - len(snapshotStore(0).validators) + 1
+    randomSeed = getRandomBytes(roundStartHeight, len(snapshotStore(0).validators))
+    
+    for i in range(len(snapshotStore(0).validators)):
+    addresses =  snapshotStore(0).address[i]
```
janhack, line 564: The indentation here is incorrect. It would also check if tabs and spaces are mixed here.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,607 @@
...
+The function has the following input parameters in the order given below:
+
+* `validatorsList`: An array of pairwise distinct 20-byte addresses.
+* `randomSeed`: A 32-byte value representing a random seed.
+
+##### Returns
+This function returns an array of bytes with the re-ordered list of addresses.
+
+##### Execution
+The shuffling algorithm is defined in [LIP 0003][lip3].
```
janhack, line 508: Sorry, I just noticed it now, but I would put the explicit code here because it is only a few lines and the specifications in LIP 0003 do not provide a clean interface with input parameters. The rather say the inputs are 

- randomSeed1 is the random seed for round r-1 as specified in LIP 0022,
- delegateList contains the address of each of the forging delegates for round r.
- 
This may be a bit misleading.

janhack (changes requested):
I left a few minor comments.

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: good point, forgot about the double use of this parameter. Fixed.

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: thanks, went for the compact suggestion

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: yes, it should, the BLS key can't exist in the blockchain for this command to be valid. Good catch.

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: indeed we do, good catch, added.

janhack: @ManuGowda @ishantiw How is it for other commands, do we always have size checks for all paramters that are enforced via the schema?

So here we are sayting that PopVerify would also include the size checks and the call to this function is also a (stateless) validation. Note that in `registerValidatorKeys` the PoP is verified again so I am not sure that is as intended.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+                  "address": {
+                      "dataType": "bytes",
+                      "fieldNumber": 1
+                  }
+                  "weight": {
+                      "dataType": "bytes",
+                      "fieldNumber": 2
+                  }
+	      }
+	  }
```
janhack, line 381: The indentation above is also incorrect. Also in this LIP sometimes the indentation is 3 spaces and sometimes 4. Not sure which editor you use, but I would recommend to set the indentation width and type in the editor (4 spaces).

proposals/lip-Introduce_PoA_module.md
janhack: Cbr.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+        validator.address = address
+        validator.roundHash = hash(randomSeed | address)
+        push validator to validatorsList
+    
+    # Reorder the validator list in lexicographical order first by roundHash, then by address
+    for every validator1 and validator2 in validatorsList
+        (validator1.roundHash < validator2.roundHash) || 
+        ((validator1.roundHash == validator2.roundHash) && (validator1.address < validator2.address))
+    
+    return shuffledAddresses = [address for every address in validatorsList]
```
janhack, line 523: I think an assignment is not necessary if you want to return the value on the right hand.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+
+##### Execution
+
+```python
+shuffleValidatorsList(validatorsAddresses, randomSeed)
+    
+    validatorsList = []
+    for every address in validatorsAddresses
+        validator.address = address
+        validator.roundHash = hash(randomSeed | address)
```
janhack, line 515: It could be good to mention after the code that `hash` is SHA256 and `|` is byte concatenation.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+##### Returns
+This function returns an array of bytes with the re-ordered list of addresses.
+
+##### Execution
+
+```python
+shuffleValidatorsList(validatorsAddresses, randomSeed)
+    
+    validatorsList = []
+    for every address in validatorsAddresses
```
janhack, line 513:
```suggestion
    for every address in validatorsAddresses:
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+        validator.address = address
+        validator.roundHash = hash(randomSeed | address)
+        push validator to validatorsList
+    
+    # Reorder the validator list in lexicographical order first by roundHash, then by address
+    for every validator1 and validator2 in validatorsList
+        (validator1.roundHash < validator2.roundHash) || 
+        ((validator1.roundHash == validator2.roundHash) && (validator1.address < validator2.address))
```
janhack, line 521: I find the lines above confusing. You would normally sort with a custom comparator function in this case. For the pseudo-code, I would maybe just write:

```
sort validatorsList in increasing order where validator1 < validator2 if 
    (validator1.roundHash < validator2.roundHash) || 
    ((validator1.roundHash == validator2.roundHash) && (validator1.address < validator2.address))
```

proposals/lip-Introduce_PoA_module.md
janhack: Can be resolved as the code is now added.

proposals/lip-Introduce_PoA_module.md
janhack: There are also some other bugs in the schema:
- missing quotation mark in `"newValidators: `
- missing comma after the address property

It would suggest to copy schema to a file `schema.json` and run
```
 prettier --write --tab-width 4 schema.json 
```

janhack (changes requested):
There are still a few minor open points left.

ishantiw (approved):
LGTM, just a minor comment

proposals/lip-Introduce_PoA_module.md
IkerAlus: true, removed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed, and added sentence at the end

proposals/lip-Introduce_PoA_module.md
IkerAlus: thanks for the good suggestion, updated

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,614 @@
...
+                       "dataType": "bytes",
+                       "fieldNumber": 1
+                   },
+                   "weight": {
+                       "dataType": "uint64",
+                       "fieldNumber": 2
+                   }
+	       }
+	   }
+       },
```
janhack, line 191: I think the two lines above in this schema are not formatted correctly.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,614 @@
...
+    validatorsList = []
+    for every address in validatorsAddresses:
+        validator.address = address
+        validator.roundHash = hash(randomSeed || address)
+        push validator to validatorsList
+    
+    # Reorder the validator list in lexicographical order first by roundHash, then by address
+    sort validatorsList in increasing order where validator1 < validator2 if 
+    (validator1.roundHash < validator2.roundHash) || 
+    ((validator1.roundHash == validator2.roundHash) && (validator1.address < validator2.address))
```
janhack, line 510: Lines 509 and 510 should be indented by 4 spaces, as they belong to the statement in line 508.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,614 @@
...
+    
+    return [address for every address in validatorsList]
```
janhack, line 512:
```suggestion
    return [validator.address for every validator in validatorsList]
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,614 @@
...
+    
+    validatorsList = []
+    for every address in validatorsAddresses:
+        validator.address = address
+        validator.roundHash = hash(randomSeed || address)
+        push validator to validatorsList
+    
+    # Reorder the validator list in lexicographical order first by roundHash, then by address
+    sort validatorsList in increasing order where validator1 < validator2 if 
+    (validator1.roundHash < validator2.roundHash) ||
```
janhack, line 509:
```suggestion
        (validator1.roundHash < validator2.roundHash) or
```

proposals/lip-Introduce_PoA_module.md
janhack, line 510:
```suggestion
        ((validator1.roundHash == validator2.roundHash) and (validator1.address < validator2.address))
```

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,614 @@
...
+        push validator to validatorsList
+    
+    # Reorder the validator list in lexicographical order first by roundHash, then by address
+    sort validatorsList in increasing order where validator1 < validator2 if 
+    (validator1.roundHash < validator2.roundHash) || 
+    ((validator1.roundHash == validator2.roundHash) && (validator1.address < validator2.address))
+    
+    return [address for every address in validatorsList]
+```
+where the `hash` function is SHA256 and `||` is byte concatenation.
```
janhack, line 514:
```suggestion
where `hash` is SHA256 and `||` is byte concatenation.
```

janhack: `hash` is the name of the function, we do not talk about the concept of a hash function here

proposals/lip-Introduce_PoA_module.md
janhack: I think the LIP has not been updated regarding this comment. In particular, when processing an authority registration command we would verify the PoP twice:
- once during the verification here
- again in the registerValidatorKeys API call

janhack (approved):
Note that there is also an older unresolved comment https://github.com/LiskHQ/lips/pull/100#discussion_r701719195.

janhack (changes requested):
There are a few remaining open comments.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,624 @@
...
+}
+```
+
+##### Properties and Default Values
+
+The properties of this schema are as follows:
+
+*   `roundEndHeight`: An integer stating the last height of the round. 
+Its initial value is set after the execution of the genesis block.
+*   `validatorsUpdateNonce`: An integer representing the number of times that the validator set has been updated with an update auhtority command.
```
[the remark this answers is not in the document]
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: updated

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed

proposals/lip-Introduce_PoA_module.md
IkerAlus: changed

proposals/lip-Introduce_PoA_module.md
IkerAlus: fixed with prettier

proposals/lip-Introduce_PoA_module.md
IkerAlus: I updated the validation of the command to just check that `registerValidatorKeys` does not fail. 
Also added the size checks for every input.

janhack: I think `registerValidatorKeys` cannot be called in the transaction verification because it induces a state transition. Also, it would now be called twice as the transaction verification also happens when executing the payload.

IkerAlus: removed the call of `registerValidatorKeys` from the verification step. Now it is only called in the execution step.

janhack: We still need to check the return value of that function. If it returns `False` the execution must fail.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -493,20 +492,17 @@ This function returns an array of bytes with the re-ordered list of addresses.
...
+    roundHash = {}
+    for address in validatorsAddresses:
+        roundHash[address] = hash(randomSeed || address)
     
     # Reorder the validator list in lexicographical order first by roundHash, then by address
-    sort validatorsList in increasing order where validator1 < validator2 if 
-        (validator1.roundHash < validator2.roundHash) or 
-        ((validator1.roundHash == validator2.roundHash) and (validator1.address < validator2.address))
+    shuffledValidatorAddresses = address in validatorsAddresses reordered by roundHash[address] 
+                                 ties broken lexicographically by address
```
janhack, line 502: This code is a bit odd and not very precise.
- validatorAddresses is already an array of addresses, so I don't understand why we write `address in validatorsAddresses`
- it does not say whether the sorting is increasing/decreasing
Suggestion:
```
  sort validatorsAddresses lexicogarphically by roundHash[address] breaking ties lexicographically by address  
```
or more specific
```
sort validatorsAddresses where address1 < adress2 if  (roundHash(address1) < roundHash(address2)) or 
    (( roundHash(address1) == roundHash(address2)) and address1 < address2)
```

janhack (changes requested):
2 comments left

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,606 @@
...
+        *   Minimum value: ⌊ ⅓ × `totalWeight`⌋+ 1
+        *   Maximum value: `totalWeight`
+
+       where ⌊⋅⌋ is the floor function.
+
+*   Rules for `trs.params.validatorsUpdateNonce` property:
+    *   The value of `trs.params.validatorsUpdateNonce` has to be equal to `chainProperties.validatorsUpdateNonce`.
+*   Rules for `trs.params.aggregationBits` and `trs.params.signature` properties:
+    *   The function `verifyWeightedAggSig(keyList, tag, netID, aggregationBits, signature, m, weights, threshold)`, specified in [LIP 0038][BLS], must return VALID where:
+        *   The `keyList` property is an array containing `getValidatorAccount(address).blsKey` for every `address` property in `snapshotStore(0).validators` array sorted in the same order, where `getValidatorAccount` is the function exposed by the [validators module][validatorsModule].
```
janhack, line 424: Comparing with the mainchain registration, maybe we also want to sort the keys and weights array lexicographically by BLS key here, see https://github.com/LiskHQ/lips/pull/109#discussion_r709184436.

proposals/lip-Introduce_PoA_module.md
```diff
@@ -0,0 +1,606 @@
...
+The properties of this schema are as follows:
+
+*   `roundEndHeight`: An integer stating the last height of the round. 
+Its initial value is set after the execution of the genesis block.
+*   `validatorsUpdateNonce`: An integer representing the number of times that the validator set has been updated with an update authority command. 
+It is initialized to 0.
+
+### Commands
+
+#### Authority Registration Command
```
janhack, line 252: Should we consider having a fee for this registration similar to DPoS with the same rational?

proposals/lip-Introduce_PoA_module.md
IkerAlus: good point, added

proposals/lip-Introduce_PoA_module.md
IkerAlus: good point, added for completeness.

proposals/lip-Introduce_PoA_module.md
IkerAlus: added

proposals/lip-Introduce_PoA_module.md
IkerAlus: Changed as suggested. 

I prefer to return a different variable just to avoid side effects with input modification.

janhack: Looks good. Can be resolved.

proposals/lip-Introduce_PoA_module.md
janhack: Looks good. Can be resolved.

janhack (approved):
Looks all good.

karmacoma (approved):

PR #112, opened 2021-09-22
Add LIP "Define state and state transitions of DPoS module" 

MaximeGagnebin:


ManuGowda (approved):
Great job @MaximeGagnebin @mehmetegemen

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+        return True
+```
+
+```python
+isPunished(unlockObject, senderAddress, height):
+    if delegateStore(unlockObject.delegateAddress).pomHeights is empty:
+        return false
+    else:
+        let lastPomHeight be the last element of delegateStore(unlockObject.delegateAddress).pomHeights 
+        # lastPomHeight is also its largest element of the pomHeights array
```
ricott1, line 728:
```suggestion
        # lastPomHeight is also the largest element of the pomHeights array
```

Otherwise I'm not sure I understand the comment

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+for each unlockObject in voterStore(senderAddress).pendingUnlocks:
+    if (hasWaited(unlockObject, senderAddress, height) 
+        and not isPunished(unlockObject, senderAddress, height)
+        and isCertificateGenerated(unlockObject)):
+        delete unlockObject from voterStore(senderAddress).pendingUnlocks
+        token.unlock(senderAddress, MODULE_ID_DPOS, TOKEN_ID_DPOS, unlockObject.amount)
+```
+
+The definition and rationale for the `isCertificateGenerated` function is part of [LIP "Introduce unlocking condition for incentivizing certificate generation"][LIP-incentivizeCertificateGeneration]. 
+The `hasWaited` and `isPunished` functions are defined below and are rationalized in [LIP 0023][LIP-0023-unlockRationale] and [LIP 0024][LIP-0024-rationale] respectively.
```
ricott1, line 707: Maybe it would be nice to still include the description of the input parameters, as it looks to me that the interface changed (I guess because it is easier with the new architecture?)

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+isPunished(unlockObject, senderAddress, height):
+    if delegateStore(unlockObject.delegateAddress).pomHeights is empty:
+        return false
+    else:
+        let lastPomHeight be the last element of delegateStore(unlockObject.delegateAddress).pomHeights 
+        # lastPomHeight is also its largest element of the pomHeights array
+        
+        if unlockObject.address == senderAddress:
+            # this is a self-unvote
+            if height – lastPomHeight < 780,000 and lastPomHeight < unlockObject.unvoteHeight + 260,000:
```
ricott1, line 732: couldnt this return true for `height < lastPomHeight`? I understand the normal use would be to set height to the block height at which the unlocking transaction is included, but there is no check that, say `height > unlockObject.unvoteHeight`

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+
+```python
+shuffleValidatorsList(validatorsAddresses, randomSeed):
+    
+    roundHash = {}
+    for address in validatorsAddresses:
+        roundHash[address] = hash(randomSeed || address)
+
+    # Reorder the validator list
+    shuffledValidatorAddresses = sort validatorsAddresses where address1 < adress2 if (roundHash(address1) < roundHash(address2))
```
ricott1, line 877:
```suggestion
    shuffledValidatorAddresses = sort validatorsAddresses where address1 < address2 if (roundHash(address1) < roundHash(address2))
```

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+missedBlocks[validators.getGeneratorAtTimestamp(previousTimestamp)] -= 1
+missedBlocks[validators.getGeneratorAtTimestamp(newBlock.header.timestamp)] -= 1
+
+for address in missedBlocks:
+    delegateStore(address).consecutiveMissedBlocks += missedBlocks[address]
+
+    # the rule below was introduced in LIP 0023
+    if (delegateStore(address).consecutiveMissedBlocks > FAIL_SAFE_MISSED_BLOCKS
+        and newHeight - delegateStore(address).lastGeneratedHeight > FAIL_SAFE_INACTIVE_WINDOW):
+        delegateStore(address).isBanned = True
```
ricott1, line 937: question: basically u can still get banned after generating the block (if in the meanwhile you missed a lot of blocks). Is the block still valid in this case?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+roundNumber = roundNumber(b.height)
+currentWeights = {}
+for address being a storeKey in delegate substore and isBanned(address) == False:
+    currentWeights[address] = delegateWeight(address, b.height)} 
+
+activeDelegates = array of the top 101 address by decreasing delegateWeight from currentWeights, ties broken by lexicographical ordering of the address
+
+# if currentWeights contains less than 101 entries
+# there will be less than 101 activeDelegates
+remove all entries from currentWeights with address in activeDelegates
```
ricott1, line 957: no need to change it necessarily, but here you could also write a simple for loop

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+
+sort currentWeights by decreasing delegateWeight, ties broken by lexicographical order of the delegateAddress
```
ricott1, line 959: Maybe not important, but in the implementation I suppose we would need an extra array (as maps are not ordered)?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+
+##### Returns
+
+This functions returns `delegateStore(address)` deserialized using `delegateStoreSchema`.
+
+
+### Endpoints for Off-Chain Services
+
+
+#### getVoter(address)
```
ricott1, line 1093:
```suggestion
#### getVoter
```

Not sure why we put the parameter in the header here. Same for other off-chain functions

ricott1 (changes requested):

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: @IkerAlus you might want to fix this too

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: yes

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: Maybe, this is up to the dev team.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: removed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: I'm not sure that I see your concern. There is indeed not check that `height > unlockObject.unvoteHeight`, but the `unvoteHeight` is set when you unvote, which must happen before you unlock. Otherwise, the unlockObject does not exist.
Maybe the implementation can add checks for those non-existent cases, but I don't think they need to be in the LIP.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: changed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1160 @@
...
+[LIP-0023-unlockRationale]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md#explicit-unlock-mechanism
+[LIP-0003]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0003.md
+[LIP-0024-applyingPOM]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#applying-a-pom-transaction
+[LIP-0024-verifyPOM]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#validity-of-a-pom-transaction
+[LIP-0024-rationale]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#rationale
+[LIP-0034-bootstrapPeriod]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0034.md#bootstrap-period
+
+[LIP-newBlockHeader]: https://research.lisk.com/t/new-block-header-and-block-asset-schema/293
+[LIP-incentivizeCertificateGeneration]: https://research.lisk.com/t/introduce-unlocking-condition-for-incentivizing-certificate-generation/300
+[LIP-validators]: https://research.lisk.com/t/introduce-validators-module/317
```
IkerAlus, line 1164: This link can be updated to the LIP repo now.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: did not change.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1164 @@
...
+        and newHeight - delegateStore(address).lastGeneratedHeight > FAIL_SAFE_INACTIVE_WINDOW):
+        delegateStore(address).isBanned = True
+
+delegateStore(newBlock.header.generatorAddress).consecutiveMissedBlocks = 0
+delegateStore(newBlock.header.generatorAddress).lastGeneratedHeight = newHeight
+
+previousTimestamp = newBlock.header.timestamp
+```
+
+After an end-of-round block `b` is executed (`isEndOfRound(b.height) == True`), the following logic is executed (this must be done after the properties related to missed blocks are updated):
```
IkerAlus, line 949: To be consistent with line 898 (and strictly correct with block header specs), here we  should write `b.header.height`. The same issue happens in several other lines of the document starting from line 576.

IkerAlus (changes requested):
A couple of little fixes, the rest looks good.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
ricott1: ah sorry, I somehow missed this is only an internal function at this point

ricott1 (approved):
all good, bravo

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: indeed done.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: indeed, changed everywhere.

IkerAlus (approved):
LGTM, great work!

mehmetegemen (approved):
Great work!

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1163 @@
...
+previousTimestamp = newBlock.header.timestamp
+```
+
+After an end-of-round block `b` is executed (`isEndOfRound(b.header.height) == True`), the following logic is executed (this must be done after the properties related to missed blocks are updated):
+
+```python
+roundNumber = roundNumber(b.header.height)
+currentWeights = {}
+for address being a storeKey in delegate substore and isBanned(address) == False:
+    currentWeights[address] = delegateWeight(address, b.header.height)}
```
mehmetegemen, line 955: `delegateWeight(address, b.header.height)}`, what is this lonely curly bracket is doing there 🙂 ?

MaximeGagnebin: chilling?
good catch, I removed it 🙂.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1176 @@
...
+
+previousTimestamp = newBlock.header.timestamp
+```
+
+After an end-of-round block `b` is executed (`isEndOfRound(b.header.height) == True`), the following logic is executed (this must be done after the properties related to missed blocks are updated):
+
+```python
+roundNumber = roundNumber(b.header.height)
+currentWeights = {}
+for address being a storeKey in delegate substore and isBanned(address) == False:
```
AndreasKendziorra, line 967: just stumbled over it: `isBanned` is a property, but not a function. So, it should be `delegateStore(address).isBanned == False`.

MaximeGagnebin: Thanks, fixed.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+```
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the DPoS module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisDPoSStoreSchema` schema, given above.
+
+* Initial checks on the properties of `genesisBlockAssetObject`:
```
IkerAlus, line 1090: should we add a check for the maximum length of the `pomHeights` array?

IkerAlus (changes requested):
Some fixes and improvements in the notation regarding the **Genesis Block Processing** section

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+genesisDPoSStoreSchema = {
+    "type": "object",
+    "required": ["validators", "voters", "snapshots", "bootstrapPeriod"],
+    "properties": {
+        "validators": {
+            "type": "array",
+            "fieldNumber": 1,
+            "items": {
+                "type": "object",
+                "required": [
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, added

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+                                    "dataType": "uint64",
+                                    "fieldNumber": 2
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "GenesisData": {
```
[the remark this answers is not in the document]
MaximeGagnebin: fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+        "delegateWeightSnapshot": snapshot.delegateWeightSnapshot
+    } serialized using snapshotStoreSchema.
+    ```
+
+* Create an entry in the genesis data substore with
+    ```python
+    storeKey = EMPTY_BYTES
+    storeValue = {
+        "height": block header height of the genesis block,
+        "initRounds": genesisBlockAssetObject.GenesisData.initRounds,
```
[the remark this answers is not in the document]
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+# set the initial delegates in the BFT module
+bftWeights = [
+    {"address": address, "bftWeight": 1} 
+    for address in genesisBlockAssetObject.GensisData.initDelegates
+    sorted by lexicographically by address
+]
+
+# compute the initial BFT threshold 
+initBFTThreshold = floor(2/3 * length(genesisBlockAssetObject.GensisData.initDelegates)) + 1
```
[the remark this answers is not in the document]
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+    for pendingUnlock in voter(address).pendingUnlocks:
+        votedAmount += pendingUnlock.amount
+
+    if getLockedAmount(address, MODULE_ID_DPOS, TOKEN_ID_DPOS) < votedAmount:
+        fail
+
+# set the initial delegates in the BFT module
+bftWeights = [
+    {"address": address, "bftWeight": 1} 
+    for address in genesisBlockAssetObject.GensisData.initDelegates
```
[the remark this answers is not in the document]
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+# compute the initial BFT threshold 
+initBFTThreshold = floor(2/3 * length(genesisBlockAssetObject.GensisData.initDelegates)) + 1
+
+setBFTParameters(initBFTThreshold,
+                     initBFTThreshold,
+                     bftWeights)
+                     
+# set the inititial delegates in the validators module                 
+setGeneratorList(initDelegates)
```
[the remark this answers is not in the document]
MaximeGagnebin: defined `initDelegates` properly.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+    ```
+
+* Create an entry in the genesis data substore with
+    ```python
+    storeKey = EMPTY_BYTES
+    storeValue = {
+        "height": block header height of the genesis block,
+        "initRounds": genesisBlockAssetObject.GenesisData.initRounds,
+        "initDelegates": genesisBlockAssetObject.GenesisData.initDelegates
+    } serialized using availableLocalIDStoreSchema.
```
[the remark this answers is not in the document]
MaximeGagnebin: fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+#### Genesis State Finalization
+
+To finalize the state of the genesis block the following logic is executed.  If any step fails, the block is discarded and has no further effect.
+
+As in the previous point, let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the DPoS module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisDPoSStoreSchema` schema, given above. Then:
+
+```python
+# register all validators in the Validators module
+for validator in genesisBlockAssetObject.validators:
+    registerValidatorKeys(
```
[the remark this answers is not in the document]
MaximeGagnebin: fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+### Genesis Block Processing
+
+
+#### Genesis Assets Schema
+
+```java
+genesisDPoSStoreSchema = {
+    "type": "object",
+    "required": ["validators", "voters", "snapshots", "bootstrapPeriod"],
```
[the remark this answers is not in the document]
MaximeGagnebin: fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+        * Accross elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`. 
+        * `sentVotes` has size is at most `MAX_NUMBER_SENT_VOTES`.
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`. 
+    * Accross elements of the `snapshots` array, `roundNumber` values must be unique.
+    * Accross elements of the `snapshots` array, the `activeDelegates` array must only contain unique entries, all of them must have length `ADDRESS_LENGTH` and they must be in lexicographic order.
+    * Accross elements of the `snapshots` array, the `delegateWeightSnapshot` array must only contain unique entries for the `delegateAddress` property and all `delegateAddress` property must have length `ADDRESS_LENGTH`.
+    * All values of the `bootstrapPeriod.initDelegate` array must be unique, must have length `ADDRESS_LENGTH` and must be in lexicographic order.
```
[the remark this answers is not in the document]
MaximeGagnebin: all fixed

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+* For each entry `validator` in `genesisBlockAssetObject.validators`, create an entry in the delegate substore with
+    ```python
+    totalVotesReceived = 0
```
ricott1, line 1112: How do we check that the total amounts of votes is not more than max unit 64?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+# check that all sentVotes and pendingUnlocks correspond to locked tokens
+for address a key of the voter substore:
+    votedAmount = 0
+    for sentVote in voter(address).sentVotes:
+        votedAmount += sentVote.amount
+    for pendingUnlock in voter(address).pendingUnlocks:
+        votedAmount += pendingUnlock.amount
+
+    if getLockedAmount(address, MODULE_ID_DPOS, TOKEN_ID_DPOS) < votedAmount:
```
ricott1, line 1191: mh, shouldn't we check for an exact match?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+# register all validators in the Validators module
+for validator in genesisBlockAssetObject.validators:
+    registerValidatorKeys(
+        validator.address,
+        validator.proofOfPossession,
+        validator.generatorKey,
+        validator.blsKey
+    )
+
+# check that all sentVotes and pendingUnlocks correspond to locked tokens
```
ricott1, line 1183: As I said in person, another option could be to instead call the lock API in the finalization step. That would avoid this extra check. I guess here it is simple enough that we could do it both ways, so it's more of a consistency thing. Another advantage of calling the API maybe is that if other modules also try to lock tokens you immediately see if the total amount of locked tokens is greater than the amount of tokens (not sure if you do this "global" check in the token module genesis processing)

ricott1 (changes requested):
Some small comments and a general comment about using the lock API

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+    * Accross elements of the `validators` array, all `address` values must be unique, all `name` values must also be unique.
+    * For all elements of the `validators` array:
+        * `address` values must have length `ADDRESS_LENGTH`.
+        * `name` values must have length between `1` and `MAX_LENGTH_NAME` (included), and must contain only characters from the set `abcdefghijklmnopqrstuvwxyz0123456789!@$&_.`.
+        * `generatorKey` values must have length 32.
+        * `blsKey` values must have length 48.
+        * `proofOfPossession` values must have length 96.
+    * Accross elements of the `voters` array, all `address` values must be unique, and must have length `ADDRESS_LENGTH`.
+    * For all elements of the `voters` array:
+        * Accross elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`.
```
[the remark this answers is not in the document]
MaximeGagnebin: added the check

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+        * `proofOfPossession` values must have length 96.
+    * Accross elements of the `voters` array, all `address` values must be unique, and must have length `ADDRESS_LENGTH`.
+    * For all elements of the `voters` array:
+        * Accross elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`. 
+        * `sentVotes` has size is at most `MAX_NUMBER_SENT_VOTES`.
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`. 
+    * Accross elements of the `snapshots` array, `roundNumber` values must be unique.
+    * Accross elements of the `snapshots` array, the `activeDelegates` array must only contain unique entries, all of them must have length `ADDRESS_LENGTH` and they must be in lexicographic order.
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, added the checks.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: I thought about it, and then concluded that it is not needed. If the protocol defines a limit, then that delegate will be banned when they reach the limit, but that's it.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+        * `blsKey` values must have length 48.
+        * `proofOfPossession` values must have length 96.
+    * Accross elements of the `voters` array, all `address` values must be unique, and must have length `ADDRESS_LENGTH`.
+    * For all elements of the `voters` array:
+        * Accross elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`. 
+        * `sentVotes` has size is at most `MAX_NUMBER_SENT_VOTES`.
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`. 
+    * Accross elements of the `snapshots` array, `roundNumber` values must be unique.
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, changed.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+
+# check that the snapshot only correspond to the last three rounds
+let h be the header height of the genesis block
+genesisRound = roundNumber(h)
+snapshotKeys = store keys (converted to uint32) of the snapshot substore ordered increasingly
+if snapshotKeys != [genesisRound-2, genesisRound-1, genesisRound]:
+    fail
+```
+
+where:
```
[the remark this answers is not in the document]
MaximeGagnebin: good point, changed.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+                     bftWeights)
+                     
+# set the inititial delegates in the validators module                 
+setGeneratorList(initDelegates)
+
+# check that the snapshot only correspond to the last three rounds
+let h be the header height of the genesis block
+genesisRound = roundNumber(h)
+snapshotKeys = store keys (converted to uint32) of the snapshot substore ordered increasingly
+if snapshotKeys != [genesisRound-2, genesisRound-1, genesisRound]:
```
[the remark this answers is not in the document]
MaximeGagnebin: changed the wording.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1472 @@
...
+```
+
+
+### Genesis Block Processing
+
+
+#### Genesis Assets Schema
+
+```java
+genesisDPoSStoreSchema = {
```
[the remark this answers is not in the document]
MaximeGagnebin: I prefer to leave it as is.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: added the check.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: I slightly prefer the way it is written now, as it is less forceful on developers. 

Also, the tokens are locked per module. So here, we compare the votes to tokens locked for the DPoS module.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: I would not. This is not a protocol issue and leaves more freedom to developers.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the DPoS module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisDPoSStoreSchema` schema, given above.
+
+* Initial checks on the properties of `genesisBlockAssetObject`:
+    * Accross elements of the `validators` array, all `address` values must be unique, all `name` values must also be unique.
```
mehmetegemen, line 1095: Did you mean `Across` instead of `Accross`?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
ricott1: maybe I'm missing something, but what kind of freedom are you thinking of? If they want to lock more tokens under the DPoS module they would need to modify the DPoS module, since AFAIK the only locking reason in DPoS is for votes. I think we should be tight when possible

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
ricott1: What do you mean with 'less forceful'? We are imposing the same constraint, just expressed differently no?

yes, that was clear to me. But there is another 'global' constraint (global meaning that it involves possibly many modules): the total locked tokens must be lower than maxUint64. Is this checked in the token module independently?

MaximeGagnebin: yes, the token module checks that there exists less 2**64 of any token.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: Maybe @IkerAlus can break the tie here. I would rather not be tight unless needed.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: For me the problem really comes from the fact that the Token module is allowed to defined locked tokens, I think we settled that in a past meeting. So I would rather use that feature and only verify here that they are locked.
The other option would be to not allow any locked tokens in the Token module and lock them from the module where they are used.
Overall, I'm not strongly opinionated here, but I prefer the current way of doing it.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: yes

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
IkerAlus: good point, resolving

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
IkerAlus: If I was me I would specify it with equality. mainly because I do not see any use case for this.  But I do not have strong opinion, maybe it would be something interesting so we can leave it as it is.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
IkerAlus: Thinking about this, for me it looks more "adequate" or "aligned" with our chain architecture to do the locking via API call like Alessandro said. But no strong opinion either.

I just want to confirm that in my understanding the token module can define locked tokens in its genesis (for a given module ID) and at the same time the DPoS module can lock via API without it being an inconsistency, right?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -898,52 +898,351 @@ shuffleValidatorsList(validatorsAddresses, randomSeed):
...
+        },
+        "voters": {
+            "type": "array",
+            "fieldNumber": 2,
+            "items": {
+                "type": "object",
+                "required": ["address", "sentVotes", "pendingUnlocks"],
+                "properties": {
+                    "address": {
+                        "dataType": "uint32",
```
mehmetegemen, line 976: Did you mean `bytes`?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -898,52 +898,351 @@ shuffleValidatorsList(validatorsAddresses, randomSeed):
...
+        * `generatorKey` values must have length 32.
+        * `blsKey` values must have length 48.
+        * `proofOfPossession` values must have length 96.
+    * Across elements of the `voters` array, all `address` values must be unique, and must have length `ADDRESS_LENGTH`.
+    * For all elements of the `voters` array:
+        * Across elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`. 
+        * `sentVotes` has size is at most `MAX_NUMBER_SENT_VOTES`.
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`.
```
mehmetegemen, line 1108: I think we should indicate that these operations must be in order. This way it doesn't tell what tie is broken after what tie.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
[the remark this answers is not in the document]
IkerAlus: just to be sure, if we do `uint32be(X)` where `X` is a negative number, what would be the result? Probably we should specify clearly the behaviour so this cases is well-defined later in line 1159.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+    * For all elements of the `voters` array:
+        * Across elements of the `sentVotes` array, all `delegateAddress` values must be unique, and must have length `ADDRESS_LENGTH`. 
+        * `sentVotes` has size is at most `MAX_NUMBER_SENT_VOTES`.
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`. 
+    * The `snapshots` array must have length less than or equal to 3.
+    * Across elements of the `snapshots` array, `roundNumber` values must be unique.
+    * Across elements of the `snapshots` array, 
+        * All values of the `activeDelegates` array must be unique, must have length `ADDRESS_LENGTH`, and must be equal to `validator.address` for `validator` an element of `validators`.
```
IkerAlus, line 1112: Perhaps it is more clear to add a new bulletpoint like:

> - for every element `activeAddress` in the `activeDelegates` array there is an element `validator` in the `validators` array with `validator.address == activeAddress`.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+        * `sentVotes` must be in lexicographic order of `delegateAddress`.
+        * `pendingUnlocks` has size is at most `MAX_NUMBER_PENDING_UNLOCKS`.
+        * `pendingUnlocks` must be ordered by lexicographical order of `delegateAddress`, ties broken by increasing `amount`, ties broken by increasing `unvoteHeight`. 
+    * The `snapshots` array must have length less than or equal to 3.
+    * Across elements of the `snapshots` array, `roundNumber` values must be unique.
+    * Across elements of the `snapshots` array, 
+        * All values of the `activeDelegates` array must be unique, must have length `ADDRESS_LENGTH`, and must be equal to `validator.address` for `validator` an element of `validators`. 
+        * Across elements of the `delegateWeightSnapshot` array, all values for the `delegateAddress` property must be unique, must have length `ADDRESS_LENGTH`, and must be equal to `validator.address` for `validator` an element of `validators`.  
+    * All values of the `genesisData.initDelegates` array must be unique, must have length `ADDRESS_LENGTH`, and must be equal to `validator.address` for `validator` an element of `validators`.
```
IkerAlus, line 1114: same suggestion as in line 1112 for `genesisData.initDelegates` array

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+    * The `genesisData.initDelegates` array must have length less than `NUMBER_ACTIVE_DELEGATES`.
```
IkerAlus, line 1115: Less or equal?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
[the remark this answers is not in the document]
IkerAlus: small precision comment around it

IkerAlus (changes requested):
Small  behaviour comments and fixes

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
ricott1: I think that in general it is a bit awkward to impose no locked tokens only for specific modules (as you don't know from the token module perspective if these modules are registered or not). So I would say it would either be no locked tokens at all, or any locked tokens (meaning, for any module ID).

Thinking out loud, if we go for the API approach we would then still have the possibility of DPoS tokens being defined "independently" in the token module, unless we first check from DPoS that there are no tokens locked for it (and only then call the API).

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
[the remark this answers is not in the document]
MaximeGagnebin: So `uint32be` is undefined for negative numbers. My point is that here it does not matter. When creating the store entries, we always use the `roundNumbers` provided in the genesis asset (even if they are 12, 18 and 333).

Then in the code, we check that they are meaningful (i.e: what they have to be), but I don't think there is any problem with store initialization.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: yes

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+                     initBFTThreshold,
+                     bftWeights)
+                     
+# set the initial delegates in the validators module                 
+validators.setGeneratorList(initDelegates)
+
+# check that the snapshot only correspond to the last three rounds
+let h be the header height of the genesis block
+genesisRound = roundNumber(h)
+snapshotKeys = store keys (converted to uint32) of the snapshot substore ordered increasingly
```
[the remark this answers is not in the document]
MaximeGagnebin: added `array of` to be sure. (though I think the notation is pretty clear)

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+if snapshotKeys is not an incrementing array or snapshotKeys[-1] != genesisRound:
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, we are sure about the expected behavior. If `h==0` and the genesis asset contains entries for the snapshot store, then this will fail.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1477 @@
...
+* Initial checks on the properties of `genesisBlockAssetObject`:
+    * Across elements of the `validators` array, all `address` values must be unique, all `name` values must also be unique.
+    * For all elements of the `validators` array:
+        * `address` values must have length `ADDRESS_LENGTH`.
+        * `name` values must have length between `1` and `MAX_LENGTH_NAME` (included), and must contain only characters from the set `abcdefghijklmnopqrstuvwxyz0123456789!@$&_.`.
+        * `generatorKey` values must have length 32.
+        * `blsKey` values must have length 48.
+        * `proofOfPossession` values must have length 96.
+    * Across elements of the `voters` array, all `address` values must be unique, and must have length `ADDRESS_LENGTH`.
+    * For all elements of the `voters` array:
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, added

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: added

ricott1 (approved):
We clarified the last remaining points, all good from my side

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: that sounds good, updated

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: ok, I added a few words to the sentence to indicate the order of tie breaking, tell me if it's fine for you.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: We concluded to leave the  check as present (tokens locked in Token genesis asset and DPoS module only verifies the locking)

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: Changed to check for equality.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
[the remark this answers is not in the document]
IkerAlus: ok! resolved!

IkerAlus (approved):
LGTM!! great work

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1481 @@
...
+
+* Create an entry in the genesis data substore with
+    ```python
+    storeKey = EMPTY_BYTES
+    storeValue = {
+        "height": block header height of the genesis block,
+        "initRounds": genesisBlockAssetObject.genesisData.initRounds,
+        "initDelegates": genesisBlockAssetObject.genesisData.initDelegates
+    } serialized using genesisDataStoreSchema.
+    ```
```
shuse2, line 1178: initializing previous timestamp store is missing

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1481 @@
...
+
+# check that all sentVotes and pendingUnlocks correspond to locked tokens
+for address a key of the voter substore:
+    votedAmount = 0
+    for sentVote in voter(address).sentVotes:
+        votedAmount += sentVote.amount
+    for pendingUnlock in voter(address).pendingUnlocks:
+        votedAmount += pendingUnlock.amount
+
+    if token.getLockedAmount(address, MODULE_ID_DPOS, TOKEN_ID_DPOS) != votedAmount:
```
shuse2, line 1206: for validator/bft, it is actively initializing, but for token module, it expects token module to initialize. Is it intentional?

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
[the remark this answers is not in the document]
shuse2: maybe it's good to have this in the comment

> If h==0 and the genesis asset contains entries for the snapshot store, then this will fail.

shuse2 (changes requested):

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
```diff
@@ -0,0 +1,1481 @@
...
+    ```python
+    storeKey = validator.name utf8-encoded
+    storeValue = {
+        "delegateAddress": validator.address
+    } serialized using nameStoreSchema.
+    ```
+
+* For each entry `voter` in `genesisBlockAssetObject.voters`, create an entry in the voter substore  with
+    ```python
+    storeKey = uint32be(voter.address)
```
shuse2, line 1154:
```suggestion
    storeKey = voter.address
```

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: yes, we did it this way intentionally, but we had arguments going both ways. One of the reasons for doing this way was: 
- in Token module, you can set locked tokens in the genesis state. So if we directly call `lock` here, you run into the scenario where the genesis state could lock tokens twice, once in Toke module directly, once from here. This is not possible in Validators or BFT, because they always start with an empty state.
- this also allows us to check for equality, and hence be sure that exactly the right amount of tokens are locked. 
- I found it more natural when checking the genesis state that you would see the locked tokens, and not see only available balance (which gets locked during processing).

If you think we should go for the other way, it would also be fine for me (it was also Alessandro's preference). I would just change the name of the function here.

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: done

proposals/lip_define_state_and_state_transitions_of_dpos_module.md
MaximeGagnebin: yes, added.

shuse2 (approved):

mehmetegemen (approved):

karmacoma (approved):

PR #117, opened 2021-11-01
Update LIP 46 with genesis block processing

IkerAlus:


proposals/lip-0046.md
```diff
@@ -242,26 +242,32 @@ The Random module exposes the following function.
...
+### Genesis Block Processing
 
-#### After Genesis Block Execution
+The following step is executed as part of the genesis block processing, see the “Update genesis block schema and processing” LIP for details.
+
+#### Genesis State Initialization
 
 After the genesis block `g` is executed, the following logic is executed:
 
 * Set the `validatorReveals` array to an empty array.
```
ricott1, line 253: I would clarify that this array is then stored in the random substore (basically, we need to say that the store is initialized to this default value).

ricott1 (changes requested):
A minor clarification

MaximeGagnebin (approved):

proposals/lip-0046.md
IkerAlus: Good point, added a point with the substore creation

ricott1 (approved):

karmacoma (approved):

PR #118, opened 2021-11-02
LIP0041: Add genesis state processing 

ricott1:


proposals/lip-0041.md
```diff
@@ -216,3 +295,5 @@ authAccount(senderAddress).nonce += 1
 ## Backwards Compatibility
 
 This LIP defines a new storage interface for the Auth module, which in turn will become part of the state tree and will be authenticated by the state root. An existing chain including the Auth module will need to perform a hardfork.
+
+[research:genesis-block]: https://research.lisk.com/t/update-genesis-block-schema-and-processing/325
```
IkerAlus, line 299: I don't see this link being used in the doc body, what was it expected to be used somwhere?

proposals/lip-0041.md
```diff
@@ -170,6 +167,88 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
 
 This function returns `true` if the transaction object contains a valid nonce, `false` otherwise. This is done using the `isValidNonce` function defined [below](#transaction-verification).
 
+### Genesis Block Processing
+
+The following two steps are executed as part of the genesis block processing.
```
IkerAlus, line 172: it is a bit weird to talk about both processing steps when there is only one which is relevant to this module.

proposals/lip-0041.md
```diff
@@ -170,6 +167,88 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+            "storeValue": {
+              "type": "object",
+              "required": [
+                "nonce",
+                "numberOfSignatures",
+                "mandatoryKeys",
+                "optionalKeys"
+              ],
+              "properties": {
+                "nonce": {
```
IkerAlus, line 216: so we are not imposing any rule on the nonce value. 

Is it not better to avoid to force `nonce == 0` (or set by default when creating substores) for every account so that devs have less room to mess with the initial state (e.g., a very large initial nonce, or a nonce bigger than `uint64`)? 
Is there any reason for the current specification apart from just having a lighter set of rules for the genesis state initialization?

IkerAlus (commented):
LGTM, just some comments to consider that may or may not imply changes.

proposals/lip-0041.md
ricott1: I had a sentence to link the genesis block LIP but later removed so that we don't have to wait. I forgot to remove the link.

proposals/lip-0041.md
```diff
@@ -170,6 +167,88 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+    "authDataSubstore": {
+      "type": "object",
+      "required": ["storePrefix", "storeEntries"],
+      "properties": {
+        "storePrefix": {
+          "dataType": "bytes"
+        },
+        "storeEntries": {
+          "type": "array",
+          "items": {
```
MaximeGagnebin, line 205: I think you are missing an `object` nesting here. 
This way of writing the schemas would be nice, and I think I took that shortcut int the discussion document, but I don't think we can actually define the `items` this way. :(

proposals/lip-0041.md
```diff
@@ -170,6 +167,88 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+            "storeKey": "bytes",
```
MaximeGagnebin, line 206: I think we're missing something like `{"dataType": "bytes", "fieldNumber": 1}` here.

proposals/lip-0041.md
```diff
@@ -170,6 +167,88 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+            "storeValue": {
+              "type": "object",
```
MaximeGagnebin, line 208: Think we are missing the field Number for `storeValue`.

MaximeGagnebin (changes requested):
Left a few schema related comments.

proposals/lip-0041.md
ricott1: edited

proposals/lip-0041.md
ricott1: yep, I mindlessly copied from the doc, should have double checked.. Thanks!

proposals/lip-0041.md
ricott1: fixed

proposals/lip-0041.md
ricott1: added layer

MaximeGagnebin (approved):

proposals/lip-0041.md
```diff
@@ -170,6 +167,96 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+                      "dataType": "bytes"
+                    },
+                    "fieldNumber": 4
+                  }
+                },
+                "fieldNumber": 2
+              }
+            }
+          },
+          "fieldNumber": 2
```
ishantiw, line 248: maybe fieldNumber can come as first prop so that it improves readability

IkerAlus (approved):
LGTM

proposals/lip-0041.md
ricott1: I would still keep it at the end as this is the template we chose to follow.

ricott1: nevermind, I was wrong. The `fieldNumber` should go second as you suggested, updated

ishantiw (approved):

proposals/lip-0041.md
```diff
@@ -175,15 +175,14 @@ During the genesis state initialization stage, the following steps are executed.
...
 
-* Check that `genesisBlockAssetObject.authDataSubstore.storePrefix` equals `STORE_PREFIX_AUTH`.
-* Check that every entry `authAccount` in `genesisBlockAssetObject.authDataSubstore.storeEntries` has a different `authAccount.storeKey`.
-* For each entry `authAccount` in `genesisBlockAssetObject.authDataSubstore.storeEntries`, if `authAccount.storeValue.numberOfSignatures` is not 0 (a multisignature account), check that the rules listed in [LIP 0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md#multisignature-account-registration) are respected. In particular:
-  * The keys in the `authAccount.storeValue.mandatoryKeys` array are pairwise distinct and ordered lexicographically. The array may be empty.
-  * The keys in the `authAccount.storeValue.optionalKeys` array are pairwise distinct and ordered lexicographically. The array may be empty.
+* Check that every entry `authAccount` in `genesisBlockAssetObject.authDataSubstore` has a different `authAccount.storeKey`.
```
MaximeGagnebin, line 178: "and that all `authAccount.storeKey` have length 20."  ?

proposals/lip-0041.md
```diff
@@ -175,15 +175,14 @@ During the genesis state initialization stage, the following steps are executed.
...
+* For each entry `authAccount` in `genesisBlockAssetObject.authDataSubstore`, if `authAccount.storeValue.numberOfSignatures` is not 0 (a multisignature account), check that the rules listed in [LIP 0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md#multisignature-account-registration) are respected. In particular:
+  * The keys in the `authAccount.storeValue.mandatoryKeys` array are pairwise distinct and in lexicographical order. The array may be empty.
```
MaximeGagnebin, line 180: Mentioned in LIP17, but if you wanted, we could add: "All elements of `mandatoryKeys` and `optionalKeys` have length 32."

MaximeGagnebin (approved):
Looks good, just left 2 nitpicky comment about checking the length of storeKey and of elements of the keys arrays.

proposals/lip-0041.md
```diff
@@ -170,6 +167,85 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
 
 This function returns `true` if the transaction object contains a valid nonce, `false` otherwise. This is done using the `isValidNonce` function defined [below](#transaction-verification).
 
+### Genesis Block Processing
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the bytes included in the block assets for the Auth module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisAuthStoreSchema` schema, given below.
```
IkerAlus, line 176: As it was pointed out in the [LIP 47 review](https://github.com/LiskHQ/lips/pull/119#discussion_r744547641), this is not a strictly correct formulation, since `genesisBlockAssetBytes` should refer to the `data` property of a deserialized entry of the `assets` array.

IkerAlus (changes requested):
One small formulation improvement

proposals/lip-0041.md
MaximeGagnebin: saw that comment... in the Token module, I wrote (probably anything along those lines should work, but the PR is yet to be reviewed):

Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.

proposals/lip-0041.md
ricott1: fixed

proposals/lip-0041.md
ricott1: edited following Maxime's suggy

IkerAlus (approved):
LGTM

MaximeGagnebin (approved):

proposals/lip-0041.md
```diff
@@ -170,6 +167,85 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
...
+genesisAuthStoreSchema = {
+  "type": "object",
+  "required": ["authDataSubstore"],
+  "properties": {
+    "authDataSubstore": {
+      "type": "array",
+      "fieldNumber": 1,
+      "items": {
+        "type": "object",
+        "required": ["storeKey", "storeValue"],
```
shuse2, line 198: Other LIPs are not using the term `storeKey` and `storeValue`. Do we want to make it consistent?

ricott1: you mean that they use the name of the property instead? so that here `storeKey` --> `address`, `storeValue` --> `authAccount`?

IkerAlus (approved):
lgtm

proposals/lip-0041.md
ricott1: updated

shuse2 (approved):

karmacoma (approved):

PR #119, opened 2021-11-04
Update LIP 0047 with genesis block processing

IkerAlus:


proposals/lip-0047.md
```diff
@@ -511,21 +647,29 @@ chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresse
 BFTThreshold = snapshotStore(0).threshold
 setBFTParameters(BFTThreshold, BFTThreshold, snapshotStore(0).validators)
 
+# Pass the BLS keys and generator keys to the Validators module
+for every validator element in genesisBlockAssetObject.validatorSubstore.validators:
+    registerValidatorKeys(validator.address, validator.proofOfPossession, validator.generatorKey, validator.blsKey)
```
AndreasKendziorra, line 652: The return value should be checked, and block execution cancelled if the the return value is not `True`.

proposals/lip-0047.md
```diff
@@ -511,21 +647,29 @@ chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresse
 BFTThreshold = snapshotStore(0).threshold
 setBFTParameters(BFTThreshold, BFTThreshold, snapshotStore(0).validators)
```
AndreasKendziorra, line 648: should't we abort the block execution if this function fails?

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
+```
+
+#### Genesis State Finalization
+
+To finalize the state of a genesis block `g`, the following logic is executed:
+
+Let `genesisBlockAssetBytes` be the bytes included in the block assets for the PoA module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, given below, then:
 
 ```python
 chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresses)
```
AndreasKendziorra, line 644: `snapshotStore(0).validators` instead of `snapshotStore(0).addresses`?

proposals/lip-0047.md
```diff
@@ -511,21 +647,29 @@ chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresse
 BFTThreshold = snapshotStore(0).threshold
 setBFTParameters(BFTThreshold, BFTThreshold, snapshotStore(0).validators)
 
+# Pass the BLS keys and generator keys to the Validators module
+for every validator element in genesisBlockAssetObject.validatorSubstore.validators:
+    registerValidatorKeys(validator.address, validator.proofOfPossession, validator.generatorKey, validator.blsKey)
+
 # Pass the list of validators to the Validators module
 for i in range(len(snapshotStore(0).validators)):
     addresses = snapshotStore(0).address[i]
```
AndreasKendziorra, line 656:
```suggestion
    addresses = snapshotStore(0).validators[i].address
```

AndreasKendziorra: the same for line 691

AndreasKendziorra (changes requested):
I see that some links are still pointing to the research forum. Maybe this could be fixed within this PR as well.

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
+
+```java
+genesisPoAStoreSchema = {
+   "type":"object",
+   "required":[
+      "validatorSubstore",
+      "snapshotSubstore"
+   ],
+   "properties":{
+      "validatorSubstore":{
```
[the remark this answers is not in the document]
ishantiw: `validators` could be part of `genesisPoAStoreSchema` and we can remove `validatorsSubstore` as mentioned by @AndreasKendziorra

ishantiw (approved):

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
 
-#### After Genesis Block Execution
+The following two steps are executed as part of the genesis block processing, see the “Update genesis block schema and processing” LIP for details. 
 
-After the genesis block `g` is executed, the following logic is executed:
+#### Genesis State Initialization
+
+To initialize the state of a genesis block `g`, the following logic is executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the bytes included in the block assets for the PoA module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, given below, then:
```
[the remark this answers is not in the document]
janhack: Consider using a short variable name such as `genesisAsset` or `assetPoA` instead of `genesisBlockAssetObject` for better readability and compactness below.

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
+The following two steps are executed as part of the genesis block processing, see the “Update genesis block schema and processing” LIP for details. 
 
-After the genesis block `g` is executed, the following logic is executed:
+#### Genesis State Initialization
+
+To initialize the state of a genesis block `g`, the following logic is executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the bytes included in the block assets for the PoA module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, given below, then: 
+
+* Checks for the `genesisBlockAssetObject.validatorSubstore`:
```
janhack, line 511: Wouldn't it be easier here and also better encourage code re-use to write:
For each entry in `genesisBlockAssetObject.validatorSubstore` perform the verification and execution logic of the authority registration command with the same parameters as input and `address` being the address of the sender account. Only the last step in the execution, callling `registerValidatorKeys`, is not done here as API calls should only be done in the genesis state finalization.

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
+
+* Checks for the `genesisBlockAssetObject.validatorSubstore`: 
+  * Check that the `address` property of every entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array is distinct pairwise.
+  * Check that the `name` property of every entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array contains only characters from the set `[a-z0-9!@$&_.]`, is not empty and is at most `MAX_LENGTH_NAME` characters long.
+  * Check that the `generatorKey` property of every entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array has length 32.
+  * Check that the `blsKey` property of every entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array has length 48.
+  * Check that the `proofOfPossession` property of every entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array has length 96.
+
+* Checks for the `genesisBlockAssetObject.snapshotSubstore`:
```
janhack, line 518: Wouldn't it be easier here and also better encourage code re-use to reference the same validation logic already described above:
- For `genesisBlockAssetObject.snapshotSubstore.activeValidators` apply the same verifications as for `trs.params.newValidators` in the update authority command verification.
- For `genesisBlockAssetObject.snapshotSubstore.threshold` apply the same verifications as for `trs.params.threshold` in the update authority command verification.

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
...
+* Create an entry in the name substore for each entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array as:
+  * `storeKey`: `validator.name` serialized as a utf-8 encoded string.
+  * The `storeValue` is the serialization of the object `validatorAddress` following `validatorAddressSchema` with `validatorAddress.address = validator.address`.
+
+* Create three entries in the snapshot substore as:
+  * The `storeKey` of each of the entries are `uint32be(0)`, `uint32be(1)`, and `uint32be(2)` respectively.
+  * The `storeValue` is the same for the three entries and equal to the serialization of `genesisBlockAssetObject.snapshotSubstore`.
+
+* Create an entry in the chain properties substore as specified above.
+  * The `storeValue` is the serialization of an object following `chainPropSchema` where all the properties are initialized to their default values.
```
janhack, line 542: Above you sate about the `roundEndHeight`:
> Its initial value is set after the execution of the genesis block.

I think this should still be true and we should not assume that the genesis height is `0`. 

I would also remove the comment about the initialization that you have in the section above as it is redundant there.

proposals/lip-0047.md
janhack: Note that we also always overwrite the value `addresses`, but we actually want to append it to a list, right?

proposals/lip-0047.md
janhack: Note that in this LIP we sometimes use `BFTThreshold` and sometimes `BFT_THRESHOLD`. As this is a variable and not a constant, I would suggest to switch to camelCase, e.g., `bftThreshold`.

proposals/lip-0047.md
janhack: Ah, now I see that this value is overwritten later. It is normally not good practice to initialize something in between with an incorrect value, even if it is not used. 

Maybe consider setting it directly to the final value or first to the genesis height and then doing
```
    chainProperties.roundEndHeight = chainProperties.roundEndHeight + len(snapshotStore(0).validators)
```
below.

proposals/lip-0047.md
```diff
@@ -511,21 +647,29 @@ chainProperties.roundEndHeight = g.header.height + len(snapshotStore(0).addresse
...
+### Block Processing
+
+The following steps are executed as part of the block processing, see the [LIP 0055][lip-0055] for details.
+
+#### After Transactions Execution
 
-After a block `b` is executed, the following logic is executed:
+After the transactions in a block `b` are executed, the following logic is executed:
 
 ```python
```
janhack, line 674: I think also after the pseudocode below we should list all functions that are used from other modules there. Currently, `setGeneratorList` does not appear in the list. In general, I would find it easier to once define the notation `moduleName.fct` and mention that in this LIP you call functions from the BFT, Random and Validator module. Then it the notation like `validatorsModule.setGeneratorList` or `Validators.setGeneratorList` makes it directly clear that this is not a local function defined in this LIP but an API.

janhack (changes requested):
The general approach towards the genesis logic looks good. My suggestion would be to rather reference existing logic from the LIP instead of copy pasting it there.

proposals/lip-0047.md
```diff
@@ -498,11 +498,147 @@ This module does not define any specific logic for other modules.
 
 This module does not define any endpoint for off-chain services.
 
-### Protocol Logic During Block Lifecycle
+### Genesis Block Processing
 
-#### After Genesis Block Execution
+The following two steps are executed as part of the genesis block processing, see the “Update genesis block schema and processing” LIP for details.
```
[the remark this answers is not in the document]
IkerAlus: yes, hopefully, it will be merged in the next days.

proposals/lip-0047.md
IkerAlus: changed to the second suggestion

proposals/lip-0047.md
IkerAlus: fixed variable names as jan suggested.
Regarding Andreas question, I think that is already addressed in line 644

proposals/lip-0047.md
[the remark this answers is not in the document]
IkerAlus: `validatorsSubstore` removed as suggested

proposals/lip-0047.md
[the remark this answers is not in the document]
IkerAlus: Added a correct formulation for the deserialization of the assets as Andreas suggested
Also, shortened `genesisBlockAssetObject` to `assetPoA` as Jan suggested

proposals/lip-0047.md
IkerAlus: I added a clarification sentence in this direction. This way we have the point of re-use, the strictly correct specs for the genesis and I don't need to revert changes  that are actually correct.

proposals/lip-0047.md
IkerAlus: I think in this case it is a bit dangerous to refer to that logic since for the update authority command verification the entries in the validator substore already exist and we do not need to check for uniqueness etc.

proposals/lip-0047.md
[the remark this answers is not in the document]
AndreasKendziorra: ok. I'll suggest to keep this comment open then and wait until the other PR is merged. Do you agree?

proposals/lip-0047.md
```diff
@@ -506,92 +507,87 @@ The following two steps are executed as part of the genesis block processing, se
...
-
-* Checks for the `genesisBlockAssetObject.snapshotSubstore`: 
-  * The entries in the `genesisBlockAssetObject.snapshotSubstore.activeValidators` array are ordered lexicographically according to `address`.
-  * For every element `activeValidator` in the `genesisBlockAssetObject.snapshotSubstore.activeValidators` array, there is an entry with `address == activeValidator.address` in the `genesisBlockAssetObject.validatorSubstore.validators` array.
-  * The `weight` property of every element `activeValidator` in the `genesisBlockAssetObject.snapshotSubstore.activeValidators` array is a positive integer.
-  * Let `totalWeight` be the sum of the `weight` property of every element `activeValidator`. Then `totalWeight` has to be less than or equal to `MAX_UINT64`.
-  * The value of `genesisBlockAssetObject.snapshotSubstore.threshold` is within the following range:
+Let `genesisBlockAssetBytes` be the bytes included in the `data` property of a deserialized entry in the block assets for the PoA module and let `assetPoA` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, then:
```
AndreasKendziorra, line 510:
```suggestion
Let `genesisBlockAssetBytes` be the bytes included in the `data` property of a deserialized entry in the block assets for the PoA module and let `assetPoA` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema. Then: 
```

AndreasKendziorra: Just a suggestion as the sentence is very long.

proposals/lip-0047.md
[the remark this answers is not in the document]
AndreasKendziorra: can be resolved

proposals/lip-0047.md
AndreasKendziorra: I think like this it is still not very clear as line 648 is just a comparison statement. But it's not clear that something fails here if `registerValidatorKeys` returns `False`. I would propose something like:

``` 
for every validator element in genesisBlockAssetObject.validatorSubstore.validators:
    if registerValidatorKeys(validator.address, validator.proofOfPossession, validator.generatorKey, validator.blsKey) != True:
        discard the block 
```

Maybe this functions becomes easier to understand if the logic is put into a function that returns True or False, and in the introduction you saysthat the block is discarded if the function returns `False`.

proposals/lip-0047.md
AndreasKendziorra: hm. I see 4 different cases where `[setBFTParameters](https://github.com/LiskHQ/lips/blob/c308e37920b31104da3f4ebd3eaebfdd8590a12f/proposals/lip-introduce_bft_module.md#execution-12)`, and already for the first I don't see clearly this will never happen. In particular, I do not see how line 644 prevents this.

proposals/lip-0047.md
AndreasKendziorra: can be resolved

proposals/lip-0047.md
```diff
@@ -732,3 +732,4 @@ TBA
 [research:validators-module]: https://research.lisk.com/t/introduce-validators-module/317
```
AndreasKendziorra, line 732: again: could we update the links from the research forum to the LIPs repo for all published LIPs? :)

AndreasKendziorra (changes requested):

proposals/lip-0047.md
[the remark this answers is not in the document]
janhack: Can be resolved.

proposals/lip-0047.md
janhack: From my side the comment can be resolved.

proposals/lip-0047.md
janhack: can be resolved

proposals/lip-0047.md
janhack: Can be resolved.

proposals/lip-0047.md
```diff
@@ -276,6 +276,7 @@ registrationTransactionParamsSchema = {
 Let `trs` be a transaction with module ID `MODULE_ID_POA` and command ID `COMMAND_ID_REGISTRATION_AUTHORITY` to be verified. The list of verification conditions for `trs.params` is as follows:
 
 * The `trs.params.name` property has to contain only characters from the set `[a-z0-9!@$&_.]`, must not be empty and has to be at most `MAX_LENGTH_NAME` characters long.
+* The `trs.params.name` property has to be unique with respect to the set of already registered authority names in the blockchain state, such that the entry with store prefix equal to `STORE_PREFIX_NAME` and store key equal to `trs.params.name` (serialized as a utf-8 encoded string) does not exist in the store.
```
janhack, line 279:
```suggestion
* The `trs.params.name` property has to be unique with respect to the set of already registered authority names in the blockchain state, i.e. the entry with store prefix equal to `STORE_PREFIX_NAME` and store key equal to `trs.params.name` (serialized as a utf-8 encoded string) does not exist in the store.
```

proposals/lip-0047.md
```diff
@@ -498,53 +499,194 @@ This module does not define any specific logic for other modules.
...
+
+* Create an entry in the name substore for each entry `validator` in the `assetPoA.validators` array as:
+  * `storeKey`: `validator.name` serialized as a utf-8 encoded string.
+  * The `storeValue` is the serialization of the object `validatorAddress` following `validatorAddressSchema` with `validatorAddress.address = validator.address`.
+
+* Create three entries in the snapshot substore as:
+  * The `storeKey` of each of the entries are `uint32be(0)`, `uint32be(1)`, and `uint32be(2)` respectively.
+  * The `storeValue` is the same for the three entries and equal to the serialization of `assetPoA.snapshotSubstore`.
+
+* Create an entry in the chain properties substore as specified above.
```
janhack, line 544:
```suggestion
* Create an entry in the chain properties substore as specified beow.
```

proposals/lip-0047.md
```diff
@@ -498,53 +499,194 @@ This module does not define any specific logic for other modules.
...
+
+* Checks for the `assetPoA.validators` (note that these checks are equivalent to those for the verification of the Authority Registration command): 
+  * Check that the `address` properties of all entries in the `assetPoA.validators` array are pairwise distinct.
+  * Check that the `name` property of every entry in the `assetPoA.validators` array contains only characters from the set `[a-z0-9!@$&_.]`, is not empty and is at most `MAX_LENGTH_NAME` characters long.
+  * Check that the `name` property of all entries in the `assetPoA.validators` array are pairwise distinct. 
+  * Check that the `generatorKey` property of every entry in the `assetPoA.validators` array has length 32.
+  * Check that the `blsKey` property of every entry in the `assetPoA.validators` array has length 48.
+  * Check that the `proofOfPossession` property of every entry in the `assetPoA.validators` array has length 96.
+
+* Checks for the `assetPoA.snapshotSubstore`:
```
janhack, line 520: We do not check the number of entries of the array as for the authority update command.

janhack: Regarding the comment above.

>  I think in this case it is a bit dangerous to refer to that logic since for the update authority command verification the entries in the validator substore already exist and we do not need to check for uniqueness etc.

Also in the authority update command we check address uniqueness. I think all checks are the same or which one specifically is different?

janhack (changes requested):
There are a few minor open points left.

proposals/lip-0047.md
IkerAlus: updated

proposals/lip-0047.md
IkerAlus: updated

proposals/lip-0047.md
IkerAlus: Yes, the checks are equivalent but the command one is against the existing state store and here it is against the content of another property in the assets data. That is why I would prefer not to just point out to those checks. I have added a similar explanatory sentence as I did for checks for the `assetPoA.validators` though.

Good point regarding the number of entries, added.

proposals/lip-0047.md
```diff
@@ -506,92 +507,87 @@ The following two steps are executed as part of the genesis block processing, se
...
+* Checks for the `assetPoA.validators` (note that these checks are equivalent to those for the verification of the Authority Registration command): 
+  * Check that the `address` properties of all entries in the `assetPoA.validators` array are pairwise distinct.
+  * Check that the `name` property of every entry in the `assetPoA.validators` array contains only characters from the set `[a-z0-9!@$&_.]`, is not empty and is at most `MAX_LENGTH_NAME` characters long.
+  * Check that the `name` property of all entries in the `assetPoA.validators` array are pairwise distinct. 
+  * Check that the `generatorKey` property of every entry in the `assetPoA.validators` array has length 32.
+  * Check that the `blsKey` property of every entry in the `assetPoA.validators` array has length 48.
+  * Check that the `proofOfPossession` property of every entry in the `assetPoA.validators` array has length 96.
+
+* Checks for the `assetPoA.snapshotSubstore`: 
+  * Check that the `address` properties of entries in the `assetPoA.snapshotSubstore.activeValidators` are pairwise distinct.
```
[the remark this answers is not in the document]
IkerAlus: I fixed it by adding `where ⌊⋅⌋ is the floor function.` in the same line it is mentioned

proposals/lip-0047.md
```diff
@@ -506,92 +507,87 @@ The following two steps are executed as part of the genesis block processing, se
...
+  * Check that entries in the `assetPoA.snapshotSubstore.activeValidators` array are ordered lexicographically according to `address`.
+  * Check that for every element `activeValidator` in the `assetPoA.snapshotSubstore.activeValidators` array, there is an entry with `address == activeValidator.address` in the `assetPoA.validators` array.
```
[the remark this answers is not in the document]
IkerAlus: maybe I am missing something but I added the pairwise distinct condition above in this set of checks.

proposals/lip-0047.md
IkerAlus: removed any location detail

proposals/lip-0047.md
```diff
@@ -636,24 +632,25 @@ genesisPoAStoreSchema = {
 
 #### Genesis State Finalization
 
-To finalize the state of a genesis block `g`, the following logic is executed:
+To finalize the state of a genesis block `g`, the following logic is executed.  If any step fails, the block is discarded and has no further effect.
 
-Let `genesisBlockAssetBytes` be the bytes included in the block assets for the PoA module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, given below, then:
+Let `genesisBlockAssetBytes` be the bytes included in the `data` property of a deserialized entry in the block assets for the PoA module and let `assetPoA` be the deserialization of `genesisBlockAssetBytes` according to the `genesisPoAStoreSchema` schema, then:
```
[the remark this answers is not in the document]
IkerAlus: updated here and above

proposals/lip-0047.md
IkerAlus: true, I thought I did it already

proposals/lip-0047.md
IkerAlus: I think it is already clear that if this function does not return true, the processing must stop. 

Anyway, I went for your first suggestion here

proposals/lip-0047.md
IkerAlus: It does not prevent that if fails, what it says is that if `setBFTParameters` (for whatever the reason) fails, then the processing of this block halts and it is discarded.

It is up to the sidechain dev to set the right values for the BFT parameters so that this function does not fail.

AndreasKendziorra: ok, I see that it is logically correct. But I think it's hard to get the logic from reading it. One reason is that it's not obvious at all that a setter function can fail. The second reason is now the inconsistency between calling `setBFTParameters` and calling `registerValidatorKeys`. For the latter one, the outcome of the call is explicitly used in the pseudo code. For the former one not. And it had to be done for `registerValidatorKeys` because "failing" is not well defined for this function.

proposals/lip-0047.md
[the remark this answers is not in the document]
AndreasKendziorra: I meant the suggestion created by the "suggestion" feature :) That means, rewriting this line to:

...there is an entry `validator` in the `genesisBlockAssetObject.validatorSubstore.validators` array with `validator.address == activeValidator.address`.

proposals/lip-0047.md
```diff
@@ -517,17 +515,16 @@ Let `genesisBlockAssetBytes` be the bytes included in the `data` property of a d
   * Check that the `blsKey` property of every entry in the `assetPoA.validators` array has length 48.
   * Check that the `proofOfPossession` property of every entry in the `assetPoA.validators` array has length 96.
 
-* Checks for the `assetPoA.snapshotSubstore`: 
+* Checks for the `assetPoA.snapshotSubstore` (note that these checks are equivalent to those for the verification of the Authority Update command):
```
AndreasKendziorra, line 518: I know that you added the statement about the equivalence based on a comment from Jan. But I think it is misleading as it is not to 100% equivalent as you pointed out. Maybe "similar", "almost identical" or "almost equivalent". Or you could just drop it.

AndreasKendziorra (changes requested):
Just one minor detail left.

proposals/lip-0047.md
```diff
@@ -498,53 +498,194 @@ This module does not define any specific logic for other modules.
...
+  * The `storeValue` is the same for the three entries and equal to the serialization of `assetPoA.snapshotSubstore`.
+
+* Create an entry in the chain properties substore as:
+  * The `storeValue` is the serialization of a `chainProperties` object following `chainPropSchema` where:
+    * `chainProperties.roundEndHeight` = `g.header.height`.
+    * `chainProperties.validatorsUpdateNonce` = 0.
+
+##### Genesis PoA Store Schema
+
+```java
```
janhack, line 549: I would suggest to run `prettier` on the schema to improve the spacing.

proposals/lip-0047.md
```diff
@@ -498,53 +498,194 @@ This module does not define any specific logic for other modules.
...
+               "blsKey",
+               "proofOfPossession",
+               "generatorKey"
+            ],
+            "properties":{
+               "name":{
+                  "dataType":"string",
+                  "fieldNumber":1
+               },
+               "address":{
```
janhack, line 574: For me the `address` would be the primary identifier and I would therefore put that one first. Additionally, I would consider enforcing lexicographic order by address. This is not really required for the protocol, but I think it makes manual verification and also the programmatic verification (distinct addresses) easier.

janhack (changes requested):
Looks good. I just have two minor comments.

proposals/lip-0047.md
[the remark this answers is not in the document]
IkerAlus: link to LIP 0060 has been added

proposals/lip-0047.md
```diff
@@ -584,6 +726,8 @@ TBA
 [lip-0038]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0038.md
 [lip-0038#public-key-registration]: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0038.md#public-key-registration-and-proof-of-possession
 [research:bft_module]: https://research.lisk.com/t/introduce-bft-module/321
-[research:random-module]: https://research.lisk.com/t/define-state-and-state-transitions-of-the-random-module/311
-[research:validators-module]: https://research.lisk.com/t/introduce-validators-module/317
-[research:validators-module#register-validator-keys]: https://research.lisk.com/t/introduce-validators-module/317#registervalidatorkeys-27
+[research:random-module]: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0046.md
```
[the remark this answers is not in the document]
IkerAlus: done

proposals/lip-0047.md
IkerAlus: good point, done

proposals/lip-0047.md
IkerAlus: changed to "similar"

proposals/lip-0047.md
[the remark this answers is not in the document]
IkerAlus: Now it is all clear, changed as suggested, thanks.

P.D: linking old suggestions doesn't seem to work

proposals/lip-0047.md
IkerAlus: good point, I fixed the schema and added the extra condition.

proposals/lip-0047.md
janhack: Okay, can be resolved.

janhack (approved):
Looks all good!

AndreasKendziorra (approved):

PR #121, opened 2021-11-08
Update LIP 0051 with genesis block processing

MaximeGagnebin:


proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+                    }
+                }
+            }
+        },
+        "escrowSubstore": {
+            "type": "array",
+            "fieldNumber": 1,
+            "items": {
+                "type": "object",
+                "required": ["escrowChainID", "tokenLocalID", "amount"],
```
IkerAlus, line 869: This should be `localID` to be consistent with definition below.

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+  ```
+
+* Create an entry in the available local ID substore with
+  ```python
+  storeKey = EMPTY_BYTES
+  storeValue = {"nextAvailableLocalID": genesisBlockAssetObject.availableLocalIDSubstore} 
+               serialized using availableLocalIDStoreSchema.
+  ```
+
+* For each entry `terminatedChainID` in `genesisBlockAssetObject.terminatedEscrowSubstore`, create an entry in the terminated escrow substore with
```
IkerAlus, line 943: What purpose would have to start a sidechain with terminated escrows of other chains? 

Also, if there is no corresponding entry in `escrowSubstore` for the terminated chain, would this substore be usable in any way?

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+genesisTokenStoreSchema = {
+    "type": "object",
+    "required": [
+        "userSubstore",
+        "supplySubstore",
+        "escrowSubstore",
+        "availableLocalIDSubstore",
+        "terminatedEscrowSubstore"
+    ],
+    "properties": {
```
IkerAlus, line 789: the field numbers of these properties have to be fixed

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+```
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks:
```
IkerAlus, line 908: It is self-explanatory to me but to be consistent with following bullet points, we should add `genesisBlockAssetObject.` to every property that is referenced in these initial checks.

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+   * Accross all elements of the `userSubstore` array, the pairs `address, tokenID` must be unique (each pair appears at most once, though a given address can appear multiple times with different token IDs). 
+   * For each such `address, tokenID` pair, the values used as `moduleID` in the `lockedBalances` array must be unique.
+   * Accross all element of the `supplySubstore` array, all values given for `localID` must be unique.
+   * Accross all elements of the `escrowSubstore` array, the pairs `escrowChainID, localID` must be unique (each pair appears at most once).
```
IkerAlus, line 912: I assume that similar check applies for all elements in `terminatedEscrowSubstore`, right?

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+  totalSupplyCorrect():
+      ownedSupply = {}
+      for (storeKey, storeValue) in user substore:
+          chainID = storeKey[20:24] # part of the key corresponding to the chain ID of the token
+          localID = storeKey[20:28] # part of the key corresponding to the local ID of the token
+          if chainID == CHAIN_ID_ALIAS_NATIVE:
+              ownedSupply[localID] += sum of availableBalance and all locked amounts of storeValue
+              
+      for (storeKey, storeValue) in escrow substore:
+          localID = storeKey[4:8] # part of the key corresponding to the local ID of the token
```
IkerAlus, line 962: is this `localID` going to be the same as in the previous for loop? in other words, are we sure that all the substore entries for every store are created ordered respect to `localID` in the store key? 

same consideration for the supply substore below.

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks:
+   * Accross all elements of the `userSubstore` array, the pairs `address, tokenID` must be unique (each pair appears at most once, though a given address can appear multiple times with different token IDs). 
+   * For each such `address, tokenID` pair, the values used as `moduleID` in the `lockedBalances` array must be unique.
```
IkerAlus, line 910: should we add a check that the module ID can't be the one assigned to the token module?

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks:
+   * Accross all elements of the `userSubstore` array, the pairs `address, tokenID` must be unique (each pair appears at most once, though a given address can appear multiple times with different token IDs).
```
IkerAlus, line 909: we are missing the check the length of `address`

proposals/lip-0051.md
```diff
@@ -769,14 +770,222 @@ if the above send function does not fail:
...
+  storeKey = uint32be(terminatedChainID)
+  storeValue = {"escrowTerminated": True} serialized using terminatedEscrowSchema.
+  ```
+  
+Once the module store is initialized, its validity is attested via the two checks below.
+* Check that for each native token the total supply is correct. 
+  This can be done by checking that the function below returns `True`.
+  ```python
+  totalSupplyCorrect():
+      ownedSupply = {}
```
IkerAlus, line 954: what would happen if that at the end of the cumulative sum an element of this array ends up being bigger than uint64? are we always safe that the function would fail in that case? probably it is better to add a check in that regard so we do not have corner cases with integer overflows.

IkerAlus (changes requested):
It is in a good state, just a few improvements and considerations

proposals/lip-0051.md
MaximeGagnebin: I would say `no`. If you do that, then you can never unlock those tokens, but that does not mean it should be forbidden.

proposals/lip-0051.md
MaximeGagnebin: The purpose is mainly: testing and regenesis. 
As this state can exist, I allowed it to be created at genesis. 
Funnily enough, there is almost no link between chains in the `terminated escrow substore` and tokens in the `escrowSubstore`. The terminated escrow substore only serves as a reminder of chains that have broken the token module rules, regardless of the existence of tokens actually escrowed to that chain.

proposals/lip-0051.md
MaximeGagnebin: indeed, fixed

proposals/lip-0051.md
MaximeGagnebin: good catch, fixed (hope I did not miss any 😅 )

proposals/lip-0051.md
MaximeGagnebin: reworded (tried to avoid the repetition though)

proposals/lip-0051.md
MaximeGagnebin: Yes, added.

proposals/lip-0051.md
MaximeGagnebin: So we don't really know in which order the `localID` will appear. But it does not really matter. Once we have the local ID, we increment the corresponding entry in the `ownedSupply` dictionary.

proposals/lip-0051.md
MaximeGagnebin: yes, added.

proposals/lip-0051.md
MaximeGagnebin: Ok, added the check

proposals/lip-0051.md
IkerAlus: ah true, `ownedSupply` is a dictionary and not just an array.  This can be resolved

IkerAlus (approved):
great work

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks on the properties of `genesisBlockAssetObject`:
+   * Accross all elements of the `userSubstore` array, all `address` values must have length `ADDRESS_LENGTH`. 
+   * Accross all elements of the `userSubstore` array, the pairs `address, tokenID` must be unique (each pair appears at most once, though a given address can appear multiple times with different token IDs). 
+   * For each such `address, tokenID` pair, the values used as `moduleID` in the `lockedBalances` array must be unique.
```
janhack, line 911: Note that in the state store we have an additional requirement:
> In the above object, lockedBalances is always kept ordered by ascending order of moduleID.

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+                        "fieldNumber": 2
+                    },
+                    "amount": {
+                        "dataType": "uint64",
+                        "fieldNumber": 3
+                    }
+                }
+            }
+        },
+        "availableLocalIDSubstore": {
```
janhack, line 886: Personally, I find it odd that `availableLocalIDSubstore` is actually corresponding to the one element in that substore. Maybe with additional nesting this could be made more clear, but this should be the decision by the developers (@mitsuaki-u ).

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+* For each entry `escrowEntry` in `genesisBlockAssetObject.escrowSubstore`, create an entry in the escrow substore  with
+  ```python
+  storeKey = uint32be(escrowEntry.escrowChainID) || uint32be(escrowEntry.localID)
+  storeValue = {"amount": escrowEntry.amount} serialized using escrowStoreSchema.
+  ```
+
+* Create an entry in the available local ID substore with
+  ```python
+  storeKey = EMPTY_BYTES
+  storeValue = {"nextAvailableLocalID": genesisBlockAssetObject.availableLocalIDSubstore}
```
janhack, line 941: Not really for this line, but should we check that `availableLocalIDSubstore` is larger than any local ID used?

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+  storeKey = uint32be(terminatedChainID)
+  storeValue = {"escrowTerminated": True} serialized using terminatedEscrowSchema.
+  ```
+  
+Once the module store is initialized, its validity is attested via the two checks below.
+* Check that for each native token the total supply is correct. 
+  This can be done by checking that the function below returns `True`.
+  ```python
+  totalSupplyCorrect():
+      ownedSupply = {}
```
janhack, line 956: Naming suggestion: I would maybe say `computedSupplies` or something similar, as it is not just one supply and ownership does not necessarily make sense for escrow or addresses without owner (e.g., used by other modules like the DEX to lock tokens).

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+      for (storeKey, storeValue) in escrow substore:
+          localID = storeKey[4:8] # part of the key corresponding to the local ID of the token
+          ownedSupply[localID] += storeValue.amount
+          
+      if ownedSupply[localID] >= 2^64 for any localID:
+          return False
+      
+      recordedSupply = {}
+      for (storeKey, storeValue) in supply substore:
+          localID = storeKey
```
janhack, line 972: A bit odd as you could have named the variable already differently above.

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+              ownedSupply[localID] += sum of availableBalance and all locked amounts of storeValue
+              
+      for (storeKey, storeValue) in escrow substore:
+          localID = storeKey[4:8] # part of the key corresponding to the local ID of the token
+          ownedSupply[localID] += storeValue.amount
+          
+      if ownedSupply[localID] >= 2^64 for any localID:
+          return False
+      
+      recordedSupply = {}
```
janhack, line 970: Cosmetic suggestion: I would find "storedSupplies" much more straightforward.

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+  ```python
+  storeKey = uint32be(terminatedChainID)
+  storeValue = {"escrowTerminated": True} serialized using terminatedEscrowSchema.
+  ```
+  
+Once the module store is initialized, its validity is attested via the two checks below.
+* Check that for each native token the total supply is correct. 
+  This can be done by checking that the function below returns `True`.
+  ```python
+  totalSupplyCorrect():
```
janhack, line 955: Maybe `areTotalSuppliesCorrect()` or `validateSupplyStoreEntries()`. The current name would be an odd phrase in English.

janhack (changes requested):
There are a few minor issues, otherwise the genesis logic looks good.

proposals/lip-0051.md
MaximeGagnebin: changed

proposals/lip-0051.md
MaximeGagnebin: this line is only here to keep the symmetry with the other loops and the choosing the dictionary key from the store key.

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+```
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks on the properties of `genesisBlockAssetObject`:
```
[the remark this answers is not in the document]
MaximeGagnebin: I added the requirement.

proposals/lip-0051.md
MaximeGagnebin: I added the requirement.

proposals/lip-0051.md
MaximeGagnebin: I added a nesting, let me know it it looks good to you @mitsuaki-u , we can always remove it.

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+      if ownedSupply[localID] >= 2^64 for any localID:
+          return False
+      
+      recordedSupply = {}
+      for (storeKey, storeValue) in supply substore:
+          localID = storeKey
+          recordedSupply[localID] = storeValue.amount
+      
+      # check if both dictionary are equal, same keys, same corresponding values
+      if recordedSupply == ownedSupply:
```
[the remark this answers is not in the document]
MaximeGagnebin: Indeed, updated the logic. Thanks.

proposals/lip-0051.md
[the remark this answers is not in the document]
janhack: This was changed for `userSubstore`, but I would similarly apply it for `supplySubstore`, `escrowSubstore` and `terminatedEscrowSubstore`.

proposals/lip-0051.md
```diff
@@ -769,14 +770,243 @@ if the above send function does not fail:
...
+          storedSupplies[localID] = storeValue.amount
+      
+      # check if both dictionary are coherent
+      for localID in computedSupplies:
+          if storedSupplies[localID] does not exist or computedSupplies[localID] != storedSupplies[localID]:
+              return False
+              
+      for localID in storedSupplies:
+          if (computedSupplies[localID] does not exist and storedSupplies[localID] != 0)
+              or (computedSupplies[localID] != storedSupplies[localID]):
```
janhack, line 992: I think the line above still should to be removed because the condition will be evaluated for
`computedSupplies[localID]` not existing and `storedSupplies[localID]==0`. Note that also it is redundant as the relevant check is already done in the for loop above.

proposals/lip-0051.md
janhack: Can be resolved from my side.

janhack (changes requested):
There are two minor open comments left from my side.

proposals/lip-0051.md
MaximeGagnebin: Indeed, removed.

proposals/lip-0051.md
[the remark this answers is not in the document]
MaximeGagnebin: Added.

janhack (approved):

proposals/lip-0051.md
```diff
@@ -769,14 +770,245 @@ if the above send function does not fail:
...
+  
+Once the module store is initialized, its validity is attested via the two checks below.
+* Check that for each native token the total supply is correct. 
+  This can be done by checking that the function below returns `True`.
+  ```python
+  validateSupplyStoreEntries():
+      computedSupplies = {}
+      for (storeKey, storeValue) in user substore:
+          chainID = storeKey[20:24] # part of the key corresponding to the chain ID of the token
+          localID = storeKey[20:28] # part of the key corresponding to the local ID of the token
```
mitsuaki-u, line 972: is the starting index of `20` here correct? it seems the same for `chainID` above

mitsuaki-u (changes requested):
Minor spelling issues

proposals/lip-0051.md
```diff
@@ -769,14 +770,227 @@ if the above send function does not fail:
...
+
+
+#### Genesis State Initialization
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the Token module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisTokenStoreSchema` schema, given above.
+
+* Initial checks on the properties of `genesisBlockAssetObject`:
+   * Accross all elements of the `userSubstore` array, all `address` values must have length `ADDRESS_LENGTH`.
```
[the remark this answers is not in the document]
MaximeGagnebin: Yes, fixed.

proposals/lip-0051.md
MaximeGagnebin: nice catch, should be 24. fixed

mitsuaki-u (approved):
LGTM 🚀

karmacoma (approved):

PR #131, opened 2022-01-14
Update LIPs 0043, 0045, 0049, 0051, 0052, 0053, 0054

ricott1:
### Token LIP

- :ok: beforeCCMSent and beforeCCMExecution must use the feeTokenID to handle the fee.

### LIP0045

- :ok: chain account:  add `messageFeeTokenID` 

### LIP0043

- :ok: include `messageFeeTokenID = LSK` to chain account creation (in both directions). You don't need to include it explicitly in the registration commands, but it should be set automatically.

### LIP0053

- :ok: If there is a direct channel, discard ccms with `status=recovered`. This change will not be included yet, but it will be part of the flat ecosystem LIP, updating the current protocol for message processing. (added this check to CCM LIP)
- :ok: Drop requirements: 
  - Validate that the first CCM in inboxUpdate.crossChainMessages has CCM.nonce == partnerChain.inbox.size.
  - Validate that all CCMs in inboxUpdate.crossChainMessages have increasing and sequential nonce property.

### LIP0049

- :ok: add `messageFeeTokenID` to registration CCM. It should then hold automatically  that it equals `LSK` if channel is with mainchain. It is still an open question how to process this exactly. (I just check against the value stored in the channel account)
- :ok: general refactoring of CCM processing

### LIP54

- :ok: add `initializeMessageRecovery` to sidechain. Modify `mainchainChainAccount` to smtg more generic (done in the new LIP, in the standard LIP 0054 there is no mention of direct recovery)
- :ok: messages from the mainchain with `status = recovered` should be rejected if there is an open channel. How do we do this? (in process function in CCM LIP, done)
- :ok: message recovery command unescrows fee only if messageFeeToken is local, passing localID of token.


proposals/lip-0049.md
```diff
@@ -38,7 +38,7 @@ In the following, we list all the properties of a cross-chain message.
 
 #### Sending Chain ID and Receiving Chain ID
 
-Used to identify the chains exchanging the cross-chain message. On the mainchain, the receiving chain ID is read to route the message to the corresponding chain outbox, as specified in [LIP "Introduce cross-chain update mechanism"][research:ccu]. The sending chain ID is used, for example, if the message triggers an error and has to be sent back.
+Used to identify the chains exchanging the cross-chain message. On the mainchain, the receiving chain ID is read to route the message to the corresponding chain outbox, as specified in [LIP 0053][lip-0053]. The sending chain ID is used, for example, if the message triggers an error and has to be sent back.
```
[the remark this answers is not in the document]
ricott1: I meant it is read.

proposals/lip-0049.md
```diff
@@ -148,8 +142,6 @@ The function `uint32be(x)` returns the big endian uint32 serialization of an int
 
 #### Chain Account
 
-Let `account(chainID)` be the entry in the interoperability store with store prefix `STORE_PREFIX_CHAIN_DATA` and store key `uint32be(chainID)`.
-
```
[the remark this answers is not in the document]
ricott1: done

proposals/lip-0045.md
```diff
@@ -74,27 +79,27 @@ Active sidechains are required to prove their liveness to the mainchain at least
 
 The life cycle of a sidechain can be split into 3 parts, corresponding to the 3 values of the account status property: ''registered'', ''active'', and ''terminated''.
 
-A sidechain registers on the mainchain with a [sidechain registration command][research:chain-registration]. This command creates the sidechain account on the mainchain, with initial status set to ''registered''. Thereafter, the mainchain account is created on a sidechain with a [mainchain registration command][research:chain-registration], with initial status set to ''registered''.
+A sidechain registers on the mainchain with a [sidechain registration command][lip-0043]. This command creates the sidechain account on the mainchain. In particular, it creates an entry in the chain data substore, with initial status set to ''registered''; an entry in the channel data substore, with empty inbox and outbox; and an entry in the validators data substore, holding the initial sidechain validators set. Thereafter, the mainchain account is similarly created on a sidechain with a [mainchain registration command][lip-0043].
```
mitsuaki-u, line 82:
```suggestion
A sidechain registers on the mainchain with a [sidechain registration command][lip-0043]. This command creates the sidechain account on the mainchain. In particular, it creates an entry in the chain data substore, with initial status set to 'registered'; an entry in the channel data substore, with empty inbox and outbox; and an entry in the validators data substore, holding the initial sidechain validators set. Thereafter, the mainchain account is similarly created on a sidechain with a [mainchain registration command][lip-0043].
```

proposals/lip-0045.md
```diff
@@ -74,27 +79,27 @@ Active sidechains are required to prove their liveness to the mainchain at least
...
 
-When a chain is terminated, a ''terminated chain'' account is created, storing the last certified state root of the terminated chain. Then, users can recover messages and tokens from the terminated sidechain with a recovery command.
+When a chain is terminated, a ''terminated state'' account is created, storing the last certified state root of the terminated chain. Then, users can recover tokens from the terminated sidechain with a state recovery command or proceed with the message recovery initialization command to kick-off the recovery of pending cross-chain messages.
```
mitsuaki-u, line 88:
```suggestion
When a chain is terminated, a 'terminated state' account is created, storing the last certified state root of the terminated chain. Then, users can recover tokens from the terminated sidechain with a state recovery command or proceed with the message recovery initialization command to kick-off the recovery of pending cross-chain messages.
```

mitsuaki-u (commented):
LGTM, some minor comments

ishantiw (approved):

proposals/lip-0051.md
```diff
@@ -1008,10 +963,10 @@ Once the module store is initialized, its validity is attested via the two check
 
 ### Mainchain Minimum Balance Specifications
 
-As specified in [LIP 0025](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0025.md), mainchain user substore entries cannot hold less than `MIN_BALANCE` of LSK token. To follow this rule:
+As specified in [LIP 0025](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0025.md), mainchain user substore entries cannot hold less than `MIN_BALANCE` of LSK token. To follow this rule:
```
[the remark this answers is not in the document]
ricott1: thanks, done

proposals/lip-0051.md
```diff
@@ -1267,46 +1175,94 @@ unlock(address, moduleID, tokenID, amount):
...
+```
+
+#### forwardMessageFee
+
+This function is only called during the execution of a cross-chain update on the mainchain.
+```python
+forwardMessageFee(ccm):
+    fee = ccm.fee
+    if (fee < 0
+        or getChannel(ccm.sendingChainID).messageFeeTokenID != getChannel(ccm.receivingChainID).messageFeeTokenID):
```
janhack, line 1189: As far as I understand `getChannel` is an API of the interoperability module so we should use the notation `Interoperability.getChannel` (same also in other places).

MaximeGagnebin: Indeed, added.

janhack (commented):
The PR now updates LIP 0047 reverting the change `master` -> `main`. I think that LIP should be excluded from the update.

proposals/lip-0051.md
```diff
@@ -916,6 +869,8 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
   * For each `address, tokenID` pair: 
     * The values used as `moduleID` in the `lockedBalances` array must be unique. 
     * The `lockedBalances` array must be in asscending order of `moduleID`.
+    * All elements of the `lockedBalances` array must have `amount != 0`.
+    * `availableBalance != 0` or `lockedBalances != []`, i.e.: the entry must contain some tokens.
```
janhack, line 873: The sentence is incomplete.

proposals/lip-0051.md
```diff
@@ -1143,26 +1090,31 @@ transferCrossChain(timestamp,
...
         or length(data) > MAX_DATA_LENGTH
         or length(senderAddress) != ADDRESS_LENGTH
         or length(recipientAddress) != ADDRESS_LENGTH
         or availableBalance(senderAddress, tokenID) < amount):
         transferCrossChain fails
 
+    # transfer is only possible for tokens native to either sending or receiving chain
+    # and if there is a direct channel, in which case a forward message will be sent
+    if (chainID not in [CHAIN_ID_ALIAS_NATIVE, receivingChainID]
+        and interoperability.isDirectChannelActive(chainID) == False):
```
janhack, line 1105: Shouldn't it be an `or` between the two conditions?

proposals/lip-0051.md
```diff
@@ -1267,46 +1175,95 @@ unlock(address, moduleID, tokenID, amount):
...
-        availableBalance(relayerAddress, TOKEN_ID_LSK_MAINCHAIN) += fee
-        escrowAmount(CCM.sendingChainID, LOCAL_ID_LSK) -= fee
+        beforeApplyCCM fails
+
+    feeTokenID = interoperability.getChannel(ccm.sendingChainID).messageFeeTokenID
+    # If the chain is the fee token native chain, unescrow and assign fee to relayer
+    if feeTokenID.chainID == interoperability.getOwnChainAccount().ID:
+        if escrowAmount(ccm.sendingChainID, feeTokenID.localID) < fee:
+            interoperability.terminateChain(ccm.sendingChainID)
+            return False
```
janhack, line 1219: In these API, we sometimes write `beforeApplyCCM fails` and sometimes return `False`. Are these two cases different, i.e., do we expect different error handling? To me `fail` would mean throwing an exception, i..e, not intended normal program flow, whereas `return False` is not failing, but simply a return value based on the expected program flow.

I think we should be consistent here and also it should be consistent with the error handling on the caller side. In particular, if in the interoperability module we write  
> if the previous call fails:
>            return False
Then we should fail here if we expect that logic to be triggered.

janhack: I would check the occurrences of `fail` and `False` in the document, as I think they are mixed in other places as well.

proposals/lip-0051.md
```diff
@@ -1267,46 +1175,95 @@ unlock(address, moduleID, tokenID, amount):
...
-        escrowAmount(CCM.receivingChainID, LOCAL_ID_LSK) += fee
+        availableBalance(payFromAddress, canonicalTokenID(feeTokenID)) -= fee
+        escrowAmount(ccm.receivingChainID, feeTokenID.localID) += fee
+    # Else, pay fee
     else:
-        if availableBalance(payFromAddress, TOKEN_ID_LSK) < fee:
+        if availableBalance(payFromAddress, feeTokenID) < fee:
             beforeSendCCM fails
-        availableBalance(payFromAddress, TOKEN_ID_LSK) -= fee
+        availableBalance(payFromAddress, feeTokenID) -= fee
```
janhack, line 1266: I was  first confused why the nested if is necessary and why to once use `canonicalTokenID` and otherwise not. Basically, also the following would work, right? This makes it more clear that you mainly check if the token is native as you then need to escrow.

```
if availableBalance(payFromAddress, canonicalTokenID(feeTokenID)) < fee:
    beforeSendCCM fails
availableBalance(payFromAddress, canonicalTokenID(feeTokenID)) -= fee

if feeTokenID.chainID == interoperability.getOwnChainAccount().ID:
    escrowAmount(ccm.receivingChainID, feeTokenID.localID) += fee
```

proposals/lip-0051.md
```diff
@@ -1267,46 +1175,95 @@ unlock(address, moduleID, tokenID, amount):
...
+
+```python
+beforeRecoverCCM(trsSender, ccm):
+    fee = ccm.fee
+    if fee < 0:
+        beforeRecoverCCM fails
+
+    feeTokenID = interoperability.getChannel(ccm.sendingChainID).messageFeeTokenID
+    # If the chain is the fee token native chain, unescrow and assign fee to relayer
+    if feeTokenID.chainID == interoperability.getOwnChainAccount().ID:
```
janhack, line 1238: If this is not the case, no fees are credited?

IkerAlus (approved):
LGTM. Great effort and work.

proposals/lip-0051.md
```diff
@@ -1171,17 +1123,17 @@ transferCrossChain(timestamp,
         }       
         serializedParams = serialization of messageParams
                            following crossChainTransferMessageParams    
-        interoperability.send(timestamp,
+        interoperability.send(senderAddress,
```
janhack, line 1126: Similar to the NFT module, I think we also need to account for the `send` API failing. Probably it is then best first to send and afterwards to adjust account balances and escrow.

proposals/lip-0051.md
MaximeGagnebin: I added a word, but for me the sentence was quite ok.

proposals/lip-0051.md
MaximeGagnebin: I think here it is correct. You can always send tokens native to the sending or receiving chain. Or, you can send tokens if you have a direct channel (this is how you can send LSK everywhere).

proposals/lip-0051.md
MaximeGagnebin: I think APIs should fail when they can't be resolved (and not return `False`). Iker pointed out that in this case it could still be better to return `False` as the API did not really fail to execute, but rather executed by terminating the chain. 
I think we don't have a clear error handling system in place, and we should probably discuss it before we change the LIP further in that regard.

proposals/lip-0051.md
MaximeGagnebin: Yes, this logic would be the same. This nesting was chosen to separate fully the cases where the fee token is native vs the case when it is not.
For us, it was more clear this way.

proposals/lip-0051.md
MaximeGagnebin: I'm not sure I understand the comment, in either case the fees are credited. I think this comment is in the same line as the previous one.
We decided to split the logic fully between fee in native token and fee non-native.

proposals/lip-0051.md
MaximeGagnebin: Updated similarly to the NFT LIP.

proposals/lip-0051.md
janhack: Now it is clear to me what is meant. But the main clause still does not have a verb.

proposals/lip-0051.md
janhack: Ah, sorry, my initial comment was due to misreading/misunderstanding the function `isDirectChannelActive`.

However, now if  `interoperability.isDirectChannelActive(chainID) == True` I can actually transfer any token? In particular, before we checked that only `LSK` could be transferred to the mainchain.

proposals/lip-0051.md
janhack: But if this condition is not `True` no further code is executed in `beforeRecoverCCM`.

proposals/lip-0051.md
janhack: All good, can be resolved.

proposals/lip-0051.md
MaximeGagnebin: Sure.

proposals/lip-0051.md
MaximeGagnebin: Yes, you can transfer `LSK` through the mainchain, and in the same way, if you have a direct channel with chainA, you can transfer any tokens native to chainA through chainA.

proposals/lip-0051.md
MaximeGagnebin: Sorry, we were talking about different functions, I thought you were talking about `beforeApplyCCM`, where the same line exists.
In this case, we decided to leave the message fee out as it is native to a terminated chain (because the fee is always native to one of the chains of the channel). It would not be a technical problem to credit that fee, it's more a token-economics question.

janhack: For consistency, I would think it is nicer to make this not depend on the type of token, i.e., recovering a message means reverting the state changes that were done due to emitting the message.

proposals/lip-0051.md
janhack: I think this point can be resolved as we handle it as part of the error handling changes.

proposals/lip-0051.md
MaximeGagnebin: Ok, will do the change.

proposals/lip-0051.md
janhack: As discussed, we now consider allow token transfers from any chain to any chain, using the forward message to the native chain if necessary.

proposals/lip-0051.md
```diff
@@ -1143,26 +1088,28 @@ transferCrossChain(timestamp,
...
-            and (chainID == CHAIN_ID_ALIAS_NATIVE or chainID == MAINCHAIN_ID))
         or length(data) > MAX_DATA_LENGTH
         or length(senderAddress) != ADDRESS_LENGTH
         or length(recipientAddress) != ADDRESS_LENGTH
         or availableBalance(senderAddress, tokenID) < amount):
         transferCrossChain fails
 
+    # transfer is only possible for tokens native to either sending or receiving chain
+    # and if there is a direct channel, in which case a forward message will be sent
+    if chainID not in [CHAIN_ID_ALIAS_NATIVE, receivingChainID, TOKEN_ID_LSK]:
```
janhack, line 1102:
```suggestion
    if chainID not in [CHAIN_ID_ALIAS_NATIVE, receivingChainID, TOKEN_ID_LSK.chainID]:
```

proposals/lip-0051.md
janhack: After re-dicussion, we concluded to only allow forwarding of `LSK` tokens as there can be issues with the type of token used for the message fee in other cases.

janhack (approved):
Nice work everyone involved!

karmacoma (approved):

PR #133, opened 2022-02-21
Update LIP0055

ricott1:
- add unsignedBlockHeader schema
- add required signature property in blockHeader schema
- modify signature validation

### DIFF
--- a/proposals/lip-0055.md
+++ b/proposals/lip-0055.md
@@ -209,12 +209,110 @@ These validations are performed before the block is processed and without access
 
 ### Block Header
 
-Block headers are serialized and deserialized accordingly to the following JSON schema.
+#### Block Header JSON Schema
 
-#### JSON Schema
+Block headers are serialized and deserialized accordingly to the following JSON schema.
 
 ```java
 blockHeaderSchema = {
+  "type": "object",
+  "required": [
+    "version",
+    "timestamp",
+    "height",
+    "previousBlockID",
+    "generatorAddress",
+    "transactionRoot",
+    "assetsRoot",
+    "stateRoot",
+    "maxHeightPrevoted",
+    "maxHeightGenerated",
+    "validatorsHash",
+    "aggregateCommit",
+    "signature"
+  ],
+  "properties": {
+    "version": {
+      "dataType": "uint32",
+      "fieldNumber": 1
+    },
+    "timestamp": {
+      "dataType": "uint32",
+      "fieldNumber": 2
+    },
+    "height": {
+      "dataType": "uint32",
+      "fieldNumber": 3
+    },
+    "previousBlockID": {
+      "dataType": "bytes",
+      "fieldNumber": 4
+    },
+    "generatorAddress": {
+      "dataType": "bytes",
+      "fieldNumber": 5
+    },
+    "transactionRoot": {
+      "dataType": "bytes",
+      "fieldNumber": 6
+    },
+    "assetsRoot": {
+      "dataType": "bytes",
+      "fieldNumber": 7
+    },
+    "stateRoot": {
+      "dataType": "bytes",
+      "fieldNumber": 8
+    },
+    "maxHeightPrevoted": {
+      "dataType": "uint32",
+      "fieldNumber": 9
+    },
+    "maxHeightGenerated": {
+      "dataType": "uint32",
+      "fieldNumber": 10
+    },
+    "validatorsHash": {
+      "dataType": "bytes",
+      "fieldNumber": 11
+    },
+    "aggregateCommit": {
+      "type": "object",
+      "fieldNumber": 12,
+      "required": [
+        "height",
+        "aggregationBits",
+        "certificateSignature"
+      ],
+      "properties": {
+        "height": {
+          "dataType": "uint32",
+          "fieldNumber": 1
+        },
+        "aggregationBits": {
+          "dataType": "bytes",
+          "fieldNumber": 2
+        },
+        "certificateSignature": {
+          "dataType": "bytes",
+          "fieldNumber": 3
+        }
+      }
+    },
+    "signature": {
+      "dataType": "bytes",
+      "fieldNumber": 13
+    }
+  }
+}
+```
+
+#### Unsigned Block Header JSON Schema
+
+Unisgned block headers are serialized and deserialized accordingly to the following JSON schema (notice the missing `signature` property).
+
+```java
+unsignedBlockHeaderSchema = {
   "type": "object",
   "required": [
     "version",
@@ -297,10 +395,6 @@ blockHeaderSchema = {
           "fieldNumber": 3
         }
       }
-    },
-    "signature": {
-      "dataType": "bytes",
-      "fieldNumber": 13
     }
   }
 }
@@ -449,7 +543,7 @@ verifyBlockSignature():
   # Remove the signature from the block header
   delete block.header.signature
   # Serialize the block header without signature
-  let serializedUnsignedBlockHeader be the serialization of block.header following the blockHeaderSchema
+  let serializedUnsignedBlockHeader be the serialization of block.header following the unsignedBlockHeaderSchema
 
   let networkIdentifier be the network identifier of the chain
 


MaximeGagnebin (approved):

proposals/lip-0055.md
```diff
@@ -306,6 +307,99 @@ blockHeaderSchema = {
 }
 ```
 
+#### Unsigned Block Header JSON Schema
+
+Unisgned block headers are serialized and deserialized accordingly to the following JSON schema (notice the missing `signature` property).
```
AndreasKendziorra, line 312:
```suggestion
Unsigned block headers are serialized and deserialized accordingly to the following JSON schema (notice the missing `signature` property).
```

AndreasKendziorra (changes requested):
Once we update LIP 27 as planned then the _block signature calculation_ in LIP 29 becomes outdated. Although that's partially the case even without updates to LIP 27. Should we add a _signature calculation_ section here?

proposals/lip-0055.md
```diff
@@ -400,6 +400,24 @@ unsignedBlockHeaderSchema = {
...
+calculateBlockSignature():
+  generatorKey = validators.getValidatorAccount(block.header.generatorAddress).generatorKey
+  let secretKey be the private key corresponding to the public generatorKey
+
+  # Serialize the block header without signature
+  let serializedUnsignedBlockHeader be the serialization of block.header following the unsignedBlockHeaderSchema
+
+  let networkIdentifier be the network identifier of the chain
+
+  return signMessage(secretKey, "LSK_BH_", networkIdentifier, serializedUnsignedBlockHeader)
```
AndreasKendziorra, line 418: I would leave a reference to LIP 37 regarding `signMessage` as it is done for `verifyMessageSig`

proposals/lip-0055.md
```diff
@@ -400,6 +400,24 @@ unsignedBlockHeaderSchema = {
...
 
+
+#### Block Signature Calculation
+
+The block header signature is calculated using the `calculateBlockSignature` function. This function returns a bytes value with length `SIGNATURE_LENGTH_BYTES`.
+
+```python
+calculateBlockSignature():
+  generatorKey = validators.getValidatorAccount(block.header.generatorAddress).generatorKey
+  let secretKey be the private key corresponding to the public generatorKey
```
AndreasKendziorra, line 411: Not important, but should we use "secret key" and "secretKey" (or "private key" and "privateKey") for consistency?

AndreasKendziorra (changes requested):

proposals/lip-0055.md
ricott1: good point, edited

proposals/lip-0055.md
ricott1: added

AndreasKendziorra (approved):

Incede (approved):

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+
+1. **Header creation**: Block header properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+2. **Block initialization**: Block assets properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+3. **Before transactions execution**: Each module can define protocol logic that is executed *before* the transactions contained in the block are processed.
+4. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. From this step until the end of the "before command execution" step, if an error occurs the transaction is invalid and it cannot be included in the block.
+5. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified.
+6. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. After this step is completed, the transaction is guaranteed to be included in the block payload.
```
AndreasKendziorra, line 145: I think it would be nice to add that the transaction can fail

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+7. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed.
+8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed.
+9. **After transactions execution**: Each module can define protocol logic that is executed *after* all the transactions contained in the block have been processed.
+9. **Block sealing**: Block assets properties that require access to the state store *after* any state transitions implied by the block are executed are inserted in this stage.
+11. **Result verification**: Block header properties, which require accessing the state store *after* all state transitions implied by the block have been executed, are inserted.
```
AndreasKendziorra, line 150: > are inserted.

If these properties are inserted instead of verified, then stage should have a different name. For example "Block Header Completion". But it's still weird to add some properties to the block (header) after the "block sealing" stage. Maybe the order of stages is not correct, or we should also rename the "block sealing" stage.

proposals/lip-0055.md
```diff
@@ -14,7 +14,7 @@ Requires: 0040, 0061
```
AndreasKendziorra, line 14: we must also add the "events root" LIP to `Requires` once it is published

AndreasKendziorra (changes requested):

proposals/lip-0055.md
```diff
@@ -39,13 +39,14 @@ Finally, this LIP specifies the validation of all block header properties in one
 
 This LIP introduces the following new block header properties:
 
-- `stateRoot`: The root of the sparse Merkle tree that is computed from the state of the blockchain. See the [Define state model and state root][lip-0040] LIP for the reason why it needs to be included in a block header.
-- `assetsRoot`: The root of the Merkle tree computed from the block assets array. See [below](#separation-between-block-header-and-block-assets) for more details.
 - `generatorAddress`: The address of the block generator. It replaces the `generatorPublicKey` property. See [below](#change-generator-public-key-to-generator-address) for more details.
-- `aggregateCommit`: This property contains the aggregate signature for a certificate for a certain block height. Based on this, any node can create a certificate for the corresponding height. See [LIP 0061][lip-0061] for more details.
+- `assetsRoot`: The root of the Merkle tree computed from the block assets array. See [below](#separation-between-block-header-and-block-assets) for more details.
```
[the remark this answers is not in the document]
MaximeGagnebin: If you agree with this change, there are a couple more occurrences of this term in the text.

proposals/lip-0055.md
```diff
@@ -400,6 +446,26 @@ unsignedBlockHeaderSchema = {
 }
 ```
 
+
+#### Block Signature Calculation
+
+The block header signature is calculated using the `calculateBlockSignature` function. This function returns a bytes value with length `SIGNATURE_LENGTH_BYTES`.
+
+```python
+calculateBlockSignature():
```
MaximeGagnebin, line 455: just wondering, is there a reason not to name this `calculateSignature(block)`.

MaximeGagnebin: I see that the function requires other inputs too, so maybe we can leave it as is. Or go totally the other way and call it `calculateSignature(block, secretKey, networkIdentifier)`.

proposals/lip-0055.md
```diff
@@ -526,6 +572,17 @@ verifyStateRoot():
 
 Here, the function `stateRoot` calculates the state root of the chain at the input height as specified in [LIP 0040][lip-0040#specs].
 
+##### Events Root
+
+The events root is the root of the sparse Merkle tree built from the events emitted during the block processing. It is verified by calling the `verifyeventRoot` function. This function returns a boolean, indicating the success of the check.
+
+```python
+verifyeventRoot():
```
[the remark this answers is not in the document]
MaximeGagnebin: It was plural before, this was changed to align with `transactionRoot` (currently used in the SDK). Any option would be fine for me.

MaximeGagnebin (changes requested):
Only small comments, mainly notation. Nice job!

proposals/lip-0055.md
```diff
@@ -68,6 +69,15 @@ The schema for the block assets allows each module to include its serialized dat
 
 Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetsRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
 
+
+## Failing and Invalid Transactions
+
+In the current protocol, a transaction that is included in a block always has its intended consequences. In particular, if an error occurs during the its processing, the transaction is simply invalid and it is not included in the block payload.
+
+With this LIP, we introduce failing transactions. A failing transaction is one for which an error occurs but which is nevertheless included in a block. 
+In particular, if the error happens after the transaction has been verified and the state transitions happening before the command execution have been processed (see the ["Block processing stage"](#block-processing-stages) section, below), the transaction is considered to be *failed*. Otherwise, the transaction is *invalid* and it is not included in the block (in a way analogous to the current protocol).
```
shuse2, line 78: I think it is better if only command execute step can produce failing transaction.
It would be nice if `afterCommandExecute` will be called on both success and failure case. The example of this scenario is handling gas fee, where gas should be handle in both success/failure cases

shuse2 (commented):
Generally LGTM

proposals/lip-0055.md
```diff
@@ -29,7 +29,7 @@ The second change is to update the block header schema. In general, it is desira
 - does not need to be changed when modules are added or removed to a running blockchain,
 - is used by every blockchain in the Lisk ecosystem regardless of the modules implemented in the individual chains.
 
-Furthermore, we update the block header schema to include new properties introduced by [LIP 0040][lip-0040], [LIP 0058][lip-0058], and [LIP 0061][lip-0061].
+Furthermore, we update the block header schema to include new properties introduced by [LIP 0040][lip-0040], [LIP 0058][lip-0058], [LIP 0061][lip-0061], and [LIP "Introduce events and event root"][lip-events].
```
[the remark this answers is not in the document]
ricott1: yes

proposals/lip-0055.md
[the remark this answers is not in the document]
ricott1: going for singular consistently

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
 14. **Result verification**: Block header properties, which require accessing the state store *after* all state transitions implied by the block have been executed, are verified.
 15. **Block storage**: The block is persisted into the database.
 16. **Peers notification**: Other peers in the P2P network are notified of the new block.
 
+
+### Block Generation Stages
+
+<img src="lip-0055/generation_stages.png" width="70%">
+
+*Figure 2: A schematic depiction of the various stages of a block generation. The steps performed in the consensus domain are indicated by blue boxes, while the steps performed in the state-machine domain are indicated by red boxes. Within the state-machine domain, the transactions processing is indicated by purple boxes. The network domain is indicated by green boxes.*
```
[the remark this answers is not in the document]
ricott1: fixed

proposals/lip-0055.md
```diff
@@ -526,6 +572,17 @@ verifyStateRoot():
 
 Here, the function `stateRoot` calculates the state root of the chain at the input height as specified in [LIP 0040][lip-0040#specs].
 
+##### Events Root
+
+The events root is the root of the sparse Merkle tree built from the events emitted during the block processing. It is verified by calling the `verifyeventRoot` function. This function returns a boolean, indicating the success of the check.
+
+```python
+verifyeventRoot():
+  return block.header.eventRoot == eventRoot(block.header.height)
```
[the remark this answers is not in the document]
ricott1: I wrote this in a way analogous to the state root processing.

proposals/lip-0055.md
ricott1: Indeed, I would leave it like this unless you have a strong opinion

proposals/lip-0055.md
ricott1: I updated the list reflecting the conclusions in the meeting

proposals/lip-0055.md
shuse2: For block generation specifically, probably this is not `Result verification`.
Maybe we can say `Block header update` or `Enclose block header` (i cannot think of good name here)? Also, signing of the block header should happen here

proposals/lip-0055.md
```diff
@@ -224,11 +260,13 @@ blockHeaderSchema = {
     "generatorAddress",
     "transactionRoot",
     "assetsRoot",
```
MaximeGagnebin, line 262:
```suggestion
    "assetRoot",
```

MaximeGagnebin: and I think there are a few more.

MaximeGagnebin (changes requested):
Looks good! There just a few `assetsRoot` remaining.

proposals/lip-0055.md
```diff
@@ -68,6 +69,15 @@ The schema for the block assets allows each module to include its serialized dat
 
 Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetsRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
 
+
+### Failing and Invalid Transactions
+
+In the current protocol, a transaction that is included in a block always has its intended consequences. In particular, if an error occurs during the its processing, the transaction is simply invalid and it is not included in the block payload.
+
+With this LIP, we introduce failing transactions. A failing transaction is one for which an error occurs but which is nevertheless included in a block. 
+In particular, if the error happens after the transaction has been verified and the state transitions happening before the *command execution* have been processed (see the ["Block processing stages"](#block-processing-stages) section below), the transaction is considered to be *failed*. Otherwise, the transaction is *invalid* and it is not included in the block (in a way analogous to the current protocol).
```
shuse2, line 78: should this section also say `afterCommandExecute` will not make transaction fail but make it invalid, or it's implied?

MaximeGagnebin (approved):
Nice job ! LGTM.

proposals/lip-0055.md
ricott1: Sorry, I forgot to push, check it now

Incede (approved):

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
-8. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If the verification fails, the transaction is invalid and the whole block is discarded.
-9. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified.
-10. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed.
-11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed.
-12. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed.
+8. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If an error occurs, the transaction is invalid and the whole block is discarded.
+9. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and it is not included in the block.
```
janhack, line 119: Here and below, I think it would be better to write the following similar to step 8:

> If an error occurs, the transaction is invalid and the whole block is discarded.

The section here is about the block processing, not the block generation, so transactions cannot be removed from a block.

proposals/lip-0055.md
```diff
@@ -300,12 +342,130 @@ blockHeaderSchema = {
...
 ```
 
+
+#### Block Signature Calculation
+
+The block header signature is calculated using the `calculateBlockSignature` function. This function returns a bytes value with length `SIGNATURE_LENGTH_BYTES`.
+
+```python
+calculateBlockSignature():
+  generatorKey = validators.getValidatorAccount(block.header.generatorAddress).generatorKey
```
janhack, line 456: I find it odd that here the Validator module is queried for the public key. Wouldn't you store this information already as part of the generator config (e.g., the EdDSA private key, which will yield both the public key and address).

proposals/lip-0055.md
```diff
@@ -66,7 +67,16 @@ As an example, blockchains created with the Lisk SDK that implement the [Random
...
 
-Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetsRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
+Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
+
+
+### Failing and Invalid Transactions
+
+In the current protocol, a transaction that is included in a block always has its intended consequences. In particular, if an error occurs during its processing as part of the block generation, the transaction is simply invalid and it is not included in the block payload.
```
janhack, line 75:
```suggestion
In the current protocol, a transaction that is included in a block always implies a successful state transition without any error. In particular, if an error occurs during its processing as part of the block generation, the transaction is simply invalid and it is not included in the block payload.
```

janhack: "intended consequences" is not so clear

proposals/lip-0055.md
```diff
@@ -66,7 +67,16 @@ As an example, blockchains created with the Lisk SDK that implement the [Random
...
+
+With this LIP, we introduce failing transactions. A failing transaction is one for which an error occurs but which is nevertheless included in a block.
```
janhack, line 77: I think we should add a more precise definition of valid/invalid and fail here, for instance:

> A transaction is valid, if the hooks `command verification`, `before command execution` and `after command execution` associated to the transaction are executed successfully without error. Otherwise, a transaction is called invalid. Only valid transactions should be added to a block during the block generation, as an invalid transaction makes the block invalid meaning that it would be discarded by any node in the network.

> A valid transaction is executed successfully, if additionally the command execution hook is executed successfully without errors. A valid transaction fails, on the other hand, if an error occurs during the command execution. In this case, all state transitions of the command execution hook are reverted. This means that the transaction has no effect except for those defined in before command execution and after command execution.

janhack: Note that we also do not say anything about which state transitions are reverted. This should also maybe added to the specifications below.

proposals/lip-0055.md
```diff
@@ -300,12 +342,130 @@ blockHeaderSchema = {
...
+      }
     }
   }
 }
 ```
 
+
+#### Block Signature Calculation
+
+The block header signature is calculated using the `calculateBlockSignature` function. This function returns a bytes value with length `SIGNATURE_LENGTH_BYTES`.
```
janhack, line 452: Maybe we could rename `SIGNATURE_LENGTH_BYTES` to `SIGNATURE_LENGTH_EDDSA_BYTES` as we also have BLS signatures which have a different byte size.

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
-10. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed.
-11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed.
-12. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed.
+8. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If an error occurs, the transaction is invalid and the whole block is discarded.
+9. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and it is not included in the block.
+10. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and the whole block is discarded.
+11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and the processing continues.
```
janhack, line 121: I think we should add here that the state transitions are reverted in case of a failure.

proposals/lip-0055.md
```diff
@@ -66,7 +67,16 @@ As an example, blockchains created with the Lisk SDK that implement the [Random
 
 The schema for the block assets allows each module to include its serialized data individually, which makes the inclusion of module data very flexible. Each module can insert a single entry in the assets. This entry is an object containing a `moduleID` property, indicating the ID of the module handling it, and a generic `data` property that can contain arbitrary serialized data.
 
-Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetsRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
+Each entry of the block assets is then inserted in a Merkle tree, whose root is included in the block header as the `assetRoot` property. Inserting the assets root rather than the full assets allows to bound the size of the block header while still authenticating the content of the block assets.
+
+
+### Failing and Invalid Transactions
```
janhack, line 73: The section is currently missing the rationale of why we add this feature. Maybe we could add 2-3 sentences about computationally expensive transactions that can still be cheaply verified in the transaction pool and during block generation.

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+
+In this section, we describe the various stages of a block generation.
+
+Similarly to the block processing, the block generation is split between the network, consensus, and state-machine domains.
+The full generation of a block is organized as follows.
+
+1. **Header creation**: Block header properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+2. **Block initialization**: Block assets properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+3. **Before transactions execution**: Each module can define protocol logic that is executed *before* the transactions contained in the block are processed.
+4. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If an error occurs, the transaction is invalid and it is not included in the block.
```
janhack, line 143: To have less content and make it easier for readers that the same thing happens here as during the block processing, could we omit the state machine hooks and say that these are the same as above?

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+
+*Figure 2: A schematic depiction of the various stages of a block generation. The steps performed in the generator domain are indicated by blue boxes, while the steps performed in the state-machine domain are indicated by red boxes. Within the state-machine domain, the transactions processing is indicated by purple boxes. The network domain is indicated by green boxes.*
+
+In this section, we describe the various stages of a block generation.
+
+Similarly to the block processing, the block generation is split between the network, consensus, and state-machine domains.
+The full generation of a block is organized as follows.
+
+1. **Header creation**: Block header properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+2. **Block initialization**: Block assets properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
```
janhack, line 141: Maybe we could be more concrete and say here that the `block initialization` hook is called for each module, allowing the module to add any information to the block asset. I am not sure if anything else happens here.

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+5. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and it is not included in the block.
+6. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block.
+7. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and the processing continues.
+8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block.
+9. **After transactions execution**: Each module can define protocol logic that is executed *after* all the transactions contained in the block have been processed.
+9. **Block sealing**: Block assets properties that require access to the state store *after* any state transitions implied by the block are executed are inserted in this stage.
```
janhack, line 149: Similar as above, should we maybe explicitly state which module hooks are executed here?

proposals/lip-0055.md
janhack: Note that in the diagram, this is also called result calculation which is inconsistent. I agree with @shuse2 that it would be good to use a different name here (block sealing seems more suitable for this hook, then for the previous one which is only about adding data to the asset).

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+11. **Result verification**: Block header properties, which require accessing the state store *after* all state transitions implied by the block have been executed, are inserted.
+12. **Peers notification**: Other peers in the P2P network are notified of the new block.
```
janhack, line 151: I am not sure about the protocol, but I would assume that here we do both block broadcasting to a subset of peers (similar to forward) as well as peer notification.

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+*Figure 2: A schematic depiction of the various stages of a block generation. The steps performed in the generator domain are indicated by blue boxes, while the steps performed in the state-machine domain are indicated by red boxes. Within the state-machine domain, the transactions processing is indicated by purple boxes. The network domain is indicated by green boxes.*
+
+In this section, we describe the various stages of a block generation.
+
+Similarly to the block processing, the block generation is split between the network, consensus, and state-machine domains.
+The full generation of a block is organized as follows.
+
+1. **Header creation**: Block header properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+2. **Block initialization**: Block assets properties that require access to the state store *before* any state transitions implied by the block are executed are inserted in this stage.
+3. **Before transactions execution**: Each module can define protocol logic that is executed *before* the transactions contained in the block are processed.
```
janhack, line 142: Above, we have hooks/steps for every piece of data added to a block (header, asset), but we somehow omit when/how transactions are added to the block (before they are executed here). I think it would be good to mention this somewhere.

proposals/lip-0055.md
```diff
@@ -105,16 +115,42 @@ The full processing of a block is organized as follows.
...
+10. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and the whole block is discarded.
+11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and the processing continues.
+12. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid and the whole block is discarded.
 13. **After transactions execution**: Each module can define protocol logic that is executed *after* all the transactions contained in the block have been processed.
 14. **Result verification**: Block header properties, which require accessing the state store *after* all state transitions implied by the block have been executed, are verified.
 15. **Block storage**: The block is persisted into the database.
 16. **Peers notification**: Other peers in the P2P network are notified of the new block.
 
+
+### Block Generation Stages
```
janhack, line 129: Currently, we do not give further details below, what actually happens in the block generation steps.

janhack (changes requested):
I left several smaller comments.

proposals/lip-0055.md
ricott1: fixed

proposals/lip-0055.md
ricott1: edited

proposals/lip-0055.md
ricott1: added

proposals/lip-0055.md
ricott1: added a sentence and fixed a typo

proposals/lip-0055.md
ricott1: diagram updated

proposals/lip-0055.md
ricott1: edited in this direction

proposals/lip-0055.md
ricott1: edited

proposals/lip-0055.md
ricott1: edited

proposals/lip-0055.md
ricott1: Mh, I'm not sure how it works in practice. Are transactions preselected or are they added one by one after the "before transactions execution" stage is completed? @shuse2 can you help me?

shuse2: The process is like
1. verify transaction
2. take state snapshot
3. before command execute, execute command, after command execute
4. add to block or restore snapshot and not include the transaction

proposals/lip-0055.md
ricott1: added a new paragraph

proposals/lip-0055.md
```diff
@@ -78,23 +101,23 @@ Name                          | Type    | Value                  | Description
...
+`SIGNATURE_LENGTH_EDDSA_BYTES`      | integer | 64                     | The length of a Ed25519 signature.
 
 Furthermore, in the following we indicate with `block` be the block under consideration and with `previousBlock` the previous block of the chain. Calling a function `fct` from another module `module` is represented by `module.fct`.
 
 ### Block Processing Stages
 
 <img src="lip-0055/stages.png" width="70%">
 
-*Figure 1: A schematic depiction of the various stages of a block processing. The steps performed in the consensus domain are indicated by blue boxes, while the steps performed in the state-machine domain are indicated by red boxes. Within the state-machine domain, the transactions processing is indicated by purple boxes. The network domain is indicated by green boxes.*
+*Figure 1: A schematic depiction of the various stages of a block processing. The stages performed in the consensus domain are indicated by blue boxes, while the stages performed in the application domain are indicated by red boxes. Within the application domain, the transactions processing is indicated by purple boxes. The network domain is indicated by green boxes.*
```
janhack, line 112: The diagram currently contains the "asset finalization" which is actually not part of the block processing, but only the block generation, I believe.

proposals/lip-0055.md
```diff
@@ -60,13 +61,35 @@ The separation between properties in the block header and properties in the bloc
...
+Otherwise, a transaction is *invalid*. 
+Only valid transactions should be added to a block during the block generation, as an invalid transaction makes the whole block invalid meaning that it would be discarded by any node in the network.
+
+A valid transaction is executed successfully if additionally the "command execution" stage is executed successfully without errors. 
+A valid transaction *fails* if  on the other hand an error occurs during the command execution. 
+In this case, all state transitions of the "command execution" stage are reverted. 
+This means that the transaction has no effect except for those defined in "before command execution" and "after command execution".
+The result of the transaction execution is logged using an event emitted at the end of the "after transaction execution" stage, indicating whether the transaction was processed succesfully or an error occured.
+
+Allowing for failing transactions has the advantage that transactions that require complex and expensive "command execution" stage would still pay the fee in case of errors, as the fee is processed in the "before command execution" stage.
```
janhack, line 88:
```suggestion
Allowing for failing transactions has the advantage that transactions that require a complex and expensive "command execution" stage would still pay the fee in case of errors, as the fee is processed in the "before command execution" stage. 
```

proposals/lip-0055.md
```diff
@@ -105,16 +128,44 @@ The full processing of a block is organized as follows.
...
-11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed.
-12. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed.
+8. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If an error occurs, the transaction is invalid and the whole block is discarded.
+9. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and the whole block is discarded.
+10. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and the whole block is discarded.
+11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and the processing continues. All state transitions performed in this stage are reverted.
```
janhack, line 134:
```suggestion
11. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and all state transitions performed in this stage are reverted. In any case, afterwards the processing continues with the next stage.
```

janhack: To make it more explicit that the transitions are only reverted in case of an error.

proposals/lip-0055.md
janhack: But then we currently do not correctly specify what actually happens, which is what Shu explains in https://github.com/LiskHQ/lips/pull/133/commits/f930b57bc303afdc034b67a284823d241bd3ceaa#r853016777.

proposals/lip-0055.md
janhack: All the new names look good from my side. We could resolve this comment.

proposals/lip-0055.md
janhack: Looks good.

proposals/lip-0055.md
ricott1: nice catch, fixed

proposals/lip-0055.md
```diff
@@ -105,16 +128,44 @@ The full processing of a block is organized as follows.
...
+5. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and it is not included in the block.
+6. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block.
+7. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and the processing continues. All state transitions performed in this stage are reverted.
+8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block.
+9. **After transactions execution**: Each module can define protocol logic that is executed *after* all the transactions contained in the block have been processed.
+10. **Assets finalization**: Each module can update or insert new information in the block assets *after* any state transitions implied by the block are executed.
```
janhack, line 164:
```suggestion
10. **Assets finalization**: Each module can update or insert new information in the block assets *after* the state transitions implied by the block are executed.
```

proposals/lip-0055.md
janhack: What I meant here is that for the block processing, we describe in detail what happens in the `static validation`, `header verification`, `result verification`, i.e., the validations for every single block header property. I think we should have something similar for the block generation, i.e., describe how each block property is set and whether it happens in the header initialization or header finalization.

janhack (changes requested):
I went over the new changes and all open comments again. The new naming looks good, is much clearer and more consistent. There are still a few open points from my side.

proposals/lip-0055.md
ricott1: added a sentence

proposals/lip-0055.md
ricott1: Added a paragraph in stage 3

proposals/lip-0055.md
ricott1: Added the section `Block header generation`

proposals/lip-0055.md
shuse2: I think as protocol, we broadcast before execution and here we only do peer notification now

janhack: This is the case in the block processing and mentioned there as step 6. This would not be possible here in the block generation, because the full block object only exists after the header finalization stage.

proposals/lip-0055.md
janhack: Looks good now.

proposals/lip-0055.md
```diff
@@ -300,25 +358,149 @@ blockHeaderSchema = {
...
+
+Here, the function `signMessage` calculates the signature of a message as specified in [LIP 0037][lip-0037#signing-and-verifying-with-ed25519].
+
+#### Block Header Generation
+
+During block generation, the block header properties are set in the following stages.
+
+- **Header initialization**:
+  - Set the `version`, `timestamp`, `height`, `previousBlockID`, `generatorAddress`, `maxHeightPrevoted`, `maxHeightGenerated` properties.
+- **Header finalization**:
```
janhack, line 488: I think the `aggregateCommit` should be moved to the header initialization as it is also verified with respect to the state before applying the state changes implied by the block.

@shuse2 Are all the other properties in the same stage in the implementation?

shuse2: `AggregateCommit` was actually inserted in the finalization step in the implementation now. I will create an issue to fix this.
Otherwise, the list is aligned with the current implementation

proposals/lip-0055.md
```diff
@@ -105,16 +128,45 @@ The full processing of a block is organized as follows.
...
+After this stage has been completed, transactions are selected one-by-one from a transaction pool. The transaction processing stages (stages 4 to 8) are repeated for each transaction selected. If stages 4, 5, 6, and 8 are executed succesfully, the transaction is valid and it is included in the block, otherwise it is invalid and therefore discarded.
+4. **Transaction verification**: Each module can define protocol logic that verifies a transaction, possibly by accessing the state store. If an error occurs, the transaction is invalid and it is not included in the block.
+5. **Command verification**: The command corresponding to the `moduleID`-`commandID` combination is verified. If an error occurs, the transaction is invalid and it is not included in the block.
+6. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block. All state transitions performed in this stage are reverted.
```
janhack, line 161:
```suggestion
6. **Before command execution**: Each module can define protocol logic that is processed *before* the command has been executed. If an error occurs, the transaction is invalid, it is not included in the block and all state transitions induced by the transaction performed up to this stage are reverted.
```

proposals/lip-0055.md
```diff
@@ -105,16 +128,45 @@ The full processing of a block is organized as follows.
...
+7. **Command execution**: The command corresponding to the `moduleID`-`commandID` combination is executed. If an error occurs, the transaction is failed and all state transitions performed in this stage are reverted. In any case, afterwards the processing continues with the next stage.
+8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid and it is not included in the block. All state transitions induced by the transaction performed up to this stage are reverted.
```
janhack, line 163:
```suggestion
8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid, it is not included in the block and all state transitions induced by the transaction performed up to this stage are reverted.
```

proposals/lip-0055.md
janhack: To make it clear that reverting state changes is conditional on the error.

proposals/lip-0055.md
janhack: To make it clear that reverting state changes is conditional on the error.

janhack (changes requested):
There are a few small open points left.

proposals/lip-0055.md
janhack: Can be resolved.

proposals/lip-0055.md
shuse2: ah sorry, i missed the header context here.
Currently for the block generation, it does not do block propagation or peer notification. It passes to the consensus and re-execute the block with the same process as if it received from the other nodes.

ricott1: Then this last step it's wrong, right? it should be something like: "Process block"

proposals/lip-0055.md
ricott1: fixed

proposals/lip-0055.md
ricott1: fixed

proposals/lip-0055.md
```diff
@@ -105,16 +129,45 @@ The full processing of a block is organized as follows.
 5. **Assets verification**: Each module verifies the respective entry in the block assets. If any check fails, the block is discarded and has no further effect.
 6. **Block forwarding**: After the initial checks, the full block is forwarded to a subset of peers.
```
janhack, line 130: This step is not shown in the diagram anymore.

proposals/lip-0055.md
```diff
@@ -105,16 +129,45 @@ The full processing of a block is organized as follows.
...
+8. **After command execution**: Each module can define protocol logic that is processed *after* the command has been executed. If an error occurs, the transaction is invalid, it is not included in the block and all state transitions induced by the transaction performed up to this stage are reverted. In that case, the block generation continues with stage 4 for another transaction from the transaction pool or stage 9.
+9. **After transactions execution**: Each module can define protocol logic that is executed *after* all the transactions contained in the block have been processed.
+10. **Assets finalization**: Each module can update or insert new information in the block assets *after* the state transitions implied by the block are executed.
+11. **Header finalization**: Block header properties, which require accessing the state store *after* all state transitions implied by the block have been executed, are inserted.
+12. **Block processing**: The block goes through the [block processing stages](#block-processing-stages).
```
janhack, line 168: I would suggest to no put this here as a stage, but mention as a commented below this list that a generated block would afterwards be passed to the block processing and will go through the stages defined above. This would yield a cleaner separation in my opinion:

- block generation: produce a block that induces a valid state transition and will be accepted by any node in the network
- block processing: executing the state changes implied by a block

Otherwise, block generation would always be block processing at the same time.

janhack: The specifications above are also currently incorrect, independently of that. Because as written now, after the header finalization the state changes implied by the transactions are already performed. So maybe step 12 should be reverting the state back to before the block processing?

janhack: Looking at the stages again, I am wondering whether the asset finalization actually makes sense together with the block generation process above because all state changes are already applied before.

Specifically, consider the following example: Assume the block reward is in the asset and it would be reduced only in the asset finalization stage. Then the state root computed in the block generation would be different from the one in the block processing (as in the block generation 1 LSK would be credited to the block generator but only 0.25 LSK in the block processing).

proposals/lip-0055.md
ricott1: added

janhack: I still do not see the block forwarding in the diagram.

ricott1: fixed

proposals/lip-0055.md
janhack: As concluded, we remove the asset finalization. This comment can therefore be resolved.

janhack (approved):
Looks all good, nice work!

karmacoma (approved):

karmacoma (approved):

PR #137, opened 2022-03-28
Change the unit naming for byte sizes and make it unambigious

janhack:


sergeyshemyakov (approved):
Looks good!

proposals/lip-0002.md
```diff
@@ -24,10 +24,10 @@ As the Lisk protocol evolves and gains popularity, a clear increase in usage and
 
 ## Rationale
 
-We propose to set a maximum block size of 15KB (kilobytes). This size will maintain the yearly blockchain growth below 50GB, even when blocks are always at full capacity, and the active delegates do not miss any blocks. The summary of the numerical implications are as follows:
+We propose to set a maximum block size of 15 KiB (15 * 1024 bytes). This size will maintain the yearly blockchain growth below 50 GiB, even when blocks are always at full capacity, and the active delegates do not miss any blocks. The summary of the numerical implications are as follows:
 
-* Maximum block size: 15KB.
-* Blockchain size will grow a **maximum** of 47.3GB per year.
+* Maximum block size: 15 KiB.
+* Blockchain size will grow a **maximum** of 47.3 GiB per year.
```
AndreasKendziorra, line 30: If I'm not wrong, it must be 45.1 if the unit is GiB, as 1 GiB = 1024^3 bytes.
And we have a maximum of `(6*60*24*365)*15*1024 bytes = 48439296000 bytes` per year.

BTW: Do we need to incorporate leap years? :)

proposals/lip-0002.md
```diff
@@ -112,17 +112,17 @@ We believe that given these numbers, 15KB is the most adequate choice in terms o
...
 
-5. Update `constants.maxPayloadLength` to 15KB.
+5. Update `constants.maxPayloadLength` to 15 KiB.
 
 6. Remove `constants.maxTxsPerBlock` from `constants.js`.
 
 7. Remove the redundant `numberOfTransactions` check from blocks verification for historical data, as it is already included in block's signature.
 
-8. Remove the redundant `payloadLength` check from blocks verification for historical data (up to 1048576KB), as it is already included in block's signature.
+8. Remove the redundant `payloadLength` check from blocks verification for historical data (up to 1048576 KiB), as it is already included in block's signature.
```
AndreasKendziorra, line 123: Here I'm not sure if changing the unit without changing the value is correct. I actually can't recall right now where the value comes from.

AndreasKendziorra (changes requested):

proposals/lip-0002.md
janhack: I just remove this value, as it is about removing the check and the value is not important for the specifications.

proposals/lip-0002.md
janhack: I just changed the unit name and was assuming the specifications consistently used 1024 as base. I updated this line.

proposals/lip-0002.md
```diff
@@ -27,7 +27,7 @@ As the Lisk protocol evolves and gains popularity, a clear increase in usage and
 We propose to set a maximum block size of 15 KiB (15 * 1024 bytes). This size will maintain the yearly blockchain growth below 50 GiB, even when blocks are always at full capacity, and the active delegates do not miss any blocks. The summary of the numerical implications are as follows:
 
 * Maximum block size: 15 KiB.
-* Blockchain size will grow a **maximum** of 47.3 GiB per year.
+* The block history will grow around 45 GiB per year (approximately 6*60*24*365.25*15/(1024*1024) GiB).
```
AndreasKendziorra, line 30: But "around" instead of "maximum" is rather misleading as we have very few transactions right now. Don't you think?

AndreasKendziorra (changes requested):

proposals/lip-0002.md
janhack: I added "assuming full blocks" to the end of the sentence.

sergeyshemyakov (approved):
Looks good!

karmacoma (approved):

PR #138, opened 2022-03-30
Allow registration of validators without BLS key for migration from Core 3 to Core 4

AndreasKendziorra:
This is needed for the snapshot block on mainnet for the migration from Lisk Core 3 to Lisk Core 4. The reason is that when processing the snapshot block, all delegates must be registered in the validators module. But registration in the validators requires the possession of a valid BLS key and the delegates on the mainchain do not have a BLS key. Therefore, we need to introduce an exception such that the delegates contained in the snapshot block used for migration from Lisk Core to Lisk Core 4 can be registered without having a valid BLS key.


ricott1 (approved):
Review done by pushing commits directly

MaximeGagnebin (approved):
LGTM, nice work!

Incede (approved):

proposals/lip-0057.md
```diff
@@ -1056,19 +1057,32 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
...
-        validator.generatorKey,
-        validator.blsKey
-    )
-    if the above returns False:
-        fail
+# Register all validators in the Validators module.
+# For the snapshot block for the migration from Lisk Core 3 to Lisk Core 4 on 
+# Lisk mainnet, apply a special rule. This must be done as validators in 
+# this snapshot block do not have a BLS key.
+if the chain is the mainchain and b is the snapshot block for the migration:
```
mehmetegemen, line 1067: This will trigger if chain is mainchain but b is not the snapshot block. Maybe you wanted to assign snapshot block to a variable b somewhere else? Because I don't also see `b` is used in below code.

proposals/lip-0044.md
```diff
@@ -54,15 +54,16 @@ In this section, we specify the substores that are part of the Validators module
...
+Name                           | Type    | Value                                                | Description
+------------------------------ | ------- | ---------------------------------------------------- | -------------------------------------------------------------------------------
+`MODULE_ID_VALIDATORS`         | uint32  | TBD                                                  | ID of the Validators module.
+`STORE_PREFIX_VALIDATORS_DATA` | bytes   | 0x0000                                               | Store prefix of the validators data substore.
+`STORE_PREFIX_GENERATOR_LIST`  | bytes   | 0x4000                                               | Store prefix of the generator list substore.
+`STORE_PREFIX_BLS_KEYS`        | bytes   | 0x8000                                               | Store prefix of the registered BLS keys substore.
+`STORE_PREFIX_GENESIS_DATA`    | bytes   | 0xc000                                               | Store prefix of the genesis data substore.
+`INVALID_BLS_KEY`              | bytes   | 48 bytes all set to 0x00                             | An invalid BLS key, used as a placeholder before a valid BLS key is registered.
```
mehmetegemen, line 64: I would expect something `INVALID` to cause error. Maybe we can say `UNREGISTERED_BLS_KEY` as in the explanation? Or `UNSET_BLS_KEY`.

ricott1: We use consistently the term `INVALID` across several LIPs: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0038.md and also in the new migration LIP, therefore I would not change it now.

proposals/lip-0057.md
mehmetegemen: I don't think it's clear, since I ask a question. I understood it now though, can be resolved.

karmacoma (approved):

PR #143, opened 2022-05-05
Update LIP 0050

MaximeGagnebin:
The register BLS key command is updated to also allow the update of the generator key, in case the latter was not set by the migration.
If the generator key was already set during the migration, the public key given in the parameters must equal the one in the state store.

### DIFF
--- a/proposals/lip-0050.md
+++ b/proposals/lip-0050.md
@@ -72,7 +72,7 @@ This substore contains an array with the addresses and the balances of all legac
 * The store prefix is set to `STORE_PREFIX_LEGACY_ACCOUNTS`.
 * The store key is an 8-byte value representing the legacy address.
 * The store value is set to the serialization using `legacyAccountsSchema` of the balance of the legacy account.
-* Notation: For the rest of this proposal, let `legacyAccounts(legacyAddress)` be the legacy accounts substore with store key `legacyAddress`.
+* Notation: For the rest of this proposal, let `legacyAccounts(legacyAddress)` be the legacy accounts substore entry with store key `legacyAddress`.
 
 ##### JSON Schema
 
@@ -113,6 +113,41 @@ def getLegacyAddress(publicKey: PublicKeyEd25519) -> bytes:
     return reversedEightBytes
 ```
 
+### Events
+
+#### AccountReclaimed
+
+This event has `typeID = TYPE_ID_ACCOUNT_RECLAIM`.
+This event is emitted when a legacy account is reclaimed.
+
+##### Topics
+
+* `legacyAddress`: the legacy address of the reclaimed account.
+* `address`: the address of the reclaimed account.
+
+##### Data
+
+```java
+accountReclaimedEventDataSchema = {
+    "type": "object",
+    "required" = ["legacyAddress", "address", "amount"],
+    "properties": {
+        "legacyAddress": {
+            "dataType": "bytes",
+            "fieldNumber": 1
+        },
+        "address": {
+            "dataType": "bytes",
+            "fieldNumber": 2
+        },
+        "amount": {
+            "dataType": "uint64",
+            "fieldNumber": 3
+        },
+    }
+}
+```
+
 ### Commands
 
 #### Reclaim
@@ -144,7 +179,7 @@ reclaimParamsSchema = {
 ##### Verification
 
 ```python
-def verify(trs) -> NoReturn:
+def verify(trs: Transaction) -> None:
     legacyAddress = getLegacyAddress(trs.senderPublicKey)
     if legacyAccounts(legacyAddress) does not exists :
         raise Exception('Public key does not correspond to a reclaimable account.')
@@ -156,12 +191,26 @@ def verify(trs) -> NoReturn:
 ##### Execution
 
 ```python
-def execute(trs) -> NoReturn:
+def execute(trs: Transaction) -> None:
     legacyAddress = getLegacyAddress(trs.senderPublicKey)
     delete legacyAccounts(legacyAddress) from the legacy accounts substore
 
     newAddress = 20-byte address derived from trs.senderPublicKey
     token.mint(newAddress, LOCAL_ID_LSK, trs.params.amount)
+
+    emitEvent(
+        moduleID=MODULE_ID_LEGACY,
+        typeID=TYPE_ID_ACCOUNT_RECLAIM,
+        topics=[
+            legacyAddress, 
+            newAddress
+        ],
+        data={
+            "legacyAddress": legacyAddress,
+            "address":newAddress,
+            "amount":trs.params.amount
+        }
+    )
 ```
 
 #### Register Keys
@@ -199,7 +248,7 @@ registerBLSKeyParamsSchema = {
 #### Verification
 
 ```python
-def verify(trs) -> NoReturn:
+def verify(trs: Transaction) -> None:
     validatorAddress = 20-byte address derived from trs.senderPublicKey
 
     if validator.getValidatorAccount(validatorAddress) does not exists :
@@ -218,7 +267,7 @@ def verify(trs) -> NoReturn:
 ##### Execution
 
 ```python
-def execute(trs) -> NoReturn:
+def execute(trs: Transaction) -> None:
     validatorAddress = 20-byte address derived from trs.senderPublicKey
         
     if validator.getValidatorAccount(validatorAddress).generatorKey == INVALID_ED25519_KEY:
--- a/proposals/lip-0050.md
+++ b/proposals/lip-0050.md
@@ -47,8 +47,8 @@ We define the following constants:
 | `MODULE_ID_LEGACY`             | bytes | TBD    | Module ID of the Legacy module.                                                                            |
 | `COMMAND_ID_RECLAIM`           | bytes | 0x0000 | Command ID of the reclaim command.                                                                         |
 | `COMMAND_ID_REGISTER_KEYS`     | bytes | 0x0001 | Command ID of the register keys command.                                                                   |
-| `TYPE_ID_ACCOUNT_RECLAIM`      | bytes | 0x0000 | The type ID of the account reclaim event.                                                                  |
-| `TYPE_ID_KEYS_REGISTERED`      | bytes | 0x0001 | The type ID of the keys registered event.                                                                  |
+| `TYPE_ID_ACCOUNT_RECLAIM`      | bytes | 0x0001 | The type ID of the account reclaim event.                                                                  |
+| `TYPE_ID_KEYS_REGISTERED`      | bytes | 0x0002 | The type ID of the keys registered event.                                                                  |
 | `STORE_PREFIX_LEGACY_ACCOUNTS` | bytes | 0x0000 | Store prefix of the legacy accounts substore. This contains the addresses and balances of legacy accounts. |
 | `LOCAL_ID_LSK`                 | bytes | 0x00000000 | Local identifier of the LSK token.                                                                     |
 | `TOKEN_ID_LSK_MAINCHAIN`       | bytes | 8 bytes all set to `0`  | Identifier of the LSK token on the Lisk mainchain.                                        |
--- a/proposals/lip-0050.md
+++ b/proposals/lip-0050.md
@@ -150,7 +150,7 @@ When a reclaim transaction `trs` is executed, the following is done:
 
 #### Register Keys
 
-This command allows migrated legacy validators register the required keys. In particular, this command is used by all delegates to register a BLS key. This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module. Transactions executing this command have:
+This command allows migrated legacy validators register the required keys. In particular, this command is used by all migrated delegates to register a BLS key. This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module. Transactions executing this command have:
 
 * `moduleID = MODULE_ID_LEGACY`,
 * `commandID = COMMAND_ID_REGISTER_KEYS`.
--- a/proposals/lip-0050.md
+++ b/proposals/lip-0050.md
@@ -46,7 +46,7 @@ We define the following constants:
 |---------------------------------|---------| -------|------------------------------------------------------------------------------------------------------------|
 | `MODULE_ID_LEGACY `             | uint32  | TBD    | Module ID of the Legacy module.                                                                            |
 | `COMMAND_ID_RECLAIM `           | uint32  | 0      | Command ID of the reclaim command.                                                                         |
-| `COMMAND_ID_REGISTER_BLS_KEY `  | uint32  | 1      | Command ID of the register BLS key command.                                                                |
+| `COMMAND_ID_REGISTER_KEYS `     | uint32  | 1      | Command ID of the register keys command.                                                                |
 | `STORE_PREFIX_LEGACY_ACCOUNTS ` | bytes   | 0x0000 | Store prefix of the legacy accounts substore. This contains the addresses and balances of legacy accounts. |
 
 #### Functions from Other Modules
@@ -148,12 +148,12 @@ When a reclaim transaction `trs` is executed, the following is done:
 * Call the function `mint(newAddress, 0, trs.params.amount)`, where `newAddress` is the [20-byte address][lip-0018#addressComputation] derived from `trs.senderPublicKey`.
   The function `mint` is defined in the [Token module][lip-0051#mint].
 
-#### Register BLS Key
+#### Register Keys
 
-This command allows migrated legacy validators without a BLS key to register one. This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module. Transactions executing this command have:
+This command allows migrated legacy validators register the required keys. In particular, this command is used by all delegates to register a BLS key. This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module. Transactions executing this command have:
 
 * `moduleID = MODULE_ID_LEGACY`,
-* `commandID = COMMAND_ID_REGISTER_BLS_KEY`.
+* `commandID = COMMAND_ID_REGISTER_KEYS`.
 
 ##### Parameters
 
@@ -171,6 +171,10 @@ registerBLSKeyParamsSchema = {
         "proofOfPossession": {
             "dataType": "bytes",
             "fieldNumber": 2
+        },
+        "generatorKey": {
+            "dataType": "bytes",
+            "fieldNumber": 3
         }
     }
 }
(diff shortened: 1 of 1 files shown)


shuse2 (approved):

proposals/lip-0050.md
```diff
@@ -171,14 +171,36 @@ registerBLSKeyParamsSchema = {
...
+            "fieldNumber": 3
         }
     }
 }
 ```
 
 ##### Execution
 
-Executing a transaction `trs` triggering an register BLS key command is done by calling the function `setValidatorBLSKey(validatorAddress, trs.params.proofOfPossession, trs.params.blsKey)` where `validatorAddress` is the 20-byte address derived from `trs.senderPublicKey` and `setValidatorBLSKey` is defined in the [Validators module][lip-0044#setValidatoBLSKey]. If this function returns false, the transaction is invalid.
+Executing a transaction `trs` triggering an register keys command is done by calling the logic below
```
ricott1, line 185:
```suggestion
Executing a transaction `trs` triggering an register keys command is done by calling the logic below.
```

proposals/lip-0050.md
```diff
@@ -171,14 +171,36 @@ registerBLSKeyParamsSchema = {
...
+
+```python
+validatorAddress = 20-byte address derived from trs.senderPublicKey
+
+if validator.getValidatorAccount(address) does not exists :
```
ricott1, line 190:
```suggestion
if validator.getValidatorAccount(validatorAddress) does not exists :
```

ricott1: and also in the rest of the code I think

proposals/lip-0050.md
```diff
@@ -171,14 +171,36 @@ registerBLSKeyParamsSchema = {
...
 }
 ```
 
 ##### Execution
 
-Executing a transaction `trs` triggering an register BLS key command is done by calling the function `setValidatorBLSKey(validatorAddress, trs.params.proofOfPossession, trs.params.blsKey)` where `validatorAddress` is the 20-byte address derived from `trs.senderPublicKey` and `setValidatorBLSKey` is defined in the [Validators module][lip-0044#setValidatoBLSKey]. If this function returns false, the transaction is invalid.
+Executing a transaction `trs` triggering an register keys command is done by calling the logic below
+
+```python
+validatorAddress = 20-byte address derived from trs.senderPublicKey
```
ricott1, line 188: already going in the direction of having valid pytohn code, we could have something like `trs.senderPublicKey.to_address()`, where the sender public key is an instance of the public key class, which has this functionality (see, eg.: https://github.com/ricott1/dingus/blob/main/dingus/types/keys.py#L22)

ricott1 (changes requested):
some typos in the code, in general, should we wait for a conclusion on the LIP-code style to finalize this?

proposals/lip-0050.md
MaximeGagnebin: yes, that's a nice idea. Though I think I would do this change when we update to the more valid python code.

proposals/lip-0050.md
```diff
@@ -171,14 +171,36 @@ registerBLSKeyParamsSchema = {
         "proofOfPossession": {
             "dataType": "bytes",
             "fieldNumber": 2
+        },
+        "generatorKey": {
```
sergeyshemyakov, line 175: The `"required"` object in the schema has `"blsKey"` instead of `"generatorKey"`, needs to be changed.

proposals/lip-0050.md
```diff
@@ -171,14 +171,36 @@ registerBLSKeyParamsSchema = {
...
+    setGeneratorKey = validator.setValidatorGeneratorKey(validatorAddress, trs.params.generatorKey)
+elif validator.getValidatorAccount(validatorAddress).generatorKey == trs.params.generatorKey):
+    setGeneratorKey = True
+    
+setBLSKey = validator.setValidatorBLSKey(validatorAddress, trs.params.proofOfPossession, trs.params.blsKey)
+
+return (setBLSKey and setGeneratorKey)
+```
+
+Those functions are defined in the [Validators module][lip-0044#setValidatoBLSKey]. If the above logic returns false, the transaction is invalid.
```
sergeyshemyakov, line 203: > If the above logic returns false, the transaction is invalid.

I think, in the spirit of error handling we should raise an error if the transaction is invalid. 
For example, consider the following piece of code:

```python
existingKey =  validator.getValidatorAccount(validatorAddress).generatorKey
if existingKey is empty or existingKey == INVALID_ED25519_KEY:
     setGeneratorKey = validator.setValidatorGeneratorKey(validatorAddress, trs.params.generatorKey)
     if not setGeneratorKey:
         raise Error("Error when setting generator key")
else:
    raise Error("Generator key is already set")
```

Or maybe this logic could be finalized after we introduce the error handling logic to Validators module.

proposals/lip-0050.md
```diff
@@ -46,7 +46,7 @@ We define the following constants:
 |---------------------------------|---------| -------|------------------------------------------------------------------------------------------------------------|
 | `MODULE_ID_LEGACY `             | uint32  | TBD    | Module ID of the Legacy module.                                                                            |
```
sergeyshemyakov, line 47: We could already change ID type to `bytes` here (module ID and two command IDs).

sergeyshemyakov (changes requested):

proposals/lip-0050.md
MaximeGagnebin: good catch! fixed.

sergeyshemyakov (approved):
Looks good now!

proposals/lip-0050.md
MaximeGagnebin: After yesterday's meeting, i would leave it as is.

proposals/lip-0050.md
```diff
@@ -190,21 +291,17 @@ This module does not expose any functions.
...
 ##### Execution
 
 ```python
-getLegacyAccount(publicKey)
-    let legacyAddress be the legacy address of publicKey
+def getLegacyAccount(publicKey: PublicKeyEd25519) -> dict:
+    legacyAddress = getLegacyAddress(publicKey)
     if there exists no entry in the legacy accounts substore with store key equal to legacyAddress:
-        return undefined
+        return {}
```
shuse2, line 304: for the endpoint, maybe it's better to return empty value like
```
{
  legacyAddress,
  balance: '0'
}
```

proposals/lip-0050.md
```diff
@@ -138,22 +178,50 @@ reclaimParamsSchema = {
 
 ##### Verification
 
-For a reclaim transaction `trs` to be valid, the legacy accounts substore must contain an entry with store key equal to `getLegacyAddress(trs.senderPublicKey)` and store value `{"balance": trs.params.amount}`, serialized using `legacyAccountsSchema`.
+```python
```
shuse2, line 181: not related to this PR, but we could remove `amount` from the param (ie: use empty param)

sergeyshemyakov: I agree with Shu, it doesn't make sense to pass a value for `amount` but fail verification if legacy account balance is not equal to `amount`.

proposals/lip-0050.md
```diff
@@ -171,14 +239,47 @@ registerBLSKeyParamsSchema = {
...
+
+```python
+def verify(trs: Transaction) -> None:
+    validatorAddress = 20-byte address derived from trs.senderPublicKey
+
+    if validator.getValidatorAccount(validatorAddress) does not exists :
+        raise Exception('Public key does not correspond to a registered validator.')
+        
+    validatorAccount = validator.getValidatorAccount(validatorAddress)
+    if (validatorAccount.generatorKey != INVALID_ED25519_KEY
```
sergeyshemyakov, line 261: The constant `INVALID_ED25519_KEY` is not defined in this LIP. Add it to the constants. Also probably some rationale and explanation would be helpful.

proposals/lip-0050.md
```diff
@@ -171,14 +239,47 @@ registerBLSKeyParamsSchema = {
...
+        and validatorAccount.generatorKey == trs.params.generatorKey
```
sergeyshemyakov, line 262: Could it be somehow that the validator account doesn't have `generatorKey` set? Also, I don't understand this line of the `if` condition, if `validatorAccount.generatorKey == trs.params.generatorKey` then input generator key equals the one set in the store.
I think of a condition like this: ```if validatorAccount.generatorKey exists and validatorAccount.generatorKey != INVALID_ED25519_KEY```.

proposals/lip-0050.md
```diff
@@ -171,14 +239,47 @@ registerBLSKeyParamsSchema = {
...
+    ):
+        raise Exception('Input generator key does not equal the one set in the store.')
+
+    if validatorAccount.blsKey != INVALID_BLS_KEY:
```
sergeyshemyakov, line 266: Can `blsKey` not exist?

proposals/lip-0050.md
```diff
@@ -171,14 +239,47 @@ registerBLSKeyParamsSchema = {
...
+
 ##### Execution
 
-Executing a transaction `trs` triggering an register BLS key command is done by calling the function `setValidatorBLSKey(validatorAddress, trs.params.proofOfPossession, trs.params.blsKey)` where `validatorAddress` is the 20-byte address derived from `trs.senderPublicKey` and `setValidatorBLSKey` is defined in the [Validators module][lip-0044#setValidatoBLSKey]. If this function returns false, the transaction is invalid.
+```python
+def execute(trs: Transaction) -> None:
+    validatorAddress = 20-byte address derived from trs.senderPublicKey
+        
+    if validator.getValidatorAccount(validatorAddress).generatorKey == INVALID_ED25519_KEY:
+        setGeneratorKey = validator.setValidatorGeneratorKey(validatorAddress, trs.params.generatorKey)
```
sergeyshemyakov, line 277: There is no usage for `setGeneratorKey`, remove the variable.

proposals/lip-0050.md
```diff
@@ -190,21 +291,17 @@ This module does not expose any functions.
 
 This function provides the legacy address and balance of the corresponding legacy accounts.
 
-##### Parameters
-
-* inputParameter1: `publicKey`, an Ed25519 public key
-
 ##### Returns
 
 Either an object with the properties `legacyAddress` and `balance`, where `legacyAddress` is the legacy address for `publicKey` and `balance` is the balance of the corresponding legacy account, or `undefined`.
```
sergeyshemyakov, line 296: In this version of the function you don't return `undefined` anymore. Say that you return empty object if there is no legacy account.

sergeyshemyakov (changes requested):

proposals/lip-0050.md
```diff
@@ -90,24 +99,55 @@ This substore contains an entry for each legacy address for which no [reclaim tr
...
+            "dataType": "uint64",
+            "fieldNumber": 3
+        },
+    }
+}
+```
+
 ### Commands
 
 #### Reclaim
```
janhack, line 153: As we wanted to use this PR to align on module LIPs, some style comments:

- I think it would be sufficient to write "This command has module ID `COMMAND_ID_RECLAIM`" as the module ID is clear from the context.
- I think the sentence "The `params` property of a reclaim transaction must obey the following schema:" can be removed.

proposals/lip-0050.md
janhack: I think this is an artifact from the time that this information was needed for undoing the transaction. I agree that we could remove it now.

proposals/lip-0050.md
janhack: Maybe it should be `!=` here assuming that if your generator key is already set, then you have to put your existing generator key in the transaction.

proposals/lip-0050.md
janhack: Yep, on Mainnet all BLS keys are set to the invalid one at regenesis.

janhack (changes requested):
The newly introduced events and updated logic with type hinting looks good, just minor comments.

Two general comments regarding the protocol:

- Regarding setting generator keys: An alternative solution would be that for all accounts without BLS key we allow to set any generator key. This means the delegates on Lisk Mainchain would be able to choose a new generator key once. The only downside would be that the bootstrap delegates could not be punished with a PoM if they change their generator key fast enough.
- Regarding reclaim and minting tokens: My preference would be that all unclaimed tokens would be kept in one account and locked with the legacy account ID as this would mean that the total supply in the Token module is actually correct .

proposals/lip-0050.md
lisk-is-vital: With the new events, it is indeed not needed. However, there is no harm in leaving it there and I think I would rather not change the transaction params unless there is a good reason to do so.

proposals/lip-0050.md
MaximeGagnebin: done

proposals/lip-0050.md
MaximeGagnebin: Sure, we could do that.

proposals/lip-0050.md
MaximeGagnebin: Not sure what you mean. As Jan said, if you don't have a BLS key, the regenesis process sets it to `INVALID`, but if you register after the migration, you must provide a valid BLS key.

proposals/lip-0050.md
MaximeGagnebin: Added it to the table. I will keep the explanation minimal here, the reasoning behind the existence of `INVALID_ED25519_KEY` should be in the migration LIP.

proposals/lip-0050.md
MaximeGagnebin: The generator key always exists (as do other properties of our schemas), the question is only:
- does it equal the INVALID key set during migration?
if not, the transaction sender should provide the already set key.

As Jan mentioned, we could relax this requirement and always allow this transaction to set the generator key. Even in the case when it is already set.

proposals/lip-0050.md
MaximeGagnebin: Changed that, also taking into account the comment by Shu.

proposals/lip-0050.md
MaximeGagnebin: Did we align on this? I would say that this is a level of detail that can be left to the LIP author and does not need to be uniform.

proposals/lip-0050.md
MaximeGagnebin: As concluded in our tactical meeting, we will not change this command.

proposals/lip-0050.md
MaximeGagnebin: Please double check the value.

proposals/lip-0050.md
MaximeGagnebin: Updated to always allow to update the generator key, if the BLS key does not already exist.

proposals/lip-0050.md
janhack: Nothing mandatory, just optional suggestions. I resolve it.

proposals/lip-0050.md
janhack: Can be resolved from my side.

proposals/lip-0050.md
```diff
@@ -42,12 +42,24 @@ The Legacy module has the module ID `MODULE_ID_LEGACY` (defined in the table bel
 
 We define the following constants:
 
-| Name                            | Type    | Value  | Description                                                                                                |
-|---------------------------------|---------| -------|------------------------------------------------------------------------------------------------------------|
```
janhack, line 46: The type IDs for events seem missing.

proposals/lip-0050.md
janhack: Can be resolved from my side.

proposals/lip-0050.md
```diff
@@ -248,8 +408,8 @@ Then, do the following:
 
 - Check if the `address` properties of the entries in `genesisBlockAssetObject.accounts` are pairwise distinct. If not, reject the block.
 - Check if the sum of the `balance` properties of the entries in `genesisBlockAssetObject.accounts` is less than 2<sup>64</sup>. If not, reject the block.
+- Check if the sum of the `balance` properties of the entries in `genesisBlockAssetObject.accounts` equals `token.getLockedAmount(ADDRESS_LEGACY_RESERVE, MODULE_ID_LEGACY, TOKEN_ID_LSK_MAINCHAIN)`.
```
janhack, line 411: If we want to do it like this, we need to update LIP 63 to create this account with the correct balance. Alternatively, we could instead mint the amount of tokens here and lock it (in the genesis state finalization stage). 

The advantage of the current approach is that the genesis block asset of the Token module contains the correct genesis supply of `LSK`. The advantage of the alternative approach is that it would make the genesis state of the Token module and Legacy module more independent (the genesis asset of the Token module does not need to be initialized based on the state of the legacy module) and it also would not require an update of LIP 63.

@sergeyshemyakov @MaximeGagnebin What do you guys think?

MaximeGagnebin: Personally, I prefer the current writing.

proposals/lip-0050.md
MaximeGagnebin: good catch.

proposals/lip-0050.md
MaximeGagnebin: As discussed in our stand up, we would leave the writing as is. The migration LIP should then be updated to create the account with `address = ADDRESS_LEGACY_RESERVE` and `locked balance = sum of all reclaim balances` in the genesis token asset.

proposals/lip-0050.md
sergeyshemyakov: The value is correct.

sergeyshemyakov (commented):
It was not completely clear to me whether it is possible to set two keys separately by a register keys command.

proposals/lip-0050.md
```diff
@@ -162,23 +279,70 @@ The `params` property of a register BLS key transaction must obey the following
...
+
+```python
+def verify(trs: Transaction) -> None:
+
+    validatorAddress = 20-byte address derived from trs.senderPublicKey
+
+    if validator.getValidatorAccount(validatorAddress) does not exists :
+        raise Exception('Public key does not correspond to a registered validator.')
+
+    if validatorAccount.blsKey != INVALID_BLS_KEY:
```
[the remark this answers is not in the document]
MaximeGagnebin: yes, changed.

proposals/lip-0050.md
```diff
@@ -162,23 +279,70 @@ The `params` property of a register BLS key transaction must obey the following
...
+```python
+def execute(trs: Transaction) -> None:
+    validatorAddress = 20-byte address derived from trs.senderPublicKey
+
+    validator.setValidatorGeneratorKey(validatorAddress, trs.params.generatorKey)
+
+    # the calls below will raise an exception if the proof of possession is invalid
+    # with respect to the given BLS key.
+    validator.setValidatorBLSKey(validatorAddress, trs.params.proofOfPossession, trs.params.blsKey)
+    emitEvent(
```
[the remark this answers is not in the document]
MaximeGagnebin: An event would be emitted by the validator module when the setting of the BLS key fails, this event should be a persistent event. So that it will stay in the event tree.
In that case, the command has no effect and no keys will be set.

proposals/lip-0050.md
```diff
@@ -162,23 +279,70 @@ The `params` property of a register BLS key transaction must obey the following
 ```java
 registerBLSKeyParamsSchema = {
     "type": "object",
-    "required": ["blsKey", "proofOfPossession"],
+    "required": ["blsKey", "proofOfPossession", "generatorKey"],
```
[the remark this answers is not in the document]
MaximeGagnebin: Exactly, they must set both at the same time.
I expanded a bit in the rationale about that.

proposals/lip-0050.md
```diff
@@ -138,47 +223,126 @@ reclaimParamsSchema = {
...
+    delete legacyAccounts(legacyAddress) from the legacy accounts substore
+
+    newAddress = 20-byte address derived from trs.senderPublicKey
+    # Unlock the tokens in the legacy reserve account and transfer them to the 
+    # account reclaiming the tokens.
+    token.unlock(ADDRESS_LEGACY_RESERVE, MODULE_ID_LEGACY, TOKEN_ID_LSK_MAINCHAIN, trs.params.amount)
+
+    token.transfer(ADDRESS_LEGACY_RESERVE, newAddress, TOKEN_ID_LSK_MAINCHAIN, trs.params.amount)
+
+    emitEvent(
```
ricott1, line 250: I think here we should drop the named fields and just write:

emitEvent(
        MODULE_ID_LEGACY,
        TYPE_ID_ACCOUNT_RECLAIM,
        serialize(
            {
                "legacyAddress": legacyAddress,
                "address":newAddress,
                "amount":trs.params.amount
            }, 
            accountReclaimedEventDataSchema
        ),
        [
            legacyAddress, 
            newAddress
        ]
    )

MaximeGagnebin: Maybe you could point me to a reference that says that using named fields should be avoided?

ricott1 (approved):

karmacoma (approved):

PR #145, opened 2022-05-30
Update LIP 0037 and LIP 0041

ricott1:
- Update LIP 0037 to make it easier to read all message tags from a single source
- Update LIP 0041 to include error handling, events, and new multisig group registration format


proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
+    # Mandatory and optional keys must be disjoint.
+    all_keys = trs.params.mandatoryKeys + trs.params.optionalKeys
+    if len(all_keys) != len(set(all_keys)):
+        raise Exception('Mandatory and optional keys arrays are not disjoint.')
+    # Union of mandatory and optional keys must contain between 1 and 64 keys (included).
+    if len(all_keys) < 1 or len(all_keys) > 64:
+        raise Exception('Invalid length of mandatory and optional keys arrays.')
+    # Number of signatures must be an integer between 1 and the total number of public keys.
+    if trs.params.numberOfSignatures < 1 or trs.params.numberOfSignatures > len(all_keys):
+        raise Exception('Invalid number of signatures.')
```
janhack, line 270: As we have also a `signatures` property in the transaction, this may be confusing. Consider:
- Changing the error to "Invalid numberOfSignatures property".

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
+    # Number of signatures must not be smaller than the number of mandatory keys.
+    if trs.params.numberOfSignatures < len(trs.params.mandatoryKeys):
+        raise Exception('Invalid number of signatures.')
```
janhack, line 273: Same as above

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
+            "mandatoryKeys": trs.params.mandatoryKeys,
+            "optionalKeys": trs.params.optionalKeys
+        }
+    )
+    for signature, key in zip(signatures, all_keys):
+        verifyMessageSig(key, MULTISIGNATURE_MSG_TAG, networkIdentifier, message, signature)
+```
+
+The function `verifyMessageSig` is defined in [LIP 0062](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0062.md#specification) and `networkIdentifier` is the network identifier of the chain.
+ schema `multisignatureGroupDataSchema` is
```
janhack, line 291: Incomplete sentence.

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
+        {
+            "address": senderAddress,
+            "nonce": trs.nonce,
+            "numberOfSignatures": trs.params.numberOfSignatures,
+            "mandatoryKeys": trs.params.mandatoryKeys,
+            "optionalKeys": trs.params.optionalKeys
+        }
+    )
+    for signature, key in zip(signatures, all_keys):
+        verifyMessageSig(key, MULTISIGNATURE_MSG_TAG, networkIdentifier, message, signature)
```
janhack, line 287: I would do this step in the command execution as it is more computationally intense (considering 64 signatures) and any error here could be caught when dry-running.

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
 
-This command updates the entries in the sender `authAccount` to the values specified in the command parameters.
+```python
+def execute(trs: Transaction) -> None:
+    senderAddress = 20-byte address derived from trs.senderPublicKey
+    authAccount(senderAddress).numberOfSignatures = trs.params.numberOfSignatures
+    authAccount(senderAddress).mandatoryKeys = trs.params.mandatoryKeys
+    authAccount(senderAddress).optionalKeys = trs.params.optionalKeys
+    
+    emitEvent(
```
janhack, line 302: What is the rationale for emitting this event if it only contains data already included in the transaction and the only topic is the transaction sender?

proposals/lip-0041.md
```diff
@@ -250,16 +452,15 @@ The Auth module does not execute any logic during the genesis state finalization
 
 #### Transaction Verification
```
janhack, line 453: I think this section should be updated as the signature verification is not the same any more as in LIP 17 as there is no exceptional case any more for the multisignature registration transaction. I think it would be best to have one new clean function here. The `verifySignatures` endpoint could then also point to the function.

proposals/lip-0041.md
```diff
@@ -268,25 +469,23 @@ Notice that a transaction in the transaction pool with a nonce greater than the
...
-Before the command referenced in a transaction `trs` is executed, the following protocol logic is applied.
-
-* Create an auth account if no account exists for the sender address:
+Before the command referenced in a transaction `trs` is executed, an auth account is created if no account exists for the sender address.
 
 ```python
-let senderAddress be the address corresponding to trs.senderPublicKey
-if there is no entry in the in the auth data substore with storeKey == senderAddress:
-  create an entry in the auth data substore with storeKey = senderAddress, initialized to default values
+def beforeCommandExecution(trs: Transaction) -> None:
```
janhack, line 475: Is there is a specific reason why this logic is performed here and split from incrementing the nonce?

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
-The verification of this command is defined in [LIP 0017](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0017.md#multisignature-account-registration).
+These specifications supersede those defined in [LIP 0017](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0017.md#multisignature-account-registration).
+
+```python
+def verify(trs: Transaction) -> None:
+    # Check that a multisignature group is not already registered for this address
+    senderAddress = 20-byte address derived from trs.senderPublicKey
+    if len(authAccount(senderAddress).mandatoryKeys) > 0 or len(authAccount(senderAddress).mandatoryKeys) > 0:
+        raise Exception('Multisignature group already registered for this address.')
+    # Mandatory keys are pairwise distinct and ordered lexicographically.
```
janhack, line 255: Maybe we can say that the `mandatoryKeys` array may be empty (this is implied by the Python code, but maybe best o mention it explicitly).

proposals/lip-0041.md
```diff
@@ -250,16 +452,15 @@ The Auth module does not execute any logic during the genesis state finalization
...
 ```python
-isValidNonce(trs):
-  let senderAddress be the address corresponding to trs.senderPublicKey
-  if there is no entry in the in the auth data substore with storeKey == senderAddress:
-    return trs.nonce == 0
-  return trs.nonce == authAccount(senderAddress).nonce
+def transactionVerification(trs: Transaction) -> None:
+  senderAddress = 20-byte address derived from trs.senderPublicKey
+  if there is no entry in the in the auth data substore with storeKey == senderAddress and trs.nonce != 0:
+    raise Exception('Invalid transaction nonce.')
```
janhack, line 461: Optional detail: Here and below we could even say which nonce is expected.

janhack (changes requested):
Nice work. I just left a few small comments.

proposals/lip-0041.md
shuse2: I think it would be nice if the state changes are communicated though events

shuse2 (approved):
Other than @janhack 's comments, LGTM

proposals/lip-0041.md
```diff
@@ -39,8 +39,12 @@ We define the following constants:
 
 | Name               | Type    | Value       | Description                             |
 | ------------------ |---------| ------------| ----------------------------------------|
-|`MODULE_ID_AUTH`    | uint32  | TBD         | ID of the Auth module.                  |
+|`MODULE_ID_AUTH`    | bytes  | TBD         | ID of the Auth module.                  |
 |`STORE_PREFIX_AUTH` | bytes   | 0x0000      | Store prefix of the auth data substore. |
+|`COMMAND_ID_REGISTER_MULTISIGNATURE_GROUP` | bytes   | 0x0000      | Command ID of the register multisignature group command. |
+|`TYPE_ID_MULTISIGNATURE_GROUP_REGISTERED` | bytes   | 0x0000      | Type ID of the signature group registered event. |
```
gkoumout, line 45: If I understand well the Events LIP, for each module the type_ID value 0x0000 is allocated for the standard event which indicates the success of the transaction (constant `EVENT_TYPE_ID_STANDARD`). If I am not missing something, this means that the value of `TYPE_ID_MULTISIGNATURE_GROUP_REGISTERED` should change..

gkoumout (commented):
Some suggestions, with no strong opinion:

- If I understand well, Auth Module replaces Keys and Sequence modules (according to current Lisk protocol documentation). I suggest mentioning this either in the abstract or perhaps better in the very end, when the hard-fork is explained.

- Say explicitly in abstract that define also events of auth module? (not sure, it might be obvious, just a suggestion..)

proposals/lip-0041.md
ricott1: changed several error messages

proposals/lip-0041.md
ricott1: changed several error messages

proposals/lip-0041.md
```diff
@@ -97,23 +103,216 @@ In this section, we describe the properties of an auth account and specify their
...
+    # Union of mandatory and optional keys must contain between 1 and 64 keys (included).
+    if len(all_keys) < 1 or len(all_keys) > 64:
+        raise Exception('Invalid length of mandatory and optional keys arrays.')
+    # Number of signatures must be an integer between 1 and the total number of public keys.
+    if trs.params.numberOfSignatures < 1 or trs.params.numberOfSignatures > len(all_keys):
+        raise Exception('Invalid number of signatures.')
+    # Number of signatures must not be smaller than the number of mandatory keys.
+    if trs.params.numberOfSignatures < len(trs.params.mandatoryKeys):
+        raise Exception('Invalid number of signatures.')
+
```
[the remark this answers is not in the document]
ricott1: added

proposals/lip-0041.md
ricott1: moved to execution step (and fixed a typo)

proposals/lip-0041.md
ricott1: It was a leftover, removed.

proposals/lip-0041.md
ricott1: agree (I think we solved this on slack)

proposals/lip-0041.md
ricott1: Good catch, changed.

proposals/lip-0041.md
ricott1: added

proposals/lip-0041.md
ricott1: not so sure anymore, maybe to have an easier nonce check in the transaction verification? although that is called also in the transaction pool, so we can not really skip the check on the account existence anyway

proposals/lip-0041.md
```diff
@@ -250,16 +452,15 @@ The Auth module does not execute any logic during the genesis state finalization
...
-* Check that the transaction nonce is equal to the account nonce by returning `isValidNonce(trs)`.
+As part of the verification of a transaction `trs`, check that the transaction nonce is equal to the account nonce or to 0 if the account does not exist. If the checks fail the transaction is invalid and the whole block is discarded.
 
 ```python
-isValidNonce(trs):
-  let senderAddress be the address corresponding to trs.senderPublicKey
-  if there is no entry in the in the auth data substore with storeKey == senderAddress:
-    return trs.nonce == 0
-  return trs.nonce == authAccount(senderAddress).nonce
+def transactionVerification(trs: Transaction) -> None:
```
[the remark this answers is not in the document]
ricott1: I restructured the code and added the auxiliary functions

proposals/lip-0041.md
ricott1: I rewrote the specs from scratch

proposals/lip-0041.md
janhack: @shuse2 Do you now the reason for it?

Not sure if it makes a performance difference if we modify the same state entry twice instead of once (I assume it would be cached).

proposals/lip-0041.md
```diff
@@ -97,23 +111,300 @@ In this section, we describe the properties of an auth account and specify their
...
+            "mandatoryKeys": trs.params.mandatoryKeys,
+            "optionalKeys": trs.params.optionalKeys 
+        }
+    )
+    allKeys = trs.params.mandatoryKeys + trs.params.optionalKeys
+    for signature, key in zip(trs.params.signatures, allKeys):
+        if not verifyMessageSig(key, MESSAGE_TAG_MULTISIG_REG, networkIdentifier, message, signature):
+            emitPersistentEvent(
+                moduleID = MODULE_ID_AUTH,
+                typeID = TYPE_ID_MULTISIGNATURE_GROUP_REGISTERED,
```
janhack, line 353:
```suggestion
                typeID = TYPE_ID_INVALID_SIGNATURE_ERROR`,
```

proposals/lip-0041.md
```diff
@@ -97,23 +111,300 @@ In this section, we describe the properties of an auth account and specify their
...
+#### isMultisignatureAccount
+
+This function checks whether the account corresponding to the input address has registered a multisignature group.
+
+```python
+def isMultisignatureAccount(address: Address) -> bool:
+    if there is no entry in the in the auth data substore with storeKey == address:
+        return False
+
+    if len(authAccount(senderAddress).mandatoryKeys) == len(authAccount(senderAddress).mandatoryKeys) == 0:
```
janhack, line 403: I don't understand this condition with two `==`. Shouldn't it be sufficient to check 
```
authAccount(senderAddress).numberOfSignatures == 0
```
Of course, we can also check that all 3 properties in the Auth store have the default values.

proposals/lip-0041.md
```diff
@@ -165,7 +456,18 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
```
janhack, line 456: This comment is rather for the function `verifySignatures` above, which still refers to LIP 17. Shouldn't it refer to the function defined here instead?

proposals/lip-0041.md
```diff
@@ -250,43 +555,101 @@ The Auth module does not execute any logic during the genesis state finalization
 
 #### Transaction Verification
 
-As part of the verification of a transaction `trs`, the following checks are applied. If the checks fail the transaction is invalid and the whole block is discarded.
+As part of the verification of a transaction `trs`, we do the following:
 
-* Check that the transaction nonce is equal to the account nonce by returning `isValidNonce(trs)`.
+* Check that the transaction nonce is equal to the account nonce or to 0 if the account does not exist. If the checks fail the transaction is invalid and the whole block is discarded.
+Notice that a transaction in the transaction pool with a nonce greater than the account nonce is considered "pending" rather than invalid, since it could become valid in the future.
+* Check the signatures property of the transaction [as explained in LIP 0017](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0017.md#signatures-replaces-signature) using `authAccount(address)`, where `address` is the address corresponding to `trs.senderPublicKey`.
```
janhack, line 562: I think this bullet point is outdated and should be removed.

proposals/lip-0041.md
```diff
@@ -250,43 +555,101 @@ The Auth module does not execute any logic during the genesis state finalization
 
 #### Transaction Verification
 
-As part of the verification of a transaction `trs`, the following checks are applied. If the checks fail the transaction is invalid and the whole block is discarded.
+As part of the verification of a transaction `trs`, we do the following:
 
-* Check that the transaction nonce is equal to the account nonce by returning `isValidNonce(trs)`.
+* Check that the transaction nonce is equal to the account nonce or to 0 if the account does not exist. If the checks fail the transaction is invalid and the whole block is discarded.
```
janhack, line 560: I think we could also remove this point as this is already specified in the code below. Maybe we want to still keep the notice about the transaction pool below.

proposals/lip-0041.md
```diff
@@ -250,43 +555,101 @@ The Auth module does not execute any logic during the genesis state finalization
...
+        schema = transactionSchema,
+        object = unisgnedTrs
+    )
+
+    # Check if the account is a multisignature account
+    senderAddress = 20-byte address derived from trs.senderPublicKey
+    
+    # Multisignature account case
+    if isMultisignatureAccount(senderAddress):
+        allKeys = trs.params.mandatoryKeys + trs.params.optionalKeys
```
janhack, line 601: I think the keys cannot come from the transaction parameters here, but should come from the Auth store instead.

proposals/lip-0041.md
```diff
@@ -250,43 +555,101 @@ The Auth module does not execute any logic during the genesis state finalization
...
+    
+    if trs.nonce != expectedNonce:
+        raise Exception(f'Invalid nonce property. Expected {expectedNonce}, got {trs.nonce}.')
+```
 
-* Check the signatures property of the transaction [as explained in LIP 0017](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0017.md#signatures-replaces-signature) using `authAccount(address)`, where `address` is the address corresponding to `trs.senderPublicKey`.
+We define the following auxiliary function to verify the validity of a signature. These specifications supersede those given in [LIP 0017](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0017.md#signatures-replaces-signature). The function `verifyMessageSig` is defined in [LIP 0062](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0062.md#specification) and `networkIdentifier` is the network identifier of the chain.
+
+```python
+def verifySignature(trs: Transaction) -> None:
```
janhack, line 587: I think tabs and spaces are mixed here. At least on GitHub the indentations are off sometimes.

proposals/lip-0041.md
```diff
@@ -250,43 +555,101 @@ The Auth module does not execute any logic during the genesis state finalization
...
+    )
+
+    # Check if the account is a multisignature account
+    senderAddress = 20-byte address derived from trs.senderPublicKey
+    
+    # Multisignature account case
+    if isMultisignatureAccount(senderAddress):
+        allKeys = trs.params.mandatoryKeys + trs.params.optionalKeys
+        if len(trs.signatures) != len(allKeys):
+		    raise Exception(f'Transactions from this multisignature account should have exactly {len(allKeys)} signatures. Found {len(trs.signatures)} signatures.')
```
janhack, line 603: I think tabs and spaces are mixed here and in other places in this function. At least on GitHub the indentations are off sometimes.

janhack (changes requested):
Overall, looks good and is better to also define `verifySignatures` in this LIP. There are still a few minor issues that I commented on.

proposals/lip-0041.md
ricott1: True, maybe I was overthinking this

proposals/lip-0041.md
shuse2: performance-wise, there is very little difference because it will be cached.
I think we could just increment here as well if there is no specific reason

proposals/lip-0041.md
```diff
@@ -250,43 +562,101 @@ The Auth module does not execute any logic during the genesis state finalization
...
+        if nonEmptySignatures != authAccount(senderAddress).numberOfSignatures:
+		    raise Exception(f'Transaction signatures does not match required number of signatures: {authAccount(senderAddress).numberOfSignatures}')
+        
+        numberOfMandatoryKeys = len(authAccount(senderAddress).mandatoryKeys)
+        for idx in range(numberOfMandatoryKeys):
+            key = authAccount(senderAddress).mandatoryKeys[idx]
+            if not verifyMessageSig(key, MESSAGE_TAG_TRANSACTION, networkIdentifier, unsignedTrsBytes, trs.signatures[idx]):
+                raise Exception(f'Invalid signature for public key {key.hex()}.')
+        
+        for idx in range(numberOfMandatoryKeys, len(allKeys)):
```
shuse2, line 622: Not related to this LIP, but python `range` operator is quite confusing 😓

shuse2 (approved):
Other than @janhack 's comment, LGTM

gkoumout (changes requested):
Added some comments. On top of them, agree to Jan's comments on the transaction verification part (bullet points, python idents, recovering keys from Auth store instead of transaction parameters..)

proposals/lip-0041.md
ricott1: Good catch. @shuse2 please notice that we changed the name of these two endpoints (maybe not terribly important)

proposals/lip-0041.md
ricott1: removed

proposals/lip-0041.md
ricott1: removed

proposals/lip-0041.md
```diff
@@ -165,7 +463,18 @@ This function verifies the nonce of a given transaction `trs` and returns a bool
 
 ##### Returns
 
-This function returns `true` if the transaction object contains a valid nonce, `false` otherwise. This is done using the `isValidNonce` function defined [below](#transaction-verification).
+This function returns `true` if the transaction object contains a valid nonce, `false` otherwise. This is done using the `verifyNonce(trs)` function defined [below](#transaction-verification).
+
+##### Execution
+
+```python
+def isValidNonce(trs: Transaction) -> bool:
```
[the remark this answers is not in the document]
ricott1: yes, fixed

proposals/lip-0041.md
```diff
@@ -94,26 +115,303 @@ In this section, we describe the properties of an auth account and specify their
...
+
+The Auth module has the following internal function.
+
+#### isMultisignatureAccount
+
+This function checks whether the account corresponding to the input address has registered a multisignature group.
+
+```python
+def isMultisignatureAccount(address: Address) -> bool:
+    if there is no entry in the in the auth data substore with storeKey == address:
```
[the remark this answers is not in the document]
ricott1: fixed

proposals/lip-0041.md
```diff
@@ -250,43 +562,101 @@ The Auth module does not execute any logic during the genesis state finalization
...
+    senderAddress = Address derived from trs.senderPublicKey
+    
+    # Multisignature account case
+    if isMultisignatureAccount(senderAddress):
+        allKeys = trs.params.mandatoryKeys + trs.params.optionalKeys
+        if len(trs.signatures) != len(allKeys):
+		    raise Exception(f'Transactions from this multisignature account should have exactly {len(allKeys)} signatures. Found {len(trs.signatures)} signatures.')
+        
+        nonEmptySignatures = [sig for sig in trs.signatures if len(sig) > 0]
+        if nonEmptySignatures != authAccount(senderAddress).numberOfSignatures:
```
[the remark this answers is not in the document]
ricott1: nice catch

proposals/lip-0041.md
ricott1: fixed

proposals/lip-0041.md
```diff
@@ -21,13 +21,13 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
 
 ## Motivation
 
-In the Lisk protocol, some verification steps are common to all transactions. In particular, to validate a transaction it is necessary to validate the [signatures](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#keys) and the [nonce](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#nonce). These validation steps are handled by the authentication (auth) module.
+In the Lisk protocol, some verification steps are common to all transactions. In particular, to validate a transaction it is necessary to validate the [signatures](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#keys) and the [nonce](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#nonce). These validation steps are handled by the authentication (Auth) module. The Auth modules supersedes the keys and sequence modules.
```
janhack, line 24:
```suggestion
In the Lisk protocol, some verification steps are common to all transactions. In particular, to validate a transaction it is necessary to validate the [signatures](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#keys) and the [nonce](https://lisk.com/documentation/lisk-sdk/protocol/accounts.html#nonce). These validation steps are handled by the authentication (Auth) module. The Auth modules supersedes the Keys and Sequence modules.
```

proposals/lip-0041.md
```diff
@@ -42,8 +42,23 @@ We define the following constants:
 |`MODULE_ID_AUTH`    | bytes  | TBD         | ID of the Auth module.                  |
 |`STORE_PREFIX_AUTH` | bytes   | 0x0000      | Store prefix of the auth data substore. |
 |`COMMAND_ID_REGISTER_MULTISIGNATURE_GROUP` | bytes   | 0x0000      | Command ID of the register multisignature group command. |
-|`TYPE_ID_MULTISIGNATURE_GROUP_REGISTERED` | bytes   | 0x0000      | Type ID of the signature group registered event. |
-|`MULTISIGNATURE_MSG_TAG` | bytes   | "LSK_RMSG_" as ASCII-encoded literal      | Message tag for the signatures contained in the register multisignature group command. |
+|`TYPE_ID_MULTISIGNATURE_GROUP_REGISTERED` | bytes   | 0x0001      | Type ID of the signature group registered event. |
+|`TYPE_ID_INVALID_SIGNATURE_ERROR` | bytes   | 0x0002      | Type ID of the invalid signature event. |
+|`MESSAGE_TAG_MULTISIG_REG` | bytes   | "LSK_RMSG_" as ASCII-encoded literal      | Message tag for the signatures contained in the register multisignature group command. |
+|`MESSAGE_TAG_TRANSACTION` | bytes   | "LSK_TX_" as ASCII-encoded literal      | Message tag for transaction signatures. |
```
janhack, line 48: I would not add constants already defined somewhere else here.

janhack (changes requested):
I resolved all my comments and added 2 minor style comments. Otherwise, there is only the topic of merging before/after command execute logic.

proposals/lip-0041.md
ricott1: fixed

proposals/lip-0041.md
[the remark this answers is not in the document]
gkoumout: Does not seem fixed to me

proposals/lip-0037.md
```diff
@@ -54,10 +54,15 @@ tagMessage(tag, networkIdentifier, message):
 
 The tag is chosen depending on the message type. Moreover, different serialization methods for the same message type require different tags. The following table defines the tags currently needed.
 
-| **message type**                                                                                                                                   | **tag**   |
-|----------------------------------------------------------------------------------------------------------------------------------------------------|-----------|
-| transaction with serialization as specified in [LIP 0028](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0028.md#transactionschema-schema) | "LSK_TX_" |
-| block header with serialization as specified in [LIP 0029](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0029.md#blockheader-schema)      | "LSK_BH_" |
+| Name               | Type    | Value       | Message type                            |
```
janhack, line 57:
```suggestion
| Constant Name               | Type    | Tag          | Message type                            |
```

proposals/lip-0037.md
```diff
@@ -54,10 +54,15 @@ tagMessage(tag, networkIdentifier, message):
...
+| ------------------ |---------| ------------| ----------------------------------------|
+|`MESSAGE_TAG_TRANSACTION`  | bytes   | "LSK_TX_" as ASCII-encoded literal      | Message tag for signing transaction with serialization specified in [LIP "Define New Transaction Schema"](https://research.lisk.com/t/define-new-transaction-schema/348/15). |
+|`MESSAGE_TAG_BLOCK_HEADER` | bytes   | "LSK_BH_" as ASCII-encoded literal      | Message tag for signing block headers with serialization specified in [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#unsigned-block-header-json-schema). |
+|`MESSAGE_TAG_MULTISIG_REG` | bytes   | "LSK_RMSG_" as ASCII-encoded literal    | Message tag for the signatures contained in the [register multisignature group command](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md#RegisterMultisignatureGroup). |
```
janhack, line 61:
```suggestion
|`MESSAGE_TAG_MULTISIG_REG` | bytes   | "LSK_RMSG_" as ASCII-encoded literal    | Message signed by the signatures contained in the [register multisignature group command](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md#RegisterMultisignatureGroup). |
```

proposals/lip-0037.md
```diff
@@ -54,10 +54,15 @@ tagMessage(tag, networkIdentifier, message):
...
+| `MESSAGE_TAG_CHAIN_REG_MESSAGE` | bytes  | "LSK_CHAIN_REGISTRATION_" as ASCII-encoded literal | Message tag for the signatures contained in the [chain registration message](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0043.md). |
```
janhack, line 62:
```suggestion
| `MESSAGE_TAG_CHAIN_REG_MESSAGE` | bytes  | "LSK_CHAIN_REGISTRATION_" as ASCII-encoded literal | Message signed by signatures contained in the [chain registration message](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0043.md). |
```

proposals/lip-0037.md
```diff
@@ -54,10 +54,15 @@ tagMessage(tag, networkIdentifier, message):
...
+| `MESSAGE_TAG_POA`         | bytes   | "LSK_POA_" as ASCII-encoded literal | Message tag for the signatures contained in the [update authority command](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0047.md#authority-registration-command). |
+| `MESSAGE_TAG_CERTIFICATE `| bytes  | "LSK_CE_" as ASCII-encoded literal | Message tag for signing certficates with serialization specified in [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#signcertificate). |
```
janhack, line 64:
```suggestion
| `MESSAGE_TAG_CERTIFICATE `| bytes  | "LSK_CE_" as ASCII-encoded literal | Certficates with serialization specified in [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#schema). |
```

janhack:
I left a few suggestions to improve the table in LIP 37. There is also one old unaddressed comment still.

proposals/lip-0041.md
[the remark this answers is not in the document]
ricott1: sorry, I had it more than once and did not notice

gkoumout (approved):
Looks good to me! Great job!

proposals/lip-0041.md
ricott1: done

janhack (approved):

karmacoma (approved):

PR #152, opened 2022-06-10
Update lip 0063 according to Legacy module migration logic

sergeyshemyakov:


proposals/lip-0063.md
```diff
@@ -332,24 +342,29 @@ In the following, let `accountsState` be the key-value store of the accounts sta
...
         userObj.address = account.address
         userObj.balance = account.token.balance
         legacyObj.accounts.append(userObj)
+        legacyReserveAmount += account.token.balance
     data = serialization of legacyObj using genesisLegacyStoreSchema
     append {"moduleID": MODULE_ID_LEGACY, "data": data} to b.assets
+    # mint all the tokens for legacy accounts here
+    Token.mint(ADDRESS_LEGACY_RESERVE, TOKEN_ID_LSK_MAINCHAIN, legacyReserveAmount)
+    Token.lock(ADDRESS_LEGACY_RESERVE, MODULE_ID_LEGACY, TOKEN_ID_LSK_MAINCHAIN, legacyReserveAmount)
 ```
```
MaximeGagnebin, line 360: I think this logic is not really following the migration paradigm.
The migration should not mint + lock tokens. This amount should be  added to the `assets` entry for the token module under the correct address.

MaximeGagnebin (changes requested):
I think the logic for handling the amounts for legacy accounts should be revisited.

janhack (changes requested):
Ah, I missed the context here. We are actually just computing the genesis block and therefore cannot call any APIs or make any state changes. As Maxime noted, we would need to modify the asset of the Token module instead.

sameersubudhi (approved):
LGTM 👍

MaximeGagnebin (approved):
According to the current specs, I think the different asset arrays need to be sorted (each module defines the sorting to be respected)
I would approve the PR as I think the Legacy module change is now settled.

janhack (changes requested):
I think there is an issue with the current specifications if `ADDRESS_LEGACY_RESERVE` has any balance.

janhack (approved):
Looks all good now!

karmacoma (approved):

PR #155, opened 2022-06-21
Add LIP:  "Define new transaction schema"

gkoumout:


proposals/lip-define_new_transaction_schema.md
```diff
@@ -98,6 +100,7 @@ transactionSchema = {
         },
         "senderPublicKey": {
             "dataType": "bytes",
+            "length": PUBLIC_KEY_LENGTH_BYTES,
```
ricott1, line 103: For this constant I would use `ED25519_PUBLIC_KEY_LENGTH` to be in line with other LIPs.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,189 @@
...
+The `params` property must follow the schema corresponding to the `moduleID`, `commandID` pair defined in the corresponding module.
+
+All transaction procedures - serialization, deserialization, signature calculation, signature validation and transaction ID - follow the same specifications already defined in [LIP 0028](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0028.md#serialization). The resulting serialization or signatures are however different when the proposed transaction schema is used.
+
+
+### Constants
+
+
+| Global Constants                |         |                                                                                                            |
+|:-------------------------------:|:-------:|------------------------------------------------------------------------------------------------------------|
```
ricott1, line 54: I think the colons (:) are centering  the entries in the table, I would remove them

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,189 @@
+```
+LIP: <lip number>
+Title: Define new transaction schema
+Author: Maxime Gagnebin <maxime.gagnebin@lightcurve.io>
+? Discussions-To: <Link to discussion in Lisk Research>
```
ricott1, line 5:
```diff
Discussions-To: https://research.lisk.com/t/define-new-transaction-schema/348/15
```

ricott1 (changes requested):
Some minor details to be fixed

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,190 @@
...
+
+Defining identifiers as type `bytes` also requires fixing the length of the identifier. This was not possible when using identifiers of type `uint32`, as the full 4 bytes of the maximal range always had to be assumed when using the identifier in the state tree.  
+
+
+### New Property Names
+
+All properties in the proposed transaction schema are equivalent to the ones defined in LIP 0028. The only changes are the renaming of `assetID` to `commandID` and of `asset` to `params`. As was described in LIP "Update Lisk SDK modular blockchain architecture".
```
janhack, line 38: - It would be nice to link to LIP 0028.
- It would be nice to link to the research forum for the LIP "Update Lisk SDK modular blockchain architecture". Note to make links easily updateable it is recommended to use shortcuts defined at the end of the markdown file such as:

```
[lip-0028]: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0028.md
```

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+## Specification
+
+The transaction schema defined in [LIP 0028](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0028.md) is superseded by the one defined below. 
+
+The `params` property must follow the schema corresponding to the `moduleID`, `commandID` pair defined in the corresponding module.
+
+All transaction procedures - serialization, deserialization, signature calculation, signature validation and transaction ID - follow the same specifications already defined in [LIP 0028](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0028.md#serialization). The resulting serialization or signatures are however different when the proposed transaction schema is used.
```
janhack, line 47: I think this is actually not true for the following reasons:
- We do not allow optional properties in schemas any more due to LIP 64. Therefore, we need a schema both for signed and unsigned transactions.
- We now use message tags for signing transactions as defined in LIP 37.

I would suggest to define these functions here using Python-style pseudocode with type-hinting. This way we also have a clean interface to use in other places.

janhack: As far as I see these would be the functions:
```
encode(assetSchema: LiskJSONSchema, trs: Object) -> bytes:

decode(assetSchema: LiskJSONSchema, transactionData: bytes) -> Object:

signTransaction(assetSchema: LiskJSONSchema, trs: Object) -> SignatureEd25519:

verifyTransactionSignature(assetSchema: LiskJSONSchema, trs: Object) -> bool:
```

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,190 @@
...
+            "fieldNumber": 7
+        }
+    }
+}
+```
+
+
+### Validation
+
+For a transaction `trs` to be valid, it must satisfy the following:
```
janhack, line 125: Here the context of where this would be checked is missing.

janhack (changes requested):
Overall, looks all good. I just left a few comments.

I did not verify the appendix as I assume that has happened already by other reviewers before the publication.

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Usually the "name", "value", "description" raw is centered, but couldn't make it here due to the "global constants" raw. Not a big deal, could change it if you insist :-)

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,190 @@
...
+### Constants
+
+
+| Global Constants                |         |                                                                                                            |
+|:-------------------------------:|:-------:|------------------------------------------------------------------------------------------------------------|
+| **Name**                        |**Value**|**Description**                                                                                             |
+| `MODULE_ID_LENGTH_BYTES `       | 4       | The length of module IDs.                                                                                  |
+| `COMMAND_ID_LENGTH_BYTES`       | 2       | The length of command IDs.                                                                                 |
+| `PUBLIC_KEY_LENGTH_BYTES `      | 32      | The length of public keys.                                                                                 |
+| `SIGNATURE_LENGTH_BYTES`        | 64      | The length of signatures.                                                                                  |
```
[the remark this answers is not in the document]
gkoumout: fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
ricott1: Do we actually have to decide ourselves where this is checked? I find this to be an implementation detail. Personally, I think this check should not be part of the application layer but rather of the block header processing (consensus). For instance, we could perform the check as part of the `transactionRoot` calculation (since we have to serialize transactions there anyway).

ishantiw (approved):

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+
+Therefore, for the cases where removing the old identifiers does not introduce any significant challenge in the ecosystem, it is plausible to proceed to this switch and use names as identifiers. This is the case for modules, commands and events.  
+
+
+### New Property Names
+
+
+All properties in the proposed transaction schema are equivalent to the ones defined in [LIP 0028][lip-0028]. The only changes are the replace of identifiers by the corresponding names and the update of terminology according to the [LIP "Update Lisk SDK modular blockchain architecture"][lip-update-lisk-sdk-modular-architecture] (renaming a module asset to a command and a transaction `asset` property to `params`). Overall, `moduleID` is replaced by `module`, `assetID` is replaced by `command` and `asset` is renamed to `params`.
```
janhack, line 39:
```suggestion
All properties in the proposed transaction schema are equivalent to the ones defined in [LIP 0028][lip-0028]. The only changes are the replacement of identifiers by the corresponding names and the update of terminology according to the [LIP "Update Lisk SDK modular blockchain architecture"][lip-update-lisk-sdk-modular-architecture] (renaming a module asset to a command and a transaction `asset` property to `params`). Overall, `moduleID` is replaced by `module`, `assetID` is replaced by `command` and `asset` is renamed to `params`.
```

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+
+The validity of each transaction is verified in the [static validation stage][lip-0055#block-processing] of the block processing. 
+
+
+### Serialization
+
+The serialization of an object of type `Transaction` is described in the following pseudocode.
+
+```python
+def encode(transactionSchema: LiskJSONSchema, trs: Transaction) -> bytes:
```
janhack, line 148: Shouldn't `paramsSchema` also be an input parameter?

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+    trs.params = encode(paramsSchema,trs.params)
+    return encode(transactionSchema,trs)
+```
+
+### Deserialization
+
+Consider a binary message `trsMsg`, corresponding to a serialized transaction. The deserialization procedure is as follows:
+
+```python
+def decode(transactionSchema: LiskJSONSchema, trsMsg: bytes) -> Transaction:
```
janhack, line 158: Shouldn't `paramsSchema` also be an input parameter?

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+    trsData = decode(transactionSchema,trsMsg)
+    trsData.params = decode(paramsSchema,trsData.params)
+    return trsData
+```
+
+### Transaction signature calculation
+Consider a data structure `unsignedTrsData` representing a valid `Transaction` object in which the signatures array is initialized to the default value (an empty array). The following function calculates a signature of the object on a certain chain with secret key `sk`. 
+
+```python
+def signTransaction(sk: PrivateKeyEd25519, unsignedTrsData: Transaction) -> SignatureEd25519:
```
janhack, line 168: Maybe we could call the function `computeTransactionSignature`. The current function name suggests that a signed transaction is returned, similar to the function (signTransaction)[https://github.com/LiskHQ/lisk-sdk/blob/ef942e74ea55f44db1af3a720093969d6a42194b/elements/lisk-transactions/src/sign.ts#L77] in the Lisk SDK.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+    "amount": 123986407700n,
+    "recipientID": '2ca4b4e9924547c48c04300b320be84e8cd81e4a',
+    "data": 'Odi et amo. Quare id faciam, fortasse requiris.'
+  },
+  "signatures": [
+    '9953f164f9664e05526c1e3a10c4631715cdcb9fd4f376bf7db5334ded3bbc8470bce023d67c7aca16cf3389ea01f3e3c011820c317f1f5a63f98bb6d6b34b07',
+    'a95fc611f7207ddaaaf7929f8f19b7c1cb2473ead20e9be99b8c0abc148b4ea35713ed296acbd6612f124698e96d57e6fde0eddbb998b86203d04ff3c3976700'
+  ]
+}
+```
```
janhack, line 209: The example above needs to be updated (moduleID and commandID).

proposals/lip-define_new_transaction_schema.md
janhack: Can be resolved from my side as it is now clearly defined.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+All properties in the proposed transaction schema are equivalent to the ones defined in [LIP 0028][lip-0028]. The only changes are the replace of identifiers by the corresponding names and the update of terminology according to the [LIP "Update Lisk SDK modular blockchain architecture"][lip-update-lisk-sdk-modular-architecture] (renaming a module asset to a command and a transaction `asset` property to `params`). Overall, `moduleID` is replaced by `module`, `assetID` is replaced by `command` and `asset` is renamed to `params`.
+
+
+## Specification
+
+The transaction schema defined in [LIP 0028][lip-0028] is superseded by the one defined [below](#json-schema). 
+
+The `params` property must follow the schema corresponding to the (`module`, `command`) pair defined in the corresponding module; we call this schema `paramsSchema`.
+
+As for the other transaction procedures:
```
janhack, line 48: I would suggest to create a section "Changes compared to LIP 28" in the rationale and write in the specfications that the serialization, deserialization, ID computation and signature computation are superseded by the specifications here. I think this would be cleaner and indeed serialization/deserialization is different due to different schema, signature calculation is different due to the message tage and ID computation is different as the serialized transaction object will be different.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,241 @@
...
+
+- Serialization and deserialization follow the same specifications already defined in [LIP 0028][lip-0028]; for completeness we include the pseudocode [below](#serialization). The resulting serialization is however different when the proposed transaction schema is used, due to the change of types for identifiers for module and command. Moreover, the transaction ID is calculated in the same way as described in [LIP 0028][lip-0028] (the SHA-256 hash of the serialized transaction object).
+- Signature calculation follows the same specifications as in [LIP 0028][lip-0028], updated to incorporate message tags introduced in [LIP 0037][lip-0037]. For completeness we provide the pseudocode [below](#transaction-signature-calculation).
+- Signature validation is done using the `verifySignatures` function defined in [LIP 0041](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md#transaction-verification).
```
janhack, line 52: I would suggest to add a stateless function 
```
validateTransactionSignature(pk: PublicKeyEd25519, sig: SignatureEd25519, networkIdentifier, unsignedSerializedTransaction: bytes) -> bool:
```
This way we define the transaction signing and signature validation fully in one place (i.e., concrete call to the crypto library and which message tag is used) and everywhere else could utilize the functions here.

proposals/lip-define_new_transaction_schema.md
janhack: @ricott1 Are you assuming there is a property `trs.schema` and `trs.assetSchema` or where should the schema come from? Apart from the interface to the state store and constants, I would try to avoid introducing global variables which are not defined anywhere.

janhack: As mentioned [here](https://github.com/LiskHQ/lips/pull/155#discussion_r937632956) I would still suggest to introduce a stateless signature validation function so that this LIP defines serialization/deserialization, signing/signature validation for transactions fully. Then this functionality is encapsulated in one place and could be updated (e.g., changing the message tag, call to the underlying crypto library) without changing many LIPs in different places.

proposals/lip-define_new_transaction_schema.md
gkoumout: I changed the table and there is no centering anymore ;-)

proposals/lip-define_new_transaction_schema.md
gkoumout: I am aware of that. I will update if I find time, otherwise remove the appendix for now and merge it later.

proposals/lip-define_new_transaction_schema.md
ricott1: I'm not sure if I agree that we need this stateless function. Concretely, the full transaction signature validation is delegated to the Auth module, and it is not stateless. Of course we could add this stateless version here, but why should it be attached in particular to the concept of transaction? the stateless signature verification is related more to validating a signature of a raw bytes object (and it is already defined in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0062.md#specification)

proposals/lip-define_new_transaction_schema.md
ricott1: cbr

proposals/lip-define_new_transaction_schema.md
ricott1: To me it could work in a way similar to how we call commands or methods (for instance in the interop LIPs): we say, `assetSchema = JSON schema corresponding to (module, command) tuple`.

proposals/lip-define_new_transaction_schema.md
ricott1: Here I would rather obtain the paramsSchema from the transaction proeprties. I agree that `transactionSchema` is not needed (see this comment https://github.com/LiskHQ/lips/pull/155#discussion_r908507070)

ricott1 (changes requested):
I think we can still add the max length and min length properties to the schema for the `module` and `command` properties. 

Last point, I think @gkoumout should be added as author to this LIP.

proposals/lip-define_new_transaction_schema.md
gkoumout: I would agree with @ricott1 for the encode function, but for the decode you can not find the paramsSchema in the same way (since you don't have access to module,command). So I include `paramsSchema` as an argument

proposals/lip-define_new_transaction_schema.md
gkoumout: Name changed. `networkIdentifier` added as input parameter. 
I kept the `Transaction` object as input argument and recover the paramsSchema from the (module,command) tuple.

proposals/lip-define_new_transaction_schema.md
gkoumout: I have added a bulletpoint in the specification that signature validation is done using the `verifySignatures` function defined in LIP0041. I think this is enough. Let me know if you think otherwise. 

I could move it from a bulletpoint in the specification to a separate section after "Transaction Signature Calculation"

proposals/lip-define_new_transaction_schema.md
gkoumout: Done

proposals/lip-define_new_transaction_schema.md
ricott1: also for decoding, you first decode the root schema, and then get the paramsSchema from the module and command

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+
+
+#### Validation
+
+For a transaction `trs` to be valid, it must satisfy the following:
+
+* `trs` must follow the `transactionSchema`. 
+* `trs.params` is of length less than or equal to `MAX_PARAMS_SIZE` .
+
+The validity of each transaction is verified in the [static validation stage][lip-0055#block-processing] of the block processing.
```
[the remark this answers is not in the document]
ricott1: I would put it in this more general language, because the full validity of a transaction also depends on the application layer

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+
+The validity of each transaction is verified in the [static validation stage][lip-0055#block-processing] of the block processing. 
+
+
+### Serialization
+
+The serialization of an object of type `Transaction` is described in the following pseudocode.
+
+```python
+def encode(paramsSchema: LiskJSONSchema, trs: Transaction) -> bytes:
```
ricott1, line 157: I think we need a different function name, otherwise it gets confusing with the generic `encode` function (which you call here)

ricott1: Concretely, I would use `encodeTransaction`

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+    trs.params = encode(paramsSchema,trs.params)
+    return encode(transactionSchema,trs)
+```
+
+### Deserialization
+
+Consider a binary message `trsMsg`, corresponding to a serialized transaction. The deserialization procedure is as follows:
+
+```python
+def decode(paramsSchema: LiskJSONSchema, trsMsg: bytes) -> Transaction:
```
ricott1, line 167: same as above, need a different name

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+    trsData.params = decode(paramsSchema,trsData.params)
+    return trsData
+```
+
+### Transaction Signature Calculation
+
+Consider a data structure `unsignedTrsData` representing a valid `Transaction` object in which the signatures array is initialized to the default value (an empty array). The following function calculates a signature of the object on a certain chain with secret key `sk`. 
+
+```python
+def computeTransactionSignature(sk: PrivateKeyEd25519, unsignedTrsData: Transaction, networkIdentifier: bytes) -> SignatureEd25519:
```
ricott1, line 178:
```suggestion
def computeTransactionSignature(sk: PrivateKeyEd25519, unsignedTrs: Transaction, networkIdentifier: bytes) -> SignatureEd25519:
```

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+    paramsSchema =  JSON schema corresponding to (module, command) pair.
+    serializedTrs = encode(paramsSchema, unsignedTrsData)
```
ricott1, line 180: update name (indeed here is very confusing to use `encode`)

proposals/lip-define_new_transaction_schema.md
ricott1, line 157: As I said before, I would prefer not to pass the `paramsSchema` at all (like what you do in the signature calculation. At the end either option is fine, but I would be consistent and either always pass the params schema or always obtain it from the transaction data

proposals/lip-define_new_transaction_schema.md
ricott1, line 178: Alternatively, we could just pass the transaction object and set the signatures to empty explicitly before signing it

ricott1 (changes requested):
I still have some comments. I think we just need to decide for a specific approach and use it consistently, either pass the `paramsSchema` everywhere or never pass it and get it from the transaction object.

ishantiw (approved):
Apart from some points raised by @ricott1 , from my end it looks good so approving

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,258 @@
...
+        "fee": {
+            "dataType": "uint64",
+            "fieldNumber": 4
+        },
+        "senderPublicKey": {
+            "dataType": "bytes",
+            "length": ED25519_PUBLIC_KEY_LENGTH,
+            "fieldNumber": 5
+        },
+        "params": {
```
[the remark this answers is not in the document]
gkoumout: Is it supported for bytes? I am not a json schemas expert

proposals/lip-define_new_transaction_schema.md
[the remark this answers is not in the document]
gkoumout: Aggreed, fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Fixed

proposals/lip-define_new_transaction_schema.md
gkoumout: Agreed. Removed it from everywhere to be consistent. 

But now we should make sure that whenever calling this function in other LIPs we are consistent.

proposals/lip-define_new_transaction_schema.md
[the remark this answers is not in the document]
ricott1: I am not sure. It's fine also not to add it

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,263 @@
...
+            "minLength": MIN_MODULE_NAME_LENGTH, 
+            "maxLength": MAX_MODULE_NAME_LENGTH,
+            "pattern": "^[a-z0-9!@&_,.]*$",
+            "fieldNumber": 1
+        },
+        "command": {
+            "dataType": "string",
+            "minLength": MIN_COMMAND_NAME_LENGTH, 
+            "maxLength": MAX_COMMAND_NAME_LENGTH,
+            "pattern": "^[a-z0-9!@&_,.]*$",
```
ricott1, line 113: Maybe this is a bit much for the schema, I would add it just in the specs. Either way, this is the last open point from my side.

ricott1 (approved):
Just the open point about the character set from my side

proposals/lip-define_new_transaction_schema.md
janhack: Can be resolved.

proposals/lip-define_new_transaction_schema.md
janhack: Can be resolved.

proposals/lip-define_new_transaction_schema.md
janhack: Looks good.

proposals/lip-define_new_transaction_schema.md
janhack: Looks good now.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,263 @@
...
+def computeTransactionSignature(sk: PrivateKeyEd25519, unsignedTrs: Transaction, networkIdentifier: bytes) -> SignatureEd25519:
+    serializedTrs = encodeTransaction(unsignedTrs)
+    return signMessage(sk, MESSAGE_TAG_TRANSACTION, networkIdentifier, serializedTrs)
+```
+
+Here `networkIdentifier` is the [network identifier](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#network-identifiers) for the chain and the function `signMessage` is defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#signing-and-verifying-with-ed25519).
+
+### Transaction Signature Validation
+
+Signature validation is done using the `verifySignatures` function defined in [LIP 0041](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md#transaction-verification).
```
janhack, line 193:
```suggestion
Signature validation is done using the `verifySignatures` function defined in [LIP 0041](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md#transaction-verification) with the input parameter `tag` equal to `MESSAGE_TAG_TRANSACTION`. 
```

proposals/lip-define_new_transaction_schema.md
janhack: Looks good now, can be resolved.

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,263 @@
...
+
+This LIP results in a hard fork as nodes following the proposed protocol will reject transactions following the previous schema, and nodes following the previous protocol will reject transactions following the proposed schema.
+
+
+## Reference Implementation
+
+TBD
+
+
+## Appendix
```
janhack, line 207: I would suggest to remove the appendix for now if it is outdated (and maybe copy it to a separate document).

proposals/lip-define_new_transaction_schema.md
janhack: Okay, that is also a good solution. Can be resolved from my side.

proposals/lip-define_new_transaction_schema.md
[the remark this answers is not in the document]
janhack: We could asks the devs if that is possible. Otherwise, the current version is also fine.

proposals/lip-define_new_transaction_schema.md
janhack: I think I prefer that approach as it is more robust.

janhack (changes requested):
I mostly resolved my comments.

I added one small suggestion and another comment regarding the appendix. I think we shouldn't have an incorrect appendix, but rather remove it for now (or comment it out).

Further, we still need to finalize the regex for module names and command names.

proposals/lip-define_new_transaction_schema.md
ricott1: I think this function should not be called by other LIPs (maybe yes, but I currently don't really see the need). But anyway, we will make sure :)

proposals/lip-define_new_transaction_schema.md
ricott1: agree. I could also prepare it, otherwise

proposals/lip-define_new_transaction_schema.md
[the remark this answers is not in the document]
ricott1: I think we can omit it, it is quite clear in the specs.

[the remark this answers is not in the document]
janhack: Sounds good.

proposals/lip-define_new_transaction_schema.md
janhack: I updated the function as suggested.

proposals/lip-define_new_transaction_schema.md
janhack: I moved it to the validation section and updated the regular expression as concluded.

proposals/lip-define_new_transaction_schema.md
janhack: I commented out the appendix for now.

janhack (approved):

ricott1 (approved):

ishantiw (approved):

proposals/lip-define_new_transaction_schema.md
```diff
@@ -177,13 +177,13 @@ def decodeTransaction(trsMsg: bytes) -> Transaction:
 
 ### Transaction Signature Calculation
 
-The following function calculates the signature with secret key `sk` of a transaction `trs` for a chain identified by the given [network identifier](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#network-identifiers).
+The following function calculates the signature with secret key `sk` of a transaction `trs` for a chain identified by the given [chainID](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md).
```
ricott1, line 180: I would still link to the section https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#chain-identifiers

ricott1 (changes requested):
One small comment, also please update the appendx

proposals/lip-define_new_transaction_schema.md
gkoumout: Indeed, I did not do it because the updated LIP37 was not merged yet..

Fixed

ricott1 (approved):

proposals/lip-define_new_transaction_schema.md
```diff
@@ -0,0 +1,263 @@
...
+
+
+## Copyright
+
+This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).
+
+
+## Motivation
+
+The Lisk protocol handles identifiers for transactions, modules, commands, and many more. In many of those cases such as modules and commands there is also a name property, which is of type string and is set to some intuitive value (e.g., "token" module, "token transfer" command). In the current protocol, such objects are referenced using their identifiers and their names have an auxiliary role. This is not so convenient for users and developers, since they have to memorize the (non-intuitive) identifier values. Merging those two properties and identifying modules and commands using their name provides a much better user/developer experience.
```
ishantiw, line 25: Just a small comment that command name example is `tokenTransfer` camelCase

ishantiw (approved):
Looks good, just a small comment

proposals/lip-define_new_transaction_schema.md
gkoumout: Good point, thanks! Fixed.

karmacoma (approved):

PR #158, opened 2022-07-13
Update events and functions in LIP0041 and LIP0044

ricott1:


proposals/lip-0044.md
```diff
@@ -389,12 +389,12 @@ def registerValidatorWithoutBLSKey(validatorAddress: Address, generatorKey: Publ
 
 #### getValidatorAccount
 
-This function is used to retrieve information about the validator account corresponding to the input `address: Address`. It returns `validatorAccount(address)`. If there is no entry corresponding to `address`, it returns an empty object.
+This function is used to retrieve information about the validator account corresponding to the input `address: Address`. It returns `validatorAccount(address)`. If there is no entry corresponding to `address`, it throws an error.
 
 ```python
 def getValidatorAccount(address: Address) -> ValidatorAccount:
```
sergeyshemyakov, line 395: I checked for the use of this function in other LIPs. I have found only one place where the logic should also change slightly, see here: https://github.com/LiskHQ/lips/blob/e38f4b1307a9a974aadb157e56d2782e9b4d3732/proposals/lip-0050.md?plain=1#L309
I think, we can just remove the check on this line. But possibly we can add try/catch to throw a more specific Exception (saying `'Public key does not correspond to a registered validator.'` instead of `'No validator account found for the input address.'`). I am fine with either option.

sergeyshemyakov (changes requested):

sergeyshemyakov (approved):
All looks good.

proposals/lip-0041.md
```diff
@@ -214,8 +215,8 @@ This command allows users to register a multisignature group for the sender acco
```
gkoumout, line 215: Don't forget to replace networkID by chainID.

proposals/lip-0041.md
```diff
@@ -350,8 +351,8 @@ def execute(trs: Transaction) -> None:
     for signature, key in zip(trs.params.signatures, allKeys):
         if not verifyMessageSig(key, MESSAGE_TAG_MULTISIG_REG, networkIdentifier, message, signature):
             emitPersistentEvent(
-                moduleID = MODULE_ID_AUTH,
-                typeID = TYPE_ID_INVALID_SIGNATURE_ERROR,
+                module = MODULE_NAME_AUTH,
```
gkoumout, line 354: Comment for the first line of this function: Reminder that Transaction type is not defined here. Not sure we should add it everywhere, but I am adding it in my LIPs so I guess it is good to do it for consistency.

gkoumout (changes requested):
Looks good to me! All changes to strings seem ok to me.

I just left a couple of comments for replacing networkID by chainID and defining Transaction type for type hinting

proposals/lip-0041.md
ricott1: thanks, fixed

proposals/lip-0041.md
ricott1: I think we decided not to define it and assume it is known.

gkoumout (approved):

karmacoma (approved):

PR #159, opened 2022-07-19
Update LIP 0047 PoA module: added type hinting, events

sergeyshemyakov:
This PR aligns the PoA module with the changes in events, LIP styling, etc.

To reviewers: please have a close look at the event. I don't emit many events because I rely on the events emitted while executing helper functions from other modules like `setValidatorGeneratorKey` or `burn`. For the authority update event I was not sure whether any sensible info should be put into topics or data.


proposals/lip-0047.md
```diff
@@ -100,26 +118,29 @@ validatorObjectSchema = {
...
    },
    "required": ["name"]
 }
 ```
 
-##### Properties and Default Values
+##### Properties
 
-`name` is a string representing the validator name, with a minimum length of 1 character and a maximum length of `MAX_LENGTH_NAME` characters. Its value is set in the genesis block or with an [authority registration command](#authority-registration-command)
+- `name` is a string representing the validator name, its value is set in the genesis block or with an [authority registration command](#authority-registration-command)
```
vjaiman, line 132: Broken links for "authority registration command" at multiple places.

proposals/lip-0047.md
```diff
@@ -29,7 +29,7 @@ A PoA blockchain is especially attractive for small projects or blockchain appli
```
mjerkov, line 29: Note: All line numbers refer to the new version of the document.

Overall, I have no major objections, comments mostly relate to readability and consistency of the document.

General comments:
- Throughout the document: check where 'validator' can be replaced with 'authority', to be more specific.
- Throughout the document: check where 'processed' can be replaced by 'executed', in the context of block or transaction (following the recommendations of LIP0069).
- Line 16: replace 'state transitions logic' with 'state transition logic'.
- Line 40: split into 2 or 3 paragraphs to improve readability, possibly starting the second with 'In particular, ...' and the third with 'With the success, ...'. Also, to avoid possible confusion, replace 'with their associated weight' with 'with their associated weights' and 'change their associated weight' with 'change their associated weights'.
- Lines 741 and 798: replace '`setGeneratorList`is' with '`setGeneratorList` is' (missing space).

Substores:
- Line 98: the figure should be corrected so that the constants for the store names are changed from 'STORE_PREFIX_...' to 'SUBSTORE_PREFIX_...'.
- Lines 111 and 143: use 'Notation:' at the beginning of the last bullet in 'Substore Prefix, Store Keys, and Store Values'; missing in Validator and Name susbstores, existant in Snapshot and Chain Properties substores
- Lines 172-173 and 222-223: replace 'store' with 'substore' in the second and third bullet points in 'Substore Prefix, Store Keys, and Store Values' of Snapshot Substore and Chain Properties Substore.

Schemas:
- Check if there can be a consistent rule for the placement of the "required" keyword: sometimes it appears before, sometimes after the "property" keyword. Also, if multiple properties are required, consider placing them each in its own line (Line 308 vs. Lines 443-447)
- Check if there can be a uniform naming convention for schemas. For example, on Line 280 the schema for 'Register Authority Command' is named 'registrationTransactionParamsSchema', while on Line 403 the schema for 'Update Authority Command' is named 'updateValidatorParams'. Suggestion: the first schema could be named 'registerValidatorParamsSchema', and the second one 'updateValidatorParamsSchema'. Or, alternatively, instead of 'Validator' one could use 'Authority', which would match the naming of the constant for the command name.
- Line 285: check if "minLength" is also needed (as in Lines 122 and 644).

Commands:
- Check if the linking and mentioning of commands can be done in a more consistent way. For example, for 'Register Authority Command' (Line 273), there are different conventions for how it is mentioned in the document: in Lines 109, 132, 136, 141, and 163 it is mentioned as 'authority registration command' (and linked), while in Line 602 it is mentioned as 'Authority Registration command' (and not linked to). Similarly, the 'Update Authority Command' (Line 396) is mentioned twice in Line 40 and once in Lines 248 and 254 as 'update authority command'. In contrast, in Line 608 it is mentioned as 'Authority Update command'.

Pseudocode:
- Consider whether the checks and executions of 'Genesis State Initialization' (Lines 602-634) can be written in pseudocode instead plain text. In the same text, check whether `storeKey` and `storeValue` can simply be called 'substore key' and 'substore value', since the first two terms have never been used before.
- Check whether the pseudocode in Lines 725-742 can be made consistent with the pseudocode elsewhere: for example, in Lines 725 and 739 possibly use 'length' instead of 'len', and in Line 733 use 'for validator in' instead of 'for every validator element in'.

Links:
- Line 711 - broken link: 'https://github.com/ethereum/guide/blob/main/poa.md' should be changed to 'https://github.com/ethereum/guide/blob/master/poa.md'

mjerkov (commented):
@sergeyshemyakov , it was easier for me to collect all the comments in one place, since they are mostly general in nature and often affect several lines of the document. If that is a problem, I will split them into several parts and spread them over the document.

mjerkov (approved):
Looks good!

proposals/lip-0047.md
```diff
@@ -50,22 +50,39 @@ As mentioned before, the sidechain developers using the Lisk SDK may specify the
...
+The LIP uses the following types.
+
+|Name            | Type              | Validation                           | Description                            |
+|----------------|-------------------|--------------------------------------|----------------------------------------|
+| Address        | bytes             | must have length `NUM_BYTES_ADDRESS` | Account address in Lisk ecosystem      |
+| RandomSeed     | bytes             | must have length 32                  | Random seed used in validator shuffling|
 
 ### Constants and Notation
 
 | **Name**                            | **Type** | **Value**                       |
```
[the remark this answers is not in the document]
vjaiman: okay.

vjaiman (approved):
Looks all good.

PR #162, opened 2022-08-04
Update LIP0041 dn LIP0044

ricott1:
Update LIP0041 and LIP0044 to change event names.


ricott1:
Merged into other PR https://github.com/LiskHQ/lips/pull/158

PR #163, opened 2022-08-04
Update LIP0040 and LIP0065

ricott1:


proposals/lip-0065.md
```diff
@@ -98,23 +100,31 @@ To summarize, we chose sparse Merkle trees because:
...
 | `EVENT_TOPIC_INIT_GENESIS_STATE`     | bytes   | `0x00`                                                | Default topic for events emitted during genesis state initialization.                |
 | `EVENT_TOPIC_FINALIZE_GENESIS_STATE` | bytes   | `0x01`                                                | Default topic for events emitted during genesis state finalization.                  |
 | `EVENT_TOPIC_BEFORE_TRANSACTIONS`    | bytes   | `0x02`                                                | Default topic for events emitted _before_ transactions execution.                    |
 | `EVENT_TOPIC_AFTER_TRANSACTIONS`     | bytes   | `0x03`                                                | Default topic for events emitted _after_ transactions execution.                     |
-| `EVENT_TYPE_ID_STANDARD`             | bytes   | `0x0000`                                              | Type ID of standard event emitted at the end of transaction execution.               |
+| `EVENT_NAME_STANDARD`                | string   | `executionResult`                                    | Name of standard event emitted at the end of transaction execution.               |
```
shuse2, line 112: maybe `transactionResult` since there are other execution steps?

proposals/lip-0065.md
```diff
@@ -123,14 +133,18 @@ Events are serialized and deserialized using the following JSON schema.
 ```java
 eventSchema = {
     "type": "object",
-    "required": ["moduleID", "typeID", "data", "topics", "index"],
+    "required": ["module", "name", "data", "topics", "index"],
```
shuse2, line 136: height is missing in required field?

proposals/lip-0040.md
```diff
@@ -52,6 +52,14 @@ Store keys are hashed in order to randomize the key of the leaf node, and as a c
 
 Given `B` substores in a module store and `N` entries in the substore, the module subtree will contain on average `Log(B) + Log(N)` layers. The expected number of operations needed to validate a witness from a leaf to the state root is `8 * MODULE_ID_LENGTH` (for the module ID), plus `8 * STORE_PREFIX_ID_LENGTH` (for the store prefix), plus `Log(N)` for the module subtree. Empty hashes are not included in the witness, and the resulting witness size is therefore `Log(M) + Log(B) + Log(N)`, where `M` indicates the number of modules registered in the chain.
 
+### Module ID
```
janhack, line 55: I would use a different name here as the new identifier for modules is now the module name. I would see what is called "module ID" at the moment rather as an internal detail of the state store that generally the developer should not need to know about or care about. It should also not be used as identifier for modules in other context, i.e., methods should use the module name and not the "module ID".

From the terminology so far, we say every module has its own `module store` which is again subdivided into `substores`. From this, the most natural terminology would be to say `store prefix` or `module store prefix` for the first 4 bytes and `substore prefix` for the 2 bytes after that.

proposals/lip-0040.md
```diff
@@ -67,20 +75,35 @@ We define the following constants:
...
 
 In summary, for an entry in the global state store identified by module ID `moduleID`, store prefix `storePrefix`, store key `storeKey`, and store value `storeValue`, the corresponding leaf node is added to the state tree using `SMT.update(moduleID || storePrefix || hash(storeKey), hash(storeValue))`, where `||` indicates the bytes-concatenation operation. This leaf node will have the following properties:
```
janhack, line 83:
```suggestion
In summary, for an entry in the global state store identified by module ID `moduleID`, store prefix `storePrefix`, store key `storeKey`, and store value `storeValue`, the corresponding leaf node is added to the state tree using `SMT.update(moduleID || storePrefix || hash(storeKey), hash(storeValue))`. This leaf node will have the following properties:
```

proposals/lip-0040.md
```diff
@@ -67,20 +75,35 @@ We define the following constants:
...
 
-```
-leaf.key = moduleID || storePrefix || hash(storeKey),
-leaf.value = hash(storeValue),
+```python
+leaf.key = moduleID + storePrefix + SHA256(storeKey),
```
janhack, line 86: Is there a specific reason why the hash function name is upper case in contrast to other function names we use? In Python3 it is also just `sha256`.

proposals/lip-0065.md
```diff
@@ -144,23 +158,33 @@ eventSchema = {
...
+The event `topics` is an array of unique byte values whose length ranges between `1` and `MAX_TOPICS_PER_EVENT`. By default, it contains one entry for the transaction ID (in the case of events emitted during the transaction processing stages), the constant `EVENT_TOPIC_INIT_GENESIS_STATE` (for events emitted during the 'genesis state initialization' stage), `EVENT_TOPIC_FINALIZE_GENESIS_STATE` (for events emitted during the 'genesis state finalization' stage), `EVENT_TOPIC_BEFORE_TRANSACTIONS` (for events emitted during the 'before transactions execution' stage), or the constant `EVENT_TOPIC_AFTER_TRANSACTIONS` (for events emitted during the 'after transactions execution' stage). Events emitted as part of the processing of a cross-chain message also contain an entry for the cross-chain message ID. An event emitted during transaction processing cannot have any of these default values in the `topics` array.
```
janhack, line 183:
```suggestion
The event `topics` is an array of unique byte values whose length ranges between `1` and `MAX_TOPICS_PER_EVENT`. By default, it contains one entry for the transaction ID (in the case of events emitted during the transaction processing stages), the constant `EVENT_TOPIC_INIT_GENESIS_STATE` (for events emitted during the 'genesis state initialization' stage), `EVENT_TOPIC_FINALIZE_GENESIS_STATE` (for events emitted during the 'genesis state finalization' stage), `EVENT_TOPIC_BEFORE_TRANSACTIONS` (for events emitted during the 'before transactions execution' stage), or the constant `EVENT_TOPIC_AFTER_TRANSACTIONS` (for events emitted during the 'after transactions execution' stage). Events emitted as part of the processing of a cross-chain message also by default always contain an entry for the cross-chain message ID. An event emitted during transaction processing cannot have any of these default values in the `topics` array.
```

janhack: The topic order is also important and should be specified. I assume that for events emitted in the CCM processing, the transaction ID of the CCU is the first topic and the cross-chain message ID the second topic.

proposals/lip-0065.md
```diff
@@ -187,8 +211,8 @@ In this LIP we introduce a new block header property, `eventRoot`. The events ro
 
 Each module defines a standard event indicating the result of a transaction processing, with the following properties:
 
-* `moduleID`: The module ID of the transaction.
-* `typeID`: The constant `EVENT_TYPE_ID_STANDARD`.
+* `module`: The `module` of the transaction.
```
janhack, line 214:
```suggestion
* `module`: The name of the module of the command executed by the transaction.
```

proposals/lip-0065.md
```diff
@@ -187,8 +211,8 @@ In this LIP we introduce a new block header property, `eventRoot`. The events ro
...
+* `name`: The constant `EVENT_NAME_STANDARD`.
 * `topics`: An array containing the transaction ID.
```
janhack, line 216:
```suggestion
* `topics`: An array containing only the transaction ID.
```

proposals/lip-0065.md
janhack: I would suggest `CommandExecutionResult` as only the command execution can fail and this event provides the information if this hook failed.

janhack: If we change the name of this event, I would also not take about "standard events" any more, maybe rename the respective section to "Default Events" and say that by default every module defines the `CommandExecutionResult` event.

janhack (changes requested):
There are a few open points from my side, otherwise it looks all good.

proposals/lip-0040.md
ricott1: no idea of the reason, but it is the standard in several other LIPs (as far back as LIP 3)

ricott1: that said, I think it would be completely clear to use `sha256` in the function

proposals/lip-0065.md
ricott1: thanks, fixed

proposals/lip-0040.md
ricott1: sounds good, changed to store and substore prefixes

proposals/lip-0065.md
ricott1: changed to default

shuse2 (approved):

janhack (approved):
Looks all good!

karmacoma (approved):

PR #167, opened 2022-08-22
Update network ID to chain ID in the protocol

ricott1:
- Update terminology: use chain ID rather than network ID in LIP37 and LIP62
- Update rationale and specify format for chain ID in LIP37
- Add BLS examples to LIP62


proposals/lip-0037.md
```diff
@@ -33,22 +33,24 @@ Replay attacks for transactions and blocks, i.e. replaying transactions or block
...
 
 ### Message Tags
 
-Every binary message must be tagged before being signed. This is done as specified by the function `tagMessage` in the pseudo code below: A specific ASCII-encoded tag and a network identifier as specified in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) are prepended to the message.
+Every binary message must be tagged before being signed. This is done as specified by the function `tagMessage` in the pseudo code below: A specific ASCII-encoded tag and a chain identifier as specified in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) are prepended to the message.
 
 ```python
-def tagMessage(tag: bytes, networkIdentifier: bytes, message: bytes) -> bytes:
+def tagMessage(tag: bytes, chainID: bytes, message: bytes) -> bytes:
     return tag + networkIdentifier + message
```
sergeyshemyakov, line 54: `tag + chainID + message`

proposals/lip-0037.md
```diff
@@ -67,26 +69,37 @@ The tag is chosen depending on the message type. Moreover, different serializati
...
+
+| Name                            | Type  | Value| Description                                                 |
+|---------------------------------|-------|------|-------------------------------------------------------------|
+| `CHAIN_ID_PREFIX_MAINNET`       | bytes | 0x00 | Chain-identifiers prefix for chains running in the mainnet. |
+| `CHAIN_ID_PREFIX_TESTNET`       | bytes | 0x01 | Chain-identifiers prefix for chains running in the testnet. |
+
 ### Signing and Verifying with Ed25519
 
-Let [`Sign`](https://tools.ietf.org/html/rfc8032#section-5.1.6) and [`Verify`](https://tools.ietf.org/html/rfc8032#section-5.1.7) be the signing and verifying functions of Ed25519 as specified in [RFC 8032](https://tools.ietf.org/html/rfc8032). Then the signature for a binary message `m` and a secret key `sk` is generated by `signMessage(sk, tag, networkIdentifier, m)` as defined below. `tag` must be the correct message tag for `m` and `networkIdentifier` the correct network identifier for the chain. The resulting signature `sig` in combination with the message `m` and the matching public key `pk` is verified by `verifyMessageSig(pk, tag, networkIdentifier, m, sig)`.
```
sergeyshemyakov, line 79: The very end of the paragraph:
`pk` is verified by `verifyMessageSig(pk, tag, networkIdentifier, m, sig)`. -> `pk` is verified by `verifyMessageSig(pk, tag, chainID, m, sig)`.

sergeyshemyakov (commented):
These two LIPs look good.
A question: I have also looked for all occurrences of "network identifier" in the lips repo and I have saw the following LIPs:

- LIP 38 BLS signatures. However there are examples of messages and signatures at the bottom, probably the shortening of netID to 4-byte chainID will require to recompute some examples.
- LIP 55 just two small changes.
- LIP 61 certificates. Should also be an easy change.

I don't know how we should update the mentioned LIPs, maybe some should be added to this PR.

proposals/lip-0037.md
```diff
@@ -1,6 +1,6 @@
 ```
 LIP: 0037
-Title: Use message tags and network identifiers for signatures
+Title: Use message tags and chain identifiers for signatures
 Author: Andreas Kendziorra <andreas.kendziorra@lightcurve.io>
 Discussions-To: https://research.lisk.com/t/use-message-tags-and-network-identifiers-for-signatures/
 Status: Draft
```
janhack, line 6: Should we add that this LIP replaces LIP 9?

ricott1: Is there a standard way to do this in the header?

proposals/lip-0037.md
```diff
@@ -33,23 +33,25 @@ Replay attacks for transactions and blocks, i.e. replaying transactions or block
...
 
-### Network Identifiers
+### Chain Identifiers
 
-Network identifiers for transaction signatures and block signatures were already introduced in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) and [LIP 0024](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0024.md#update-to-the-block-header-signing-procedure) respectively to prevent replay attacks on other chains. Here, we specify how to use a network identifier for any kind of message to prevent replay attacks on other chains in general.
+Chain identifiers for transaction signatures and block signatures were already introduced in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) and [LIP 0024](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0024.md#update-to-the-block-header-signing-procedure) (as network identifiers) to prevent replay attacks on other chains. Here, we specify how to use a chain identifier for any kind of message to prevent replay attacks on other chains in general.
```
janhack, line 42: I would write here explicitly that the specifications in this LIP supersede those in LIP 9. The way in which we use chain ID for signing is the same as we used network identfiers, however, the length and how they are obtained is now different. I think it would be nice to also elaborate on the reasoning, for instance, something like the following:

> For users to be able to easily verify that they sign a transaction for the correct blockchain, we choose to abandon the usage of 32-byte network identifiers that are computed as a hash value from the genesis block ID and community identifier. Instead, the new chain identifier which also serve the purpose to prevent transaction replay among different blockchains, has a length of only 4 bytes. It further can be chosen by the blockchain creator. For the official mainnet blockchain of a project, we always recommend to register the identifier on the Lisk Mainchain to ensure uniqueness in the Lisk ecosystem.

proposals/lip-0037.md
janhack: See https://github.com/LiskHQ/lips/blob/main/proposals/lip-0001.md#lip-header-preamble

proposals/lip-0037.md
```diff
@@ -33,23 +33,25 @@ Replay attacks for transactions and blocks, i.e. replaying transactions or block
...
+
+Chain identifiers are 4-byte values that follow a specific format: the first byte is used to identify the network in which the chain is running (either the mainnet, official testnet, or any other test network); the other 3 bytes identify the chain within the network. We include the network-specific prefix explicitly to ensure that a chain does not use the same chain identifier in the test network as in the mainnet.
 
 ## Specification
 
 ### Message Tags
 
-Every binary message must be tagged before being signed. This is done as specified by the function `tagMessage` in the pseudo code below: A specific ASCII-encoded tag and a network identifier as specified in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) are prepended to the message.
+Every binary message must be tagged before being signed. This is done as specified by the function `tagMessage` in the pseudo code below: A specific ASCII-encoded tag and a chain identifier as specified in [LIP 0009](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0009.md#specification) are prepended to the message.
```
janhack, line 50: The chain identifier should be specified in this LIP and is different compared to the network identifier in LIP 9 which is defined as `network identifier = H(nethash + community identifier)`.

proposals/lip-0037.md
```diff
@@ -67,26 +69,37 @@ The tag is chosen depending on the message type. Moreover, different serializati
...
+    "LSK_" + TAG + "_"
 
 has to be used where:
 
 - Strings in double quotes are ASCII-encoded literals.
 - TAG is a unique string indicating the scheme and, optionally, additional information, e.g., to specify a version number as shown [above](#message-tags). TAG must contain only ASCII characters between 0x21 and 0x7e (inclusive), except that it must not contain underscore (0x5f).
 
+### Chain Identifiers
+
+Chain identifiers are 4-byte values. The first byte is set to `CHAIN_ID_PREFIX_MAINNET` for chains running in the mainnet network and to `CHAIN_ID_PREFIX_TESTNET` for chains running in the testnet network. The other 3 bytes are unique to that chain.
```
janhack, line 81:
```suggestion
Chain identifiers are 4-byte values. The first byte is set to `CHAIN_ID_PREFIX_MAINNET` for chains running in the mainnet network and to `CHAIN_ID_PREFIX_TESTNET` for chains running in the testnet network. The other 3 bytes must be chosen uniquely for the respective blockchain, i.e., no other blockchain created with the Lisk SDK should use the same 3 bytes.
```

janhack: To clarify what means unique here.

proposals/lip-0037.md
```diff
@@ -67,26 +69,37 @@ The tag is chosen depending on the message type. Moreover, different serializati
...
+
+The following table defines the chain-identifiers prefixes currently specified.
+
+| Name                            | Type  | Value| Description                                                 |
+|---------------------------------|-------|------|-------------------------------------------------------------|
+| `CHAIN_ID_PREFIX_MAINNET`       | bytes | 0x00 | Chain-identifiers prefix for chains running in the mainnet. |
```
janhack, line 87:
```suggestion
| `CHAIN_ID_PREFIX_MAINNET`       | bytes | 0x00 | Chain-identifier prefix for mainnet blockchains. |
```

janhack: "Running in the mainnet" sounds odd.

janhack: Do we mention here that it will be a requirement to have the prefix `0x00` to connect to Lisk Mainnet for interoperability (I am assuming there that we would make this a requirement) and reference to the chain registration LIP?

proposals/lip-0037.md
```diff
@@ -67,26 +69,37 @@ The tag is chosen depending on the message type. Moreover, different serializati
...
+| `CHAIN_ID_PREFIX_TESTNET`       | bytes | 0x01 | Chain-identifiers prefix for chains running in the testnet. |
```
janhack, line 88:
```suggestion
| `CHAIN_ID_PREFIX_TESTNET`       | bytes | 0x01 | Chain-identifier prefix for testnet blockchains. |
```

proposals/lip-0038.md
```diff
@@ -79,41 +80,46 @@ The Lisk protocol could contain several transaction types that perform such a re
...
 
 ```python
-createAggSig(keysList, pubKeySignaturePairs):
+def createAggSig(keysList: list[bytes], pubKeySignaturePairs: list[tuple[bytes, bytes]]) -> tuple[bytes, bytes]:
     aggregationBits = byte string of length ceil(length(keyList)/8) with all bytes set to 0
```
janhack, line 87: integer division `//8` ?

proposals/lip-0062.md
```diff
@@ -78,3 +78,30 @@ TBD
...
+
+#### `verifyBLS`
+
+```python
+pk = 0xa491d1b0ecd9bb917989f0e74f0dea0422eac4a873e5e2644f368dffb9a6e20fd6e10c1b77654d067c0618f6e5a7f79a
+tag = b"LSK_TX_"
+chainID = 0x00000000
+message = 0xbeaf
+sig = 0x80c3da661b5bb80bb841367255f7b087b969c075661895b7ac8b74b72360be54693b3485eff7d816924517a21ef1c3a30a8f9402572d5a63a7ff2f71ca6929a8c3d7f75fd72edd1aa478ecc09966a133e829600f0111a1e40bbe35db61e8c689
+verifyBLS(pk, tag, chainID, message, sig) = VALID
```
janhack, line 106: - Don't we assume now that the function returns a Boolean?
- Shouldn't it be `==` instead of `=`?

proposals/lip-0062.md
```diff
@@ -78,3 +78,30 @@ TBD
...
+The following tests supersede the corresponding tests given in [LIP 0038](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0038.md#test-vectors).
+
+#### `signBLS`
+
+```python
+sk = 0x263dbd792f5b1be47ed85f8938c0f29586af0d3ac7b977f21c278fe1462040e3
+tag = b"LSK_TX_"
+chainID = 0x00000000
+message = 0xbeaf
+signBLS(sk, tag, chainID, message) = 0x80c3da661b5bb80bb841367255f7b087b969c075661895b7ac8b74b72360be54693b3485eff7d816924517a21ef1c3a30a8f9402572d5a63a7ff2f71ca6929a8c3d7f75fd72edd1aa478ecc09966a133e829600f0111a1e40bbe35db61e8c689
```
janhack, line 95: - Shouldn't it be `==` instead of `=`?

proposals/lip-0038.md
```diff
@@ -323,20 +329,20 @@ PopVerify(pk, proof) = INVALID
...
 
 ```python
 pk = 0xa491d1b0ecd9bb917989f0e74f0dea0422eac4a873e5e2644f368dffb9a6e20fd6e10c1b77654d067c0618f6e5a7f79a
 tag = 0x4c534b5f54585f # ASCII-encoding of "LSK_TX_"
-networkIdentifier = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
+chainID = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
 message = 0xbeaf
 sig = 0xa6f889695b4ee393c6ede6af2215019cf9d7e004781b98ea12d6b227212126687ecde1c2a08e38e2d5c18eab2881879102e91ac5f0e9813126d6d68262af149ba6c25ffb88e6688fec49b5199cec863c0eb54110fdb6d92c6570f3ca9c1910b9
-verifyBLS(pk, tag, networkIdentifier, message, sig) = VALID
+verifyBLS(pk, tag, chainID, message, sig) = VALID
```
janhack, line 345:
```suggestion
verifyBLS(pk, tag, chainID, message, sig) == True
```

proposals/lip-0038.md
```diff
@@ -395,9 +401,9 @@ createAggSig(KEYS_LIST, pubKeySignaturePairs) = [0x4001, 0x89a3b2642b285b129ff91
 aggregationBits = 0x4001
 signature = 0x89a3b2642b285b129ff9196fa310eea2fb77468b3919f6530ebef402a5c7d589b3d6cfa0bb884b38b365047a7a6a216614a72941479d2753438b3657a430318f43dbfd895632db1c23d988097115fe1d0c4aa5d2ef42da50b8027bd2e410d781
 tag = 0x4c534b5f43455f # ASCII-encoding of "LSK_CE_"
-networkIdentifier = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
+chainID = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
 message = 0xbeaf
-verifyAggSig(KEYS_LIST, aggregationBits, signature, tag, networkIdentifier, message) = VALID
+verifyAggSig(KEYS_LIST, aggregationBits, signature, tag, chainID, message) = VALID
```
janhack, line 406:
```suggestion
verifyAggSig(KEYS_LIST, aggregationBits, signature, tag, chainID, message) == True
```

proposals/lip-0038.md
```diff
@@ -323,20 +329,20 @@ PopVerify(pk, proof) = INVALID
 ```python
```
janhack, line 329: For the test vectors, we assume that the functions `Verify` and `FastAggregateVerify` return either `VALID` or `INVALID`, which contradicts the specifications above. We could change it here or update the function code above to something like
```
return FastAggregateVerify(keys, taggedMessage, signature) == VALID
```

janhack: Moreover, in the test vectors we always use `=` instead of `==` in the last line.

janhack (changes requested):
There are a few minor points that I commented on, otherwise it looks good.

proposals/lip-0037.md
ricott1: I thought that this LIP should be unaware of other modules using the chainID for something else. Because otherwise we would also need to add the requirement that all token IDs in the mainnet start with that prefix etc..

proposals/lip-0038.md
ricott1: but here we take the ceil rather than the floor. For sure there is a nice way of doing it, I was just trying not to overthink it

proposals/lip-0037.md
ricott1: added

proposals/lip-0037.md
ricott1: added something in this direction

proposals/lip-0037.md
ricott1: Do you think I have to write this explicitly? It is somehow obvious that here we are given the new specs

sergeyshemyakov (approved):
Looks good now.

proposals/lip-0037.md
janhack: I like the new formulation :-)

proposals/lip-0037.md
janhack:
```suggestion
Every binary message must be tagged before being signed. This is done as specified by the function `tagMessage` in the pseudo code below: A specific ASCII-encoded tag and a chain identifier are prepended to the message.
```

janhack: I just wanted to say the reference to LIP 9 here is rather confusing then helpful so I would just remove it.

proposals/lip-0038.md
janhack: Ah, true, all good.

proposals/lip-0038.md
```diff
@@ -295,48 +301,48 @@ Invalid cases:
...
+PopVerify(pk, proof) == INVALID
 ```
 
 #### `signBLS`
 
 ```python
 sk = 0x263dbd792f5b1be47ed85f8938c0f29586af0d3ac7b977f21c278fe1462040e3
 tag = 0x4c534b5f54585f # ASCII-encoding of "LSK_TX_"
-networkIdentifier = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
+chainID = 0x9ee11e9df416b18bf69dbd1a920442e08c6ca319e69926bc843a561782ca17ee
```
janhack, line 332: Here the `chainID` is not 4-bytes, but I guess it should not impact the test.

proposals/lip-0038.md
```diff
@@ -79,42 +80,47 @@ The Lisk protocol could contain several transaction types that perform such a re
...
             keys.append(keysList[i])
             weightSum += weights[i]
     if weightSum < threshold:
-        return INVALID
-    return FastAggregateVerify(keys, taggedMessage, signature)
+        return False
+    return FastAggregateVerify(keys, taggedMessage, signature) == VALID
 ```
 
 Note that the public keys in `pubKeySignaturePairs` need to be distinct when calling `createAggSig.` Otherwise, validation via `verifyAggSig` and `verifyWeightedAggSig` will fail.
```
janhack, line 126: In the test vector section below there are still some occurrences of `=` instead of `==`.

proposals/lip-0062.md
```diff
@@ -39,35 +39,72 @@ Introduction of another hash function to the Lisk protocol was suggested in [LIP
...
+### Signing and Verifying with Ed25519
+
+The Ed25519 signature for a binary message `message` and a secret key `sk` is generated by `signMessage(sk, tag, chainID, message)` as defined below. `tag` must be the correct message tag for `message` as defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md), and `chainID` the correct chain ID for the chain. The resulting signature `signature` in combination with the message `message` and the matching public key `pk` is verified by `verifyMessageSig(pk, tag, chainID, message, signature)`.
 
 ```python
-signMessage(sk, tag, networkIdentifier, message):
-    taggedMessage = tag || networkIdentifier || message
-    hashedMessage = hash(taggedMessage)
+def signMessage(sk: bytes, tag: bytes, chainID: bytes, message: bytes) -> bytes:
```
janhack, line 47: For clarity, should we rename `signMessage` to `signEd25519` (or something similar) because we also have `signBLS`.

proposals/lip-0062.md
```diff
@@ -39,35 +39,72 @@ Introduction of another hash function to the Lisk protocol was suggested in [LIP
...
 The `signMessage` and `verifyMessageSig` functions defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#signing-and-verifying-with-ed25519) are superseded by the functions defined in this LIP.
 
 ### Signing and Verifying with BLS
 
-For the BLS signature scheme [used in Lisk](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0038.md#specification), `Sign` (identical to <code>[CoreSign](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.6)</code>) and `Verify` (identical to <code>[CoreVerify](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.7)</code>) are the signing and verifying functions as specified in [BLS Signatures draft-irtf-cfrg-bls-signature-04](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04).
+```python
+def signBLS(sk: bytes, tag: bytes, chainID: bytes, message: bytes) -> bytes:
+    taggedMessage = tag + chainID + message
+    hashedMessage = sha256(taggedMessage)
+    return CoreSign(sk, hashedMessage)
```
janhack, line 69:
```suggestion
    return Sign(sk, hashedMessage)
```

janhack: To use the same name as in LIP 38.

proposals/lip-0062.md
```diff
@@ -39,35 +39,72 @@ Introduction of another hash function to the Lisk protocol was suggested in [LIP
...
+```
+```python
+def verifyBLS(pk: bytes, tag: bytes, chainID: bytes, message: bytes, signature: bytes) -> bool:
+    taggedMessage = tag + chainID + message
+    hashedMessage = sha256(taggedMessage)
+    return CoreVerify(pk, hashedMessage, signature) == VALID
```
janhack, line 75:
```suggestion
    return Verify(pk, hashedMessage, signature) == VALID
```

janhack (changes requested):
I have a few minor, short points left, otherwise the PR is basically ready.

Did @ishantiw review the new BLS test cases? Because, I did not check any of the test cases in the LIP.

proposals/lip-0038.md
ricott1: Indeed, I prefer not to change these test vectors that were prepared by Andreas

proposals/lip-0038.md
ricott1: fixed

proposals/lip-0062.md
ricott1: In LIP38 it is exaplined that indeed they are the same functions (here I rather used the Core name to avoid confusion and link directly to specs), so I don't think it is needed to change LIP38. What do you mean by ' Same for CoreVerify below.'?

proposals/lip-0062.md
ricott1: I agree, but I will ask since this could impact other LIPs/code

proposals/lip-0062.md
janhack: I meant the same applies to `CoreVerify` and `Verify`.

Okay, then let us keep it like it is.

janhack (approved):
Looks all good.

karmacoma (approved):

PR #170, opened 2022-08-30
Update lip 0060

ricott1:
Fix links in LIP60


proposals/lip-0060.md
```diff
@@ -116,7 +116,7 @@ The block header is processed as follows:
   * Check that the block header follows the block header schema.
   * The value `b.header.version` can be any `uint32` integer.
```
sergeyshemyakov, line 117: `b.header.version` is always some `uint32` integer.

proposals/lip-0060.md
```diff
@@ -116,7 +116,7 @@ The block header is processed as follows:
...
   * The value of `b.header.transactionRoot` is equal to `EMPTY_HASH`.
-  * The value `b.header.assetsRoot` is validated as specified in [Update block schema and block processing LIP][research:update-block-schema#assets-root].
+  * The value `b.header.assetsRoot` is validated as specified in [LIP 0055][lip-0055#assets-root].
   * The value `b.header.timestamp` is Unix time in seconds and can be any value in the `uint32` range.
```
sergeyshemyakov, line 120: Do we need to check that the timestamp "is Unix time in seconds"? Or do we imply only the check "can be any value in the `uint32` range"? 
Because the second check is always true and thus I would remove this line.

proposals/lip-0060.md
```diff
@@ -116,7 +116,7 @@ The block header is processed as follows:
...
   * The value `b.header.height` can be any value in the `uint32` range.
   * The value `b.header.previousBlockID` can be any 32-byte value.
```
sergeyshemyakov, line 122: The same as above.

sergeyshemyakov (approved):
Looks good, but have a look at my 3 remarks about `uint32` checks.

proposals/lip-0060.md
ricott1: I think it's nice to have the extra clarity

proposals/lip-0060.md
janhack: I think this is just a way of saying there are no additional validations.

proposals/lip-0060.md
janhack: I would also keep it as is.

janhack (approved):
Looks all good.

karmacoma (approved):

PR #171, opened 2022-09-05
Update LIP 0063 "Define mainnet configuration and migration for Lisk Core v4" with respect to the last module updates

sergeyshemyakov:


proposals/lip-0063.md
```diff
@@ -365,19 +375,20 @@ def addTokenModuleEntry():
     tokenObj.supplySubstore = createSupplySubstoreArray()
     tokenObj.escrowSubstore = []
     tokenObj.availableLocalIDSubstore = {}
+    tokenObj.supportedTokensSubstore = [{"chainID": MAINCHAIN_ID, "supportedTokenIDs": [TOKEN_ID_LSK]}]
```
janhack, line 378: Checking the function `isTokenSupported` in the Token module and rationale section, I think this is not needed (although it is also not incorrect).  Not sure about this one.

janhack (approved):
I have one small optional discussion point, but the PR looks good so I already approve it.

proposals/lip-0063.md
sergeyshemyakov: Agree, removed. I think, we should not put LSK into supported tokens substore.

vjaiman (approved):
Looks all good.

karmacoma (approved):

PR #174, opened 2022-09-23
Update LIP 0051 (Token Module) for events, error-handling

gkoumout:


sergeyshemyakov (approved):
I checked for typos and haven't found any.

proposals/lip-0051.md
```diff
@@ -123,71 +128,166 @@ The following constants are used throughout the document:
...
+| `EVENT_NAME_BEFORE_CCM_FORWARDING`   | string  | "beforeCCMForwarding"  | Used for events emitted during calls to the [beforeCrossChainMessageForwarding](#beforeCrossChainMessageForwarding) function. |
+| `EVENT_NAME_ALL_TOKENS_SUPPORTED`| string  | TBD    | Used for the [allTokensSupported](#alltokenssupported) event  |
+| `EVENT_NAME_ALL_TOKENS_SUPPORTED_REMOVED`| string  | TBD    | Used for the [allTokensSupportedRemoved](#alltokenssupportedremoved) event  |
+| `EVENT_NAME_ALL_TOKENS_FROM_CHAIN_SUPPORTED`| string  | TBD    | Used for the [allTokensFromChainSupported](#alltokensfromchainsupported) event  |
+| `EVENT_NAME_ALL_TOKENS_FROM_CHAIN_SUPPORTED_REMOVED`| string  | TBD    | Used for the [allTokensFromChainSupportedRemoved](#alltokensfromchainsupportedremoved) event  |
+| `EVENT_NAME_TOKEN_ID_SUPPORTED`      | string  | TBD    | Used for the [tokenIDSupported](#tokenidsupported) event  |
+| `EVENT_NAME_TOKEN_ID_SUPPORT_REMOVED`      | string  | TBD    | Used for the [tokenIDSupportRemoved](#tokenidsupportremoved) event  |
```
shuse2, line 188: I think since the last event is `EVENT_NAME_TOKEN_ID_SUPPORT_REMOVED`, `EVENT_NAME_ALL_TOKENS_SUPPORTED_REMOVED` and `EVENT_NAME_ALL_TOKENS_FROM_CHAIN_SUPPORTED_REMOVED` these should be `XXX_SUPPORT_REMOVED` event

proposals/lip-0051.md
```diff
@@ -503,242 +641,1892 @@ crossChainTransferMessageParams = {
...
+        },
+        "result": {
+            "dataType": "uint32",
+            "fieldNumber": 2
+        }
+    }
+}
+```
+
+#### initializeUserStore
```
shuse2, line 1120: `user store` is internal name, so maybe better to say `initializeUserAccount`?

proposals/lip-0051.md
```diff
@@ -503,242 +641,1892 @@ crossChainTransferMessageParams = {
...
+        },
+        "result": {
+            "dataType": "uint32",
+            "fieldNumber": 5
+        }
+    }
+}
+```
+
+#### initializeEscrowStore
```
shuse2, line 1162: same as above, maybe better to say `initializeEscrowAccount`?

proposals/lip-0051.md
```diff
@@ -814,484 +2596,147 @@ genesisTokenStoreSchema = {
...
-  * Across all element of the `supplySubstore` array, all values given for `localID` must be unique.
-  * The `supplySubstore` array must be in ascending order of `localID`.
-  * Across all elements of the `escrowSubstore` array, the pairs `escrowChainID, localID` must be unique (each pair appears at most once).
-  * The `escrowSubstore` array must be in ascending order of `escrowChainID`. For a given `escrowChainID`, the entries must be in ascending order of `localID`.
-  * In the `terminatedEscrowSubstore` array, each element must be unique (appears at most once).
-  * The `terminatedEscrowSubstore` array must be in ascending order.
+  * Across all element of the `supplySubstore` array, all values given for `tokenID` must be unique.
+  * The `supplySubstore` array must be in ascending order of `tokenID`.
+  * Across all elements of the `escrowSubstore` array, the `tokenID` must be unique (it appears at most once).
+  * The `escrowSubstore` array must be in ascending order of `tokenID`.
```
shuse2, line 2660: missing validation for supportedTokenIDs store?

proposals/lip-0051.md
```diff
@@ -814,484 +2596,147 @@ genesisTokenStoreSchema = {
...
-          if chainID == CHAIN_ID_ALIAS_NATIVE:
-              computedSupplies[localID] += sum of availableBalance and all locked amounts of storeValue
+          tokenID = storeKey[ADDRESS_LENGTH:ADDRESS_LENGTH + TOKEN_ID_LENGTH] # Part of the key corresponding to the token ID of the token
+          if getChainID(tokenID) == OWN_CHAIN_ID:
+              computedSupplies[tokenID] += sum of availableBalance and all locked amounts of storeValue
 
       for (storeKey, storeValue) in escrow substore:
-          localID = storeKey[4:8] # Part of the key corresponding to the local ID of the token
-          computedSupplies[localID] += storeValue.amount
+          tokenID = storeKey[:8] # Part of the key corresponding to the token ID of the token
```
shuse2, line 2717: key for escrow is tokenID so no need to slice here?

proposals/lip-0051.md
```diff
@@ -205,41 +305,20 @@ supplyStoreSchema = {
...
-The default value for this substore is `{"nextAvailableLocalID": 0}` serialized using `availableLocalIDStoreSchema`.
 
 #### Escrow Substore
 
 The Token module store contains an entry dedicated to storing information about native tokens which have been sent to another chain. The state contains an entry with:
 
-* The store prefix is set to `STORE_PREFIX_ESCROW`.
-* Each store key is the identifier of the chain to which the tokens are escrowed, and the local ID of the escrowed token: `uint32be(escrowedChainID)||uint32be(tokenLocalID)`.
+* The substore prefix is set to `SUBSTORE_PREFIX_ESCROW`.
+* Each store key is the identifier of the chain to which the tokens are escrowed, and the token ID of the escrowed token: `escrowedChainID + tokenID`.
```
shuse2, line 315: this should be just `tokenID`

proposals/lip-0051.md
```diff
@@ -253,65 +332,50 @@ If any state transition would reduce the `amount` property of an entry to zero,
...
     "properties": {
-        "availableBalance": {
-            "dataType": "uint64",
-            "fieldNumber": 1
-        },
-        "lockedBalances": {
-            "type": "array",
-            "fieldNumber": 2,
+        "supportedTokenIDs" : {
+            "dataType": "array",
```
shuse2, line 350:
```suggestion
            "type": "array",
```

proposals/lip-0051.md
```diff
@@ -814,484 +2596,147 @@ genesisTokenStoreSchema = {
...
             "type": "array",
-            "fieldNumber": 5,
+            "fieldNumber": 4,
             "items": {
-                "type": "uint32"
+                "type": object,
+                "required": ["chainID", "supportedTokenIDs"],
+                "properties": {
+                    "chainID": {
+                        "type": "bytes",
```
shuse2, line 2621:
```suggestion
                        "dataType": "bytes",
```

proposals/lip-0051.md
```diff
@@ -814,484 +2596,147 @@ genesisTokenStoreSchema = {
...
+                        "length": CHAIN_ID_LENGTH,
+                        "fieldNumber": 1
+                    },
+                    "supportedTokenIDs" : {
+                        "dataType": "array",
```
shuse2, line 2626:
```suggestion
                        "type": "array",
```

proposals/lip-0051.md
```diff
@@ -814,484 +2596,147 @@ genesisTokenStoreSchema = {
...
+)
+```
 
 * For each entry `escrowEntry` in `genesisBlockAssetObject.escrowSubstore`, create an entry in the escrow substore with:
-  ```python
-  storeKey = uint32be(escrowEntry.escrowChainID) || uint32be(escrowEntry.localID)
-  storeValue = {"amount": escrowEntry.amount} serialized using escrowStoreSchema.
-  ```
+```python
+storeKey = escrowEntry.tokenID
```
shuse2, line 2687: store key should be concat with escrowChainID?

proposals/lip-0051.md
gkoumout: Good point. Fixed.

proposals/lip-0051.md
gkoumout: changed.

proposals/lip-0051.md
gkoumout: done.

proposals/lip-0051.md
gkoumout: Good catch. Perhaps this was not updated at all since we don't even use the proper constant. In fact, the escrow substore has keys `escrowedChainID + tokenID`, so I think it should be 

`tokenID = storeKey[CHAIN_ID_LENGTH:]`

proposals/lip-0051.md
gkoumout: I think it is correct as is: escrow has as key both `chainID` and `tokenID`.

shuse2: ah yes, i think genesis schema `escrowChainID`  was removed, so i got confused. This is correct, but the genesis escrow schema should have `escrowChainID`?

proposals/lip-0051.md
gkoumout: Good catch. Added.

proposals/lip-0051.md
gkoumout: This is correct, added.

proposals/lip-0051.md
gkoumout: You are right. Added.

proposals/lip-0051.md
```diff
@@ -2420,19 +2419,18 @@ def supportAllTokensFromChainID(chainID: ChainID) -> None:
...
 
-This function is called in case where all tokens from a specific chain are supported and we want to remove this support. After calling this function no tokens of the specified chain are supported. 
-
+This function is called to remove support for all tokens of a specified chain. 
 
 ```python
 def removeAllTokensSupportFromChainID(chainID: ChainID) -> None:
     if there exists entry in the supported tokens substore with key == ALL_SUPPORTED_TOKENS_KEY:
-        return
+        raise Exception('Invalid operation. All tokens from all chains are supported.')
```
shuse2, line 2427: i think this operation is not needed because if chainID doesn't exist in data, it will fail, and if it exist it never has ALL_SUPPORTED_TOKEN.

Or if we say
```
if supportedTokens(chainID) does not exist:
        return
```
as below, this should also return?

proposals/lip-0051.md
```diff
@@ -2420,19 +2419,18 @@ def supportAllTokensFromChainID(chainID: ChainID) -> None:
...
 
-    if chainID == OWN_CHAIN_ID:
-        return
+    if chainID == OWN_CHAIN_ID or chainID == MAINCHAIN_ID:
```
shuse2, line 2429: in other function, it's just ignoring?

proposals/lip-0051.md
```diff
@@ -503,243 +644,1891 @@ crossChainTransferMessageParams = {
...
+        value = encode(
+            schema=supplyStoreSchema,
+            object={"totalSupply": 0}
+        )  
+
+    emitEvent(
+        module=MODULE_NAME_TOKEN,
+        name=EVENT_NAME_INITIALIZE_TOKEN,
+        data={
+                "tokenID": nextTokenID,
```
ricott1, line 1713: indendation error

proposals/lip-0051.md
```diff
@@ -503,243 +644,1891 @@ crossChainTransferMessageParams = {
...
+        raise Exception("Token ID is not from the current chain")
+    
+    return escrowAmount(escrowChainID, tokenID)
+```
+
+
+#### initializeToken
+
+```python
+def initializeToken() -> TokenID:
```
ricott1, line 1678: I think it would be better to allow to pass the desired tokenID as input and just check if it was used before (this is easy by a single read from the supply store). This is more consistent to what we currently do with chain IDs in interoperability (users can select the desired value as long as it is free), and it also slightly simplifies the protocol. 

I don't think it is really a problem that you would have to try several values, because this is arguably solved in the UI (same reasoning as in interoperability)

ricott1: If we go for this, we would remove the `MAX_AVAILABLE_ID_REACHED` result and just add a `TOKEN_ID_ALREADY_ASSIGNED` result (or smtg like that)

proposals/lip-0051.md
```diff
@@ -503,243 +644,1891 @@ crossChainTransferMessageParams = {
...
+    )
+    
+    # call interop. module to create CCM
+    Interoperability.send(
+        sendingAddress=senderAddress,
+        module=MODULE_NAME_TOKEN,
+        crossChainCommand=CROSS_CHAIN_COMMAND_NAME_TRANSFER,
+        receivingChainID=receivingChainID,
+        fee=messageFee,
+        status=CCM_STATUS_OK,
```
ricott1, line 1616: remove this line

proposals/lip-0051.md
```diff
@@ -321,178 +385,255 @@ The module provides the following commands to modify token entries.
...
-availableBalance(senderAddress, tokenID) -= amount
-availableBalance(recipientAddress, tokenID) += amount
+```python
+def execute(trs: Transaction):
+
+    senderAddress = SHA256(trs.senderPublicKey)[:ADDRESS_LENGTH]         # derive sender address from trs.senderPublicKey
+    tokenID = trs.params.tokenID
+    recipientAddress = trs.params.recipientAddress
+    amount = trs.params.amount
+    accountInitializationFee = trs.params.accountInitializationFee
```
ricott1, line 476: this variable is not used

gkoumout: Good catch, not used anymore in internal. Removed.

proposals/lip-0051.md
gkoumout: done.

proposals/lip-0051.md
gkoumout: fixed.

proposals/lip-0051.md
gkoumout: I think this is the only remaining open point now. I suggest discussing this with @MaximeGagnebin on Thursday and reach a conclusion.

MaximeGagnebin: For me the main drawback is that you don't have a good way to get an available ID. This was solved before with the  `nextAvailableID` store, but we removed this one.
To be clear, I think that we should really have a non-failing token initialization, or at least a way to call it without the risk of it failing. The reason being that this API call could happen as part of something way more general and you don't want the user to have to deal with the tokenID.

In general, I think I prefer the way it is now. If we are using UI arguments, you could also say that "whatever tokenID you get assigned, the UI will give it a proper name and will display it in a meaningful way".

MaximeGagnebin (approved):

proposals/lip-0051.md
```diff
@@ -503,240 +637,1915 @@ crossChainTransferMessageParams = {
...
+        emitFailedRecoverEvent(terminatedChainID, storeKey[:ADDRESS_LENGTH], EMPTY_BYTES, 0, RECOVER_FAIL_INVALID_INPUTS)
+        raise Exception("Invalid arguments.")
+
+    address = storeKey[:ADDRESS_LENGTH]
+    tokenID = storeKey[ADDRESS_LENGTH:ADDRESS_LENGTH + TOKEN_ID_LENGTH]
+    account = decode(schema=userStoreSchema, object=storeValue)
+    totalAmount = sum of availableBalance and all locked amounts of account
+
+    if (
+        getChainID(tokenID) != OWN_CHAIN_ID
```
ricott1, line 2325: In this case we should emit an event with a different error code and raise a different Exception (insufficient escrow is not really correct)

proposals/lip-0051.md
gkoumout: In the current version (after discussing within research last week) the following solution is chosen:

- we have a function `isTokenIDAvailable(tokenID)` to tell is the specified ID can be used
- `initializeToken(tokenID)` fails in case the `tokenID` is already used.
- Therefore `tokenID` is chosen by developers and is not consecutive.
- `getNextAvailableTokenID()` gives the ID after the maxID used until now. Therefore in case someone uses `tokenID` equal to the maximum uint32 value, then this function returns that there are no available tokenIDs..

@ricott1 @shuse2 @MaximeGagnebin Let me know if you think there is something extra to discuss here, or you think it can be merged like this..

ricott1: this looks fine to me

shuse2: LGTM

MaximeGagnebin: LGTM

ricott1 (approved):

karmacoma (approved):

PR #176, opened 2022-10-04
Update LIPs: Minor fixes to Interoperability

ricott1:
Implement some minor fixes to the Interoperability protocol:

- Automatically assign `CCM_STATUS_CODE_OK` when sending a new ccm
- Consistently emit ccm events after termination
- Fix channel terminated message properties and execution
- Add extra verification to sidechain registration command
- Various minor fixes
- Code refactoring


proposals/lip-0045.md
```diff
@@ -376,7 +380,7 @@ channelSchema = {
                     "fieldNumber": 1
                 },
                 "size": {
-                    "dataType": "uint64",
```
janhack, line 379: Why don't we use `uint64` everywhere here? In terms of storage it will be the same as long as we are in the `uint32` range. Note that we could probably have 1 M CCMs per day => 3600 M CCMs in 10 years.

proposals/lip-0045.md
```diff
@@ -741,6 +745,118 @@ def isLive(chainID: ChainID) -> bool:
         return True
 ```
 
+#### send
```
janhack, line 748:
```suggestion
#### sendInternal
```

proposals/lip-0045.md
```diff
@@ -741,6 +745,118 @@ def isLive(chainID: ChainID) -> bool:
...
+        "module": module,
+        "crossChainCommand": crossChainCommand,
+        "sendingChainID": ownChainAccount.chainID,
+        "receivingChainID": receivingChainID,
+        "fee": fee,
+        "status": status,
+        "params": EMPTY_BYTES
+    }
+
+    paramsSchema = JSON schema associated with (module, crossChainCommand)
```
janhack, line 773: I would move this also into the try so that an event is emitted if the schema does not exist.

proposals/lip-0049.md
```diff
@@ -232,15 +236,15 @@ def apply(ccu: Transaction, ccm: CCM) -> None:
         for each module mdl for which verifyCrossChainMessage exists:
             mdl.verifyCrossChainMessage(ccu, ccm)
     except:    
+        terminateChain(ccm.sendingChainID)
```
janhack, line 239: Why do we move this function call before the event emission here and in other places? Note that `terminateChain` sends a message (with associated event) so this event would then be emitted before.

proposals/lip-0043.md
```diff
@@ -311,7 +311,6 @@ def execute(trs: Transaction) -> None:
 
     # Send registration CCM to the sidechain.
```
janhack, line 312: By the way, why do we not use `sendInternal` here?

proposals/lip-0049.md
```diff
@@ -301,19 +305,23 @@ def apply(ccu: Transaction, ccm: CCM) -> None:
...
-        terminateChain(ccm.sendingChainID)  
 ```
 
 #### forward
 
 The `forward` function forwards a CCM to the specified receiving chain. This function is only called on the mainchain when processing a CCU. If an error occurs, the CCM is bounced back to the sending chain or discarded.
 
+<img src="lip-0049/forward.png" width="80%">
+
+_Figure 2: A schematic of the steps executed in the `forward` function. Blue boredered boxes indicate the emission of an event. Function called from other modules during the CCM execution (indicated as `mdl.<FUNCTION_NAME>` are repeated for each module in the chain for which the function exists._
```
janhack, line 323:
```suggestion
_Figure 2: A schematic of the steps executed in the `forward` function. Blue boredered boxes indicate the emission of an event. Functions called from other modules during the CCM execution (indicated as `mdl.<FUNCTION_NAME>`) are repeated for each module in the chain for which the function exists._
```

proposals/lip-0049.md
```diff
@@ -220,6 +220,10 @@ def validateFormat(ccm: CCM) -> None:
 
 The `apply` function applies a cross-chain message, in the context of a cross-chain update `ccu`.
 
+<img src="lip-0049/apply.png" width="80%">
+
+_Figure 1: A schematic of the steps executed in the `apply` function. Blue boredered boxes indicate the emission of an event. Function called from other modules during the CCM execution (indicated as `mdl.<FUNCTION_NAME>` are repeated for each module in the chain for which the function exists._
```
janhack, line 225:
```suggestion
_Figure 1: A schematic of the steps executed in the `apply` function. Blue boredered boxes indicate the emission of an event. Functions called from other modules during the CCM execution (indicated as `mdl.<FUNCTION_NAME>`) are repeated for each module in the chain for which the function exists._
```

janhack (changes requested):
I just added a few small comments and questions.

proposals/lip-0045.md
ricott1: This comes from a discussion we had with the SDK to ensure that it fits into `number` type and we can avoid using `bigint` for the moment. If this becomes a problem in the future, we can upgrade to uint64 and bigints. Maybe @shuse2 has some extra input?

proposals/lip-0049.md
ricott1: I wanted to be consistent with what we do in LIP53 (e.g., here: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#execution). In general, I think both orderings would be fine.

proposals/lip-0053.md
```diff
@@ -534,9 +542,9 @@ def execute(ccu: Transaction) -> None:
             return
 
     for ccmBytes in crossChainMessages:
-        appendToInboxTree(ccu.params.sendingChainID, ccmBytes)
         ccm = decode(crossChainMessageSchema, ccmBytes)
         apply(ccu, ccm)
+        appendToInboxTree(ccu.params.sendingChainID, ccmBytes)
```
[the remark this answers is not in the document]
ricott1: I wanted to be consistent with the execution of the mainchain CCU. Notice that decoding cannot fail at this point, and apply always returns (all exceptions are catched).

proposals/lip-0045.md
shuse2: Current tree takes uint32, and if we change to uint64, we need to update to use `bigint` where we have difference in the performance and extra workload.
That's why we suggested to go with uint32 for now, and update later if needed

proposals/lip-0045.md
ricott1: done

proposals/lip-0045.md
ricott1: It is true that it is somewhat inconsistent with what we have in LIP 0031: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0031.md#proof-serialization

proposals/lip-0053.md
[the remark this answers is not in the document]
janhack: Okay, sounds good.

proposals/lip-0049.md
janhack: Okay, sounds good.

proposals/lip-0045.md
janhack: Okay, if this is better for the implementation, sounds good to me. Not sure if we should update LIP 0031. The comment can be resolved from my side.

janhack (approved):
Looks good!

gkoumout (changes requested):
All good from my side, just caught a couple of typos.

proposals/lip-0043.md
```diff
@@ -596,7 +597,7 @@ def execute(trs: Transaction) -> None:
     ccmID = sha256(encode(crossChainMessageSchema, ccm))
     emitEvent(
         module = MODULE_NAME_INTEROPERABILITY,
-        name = EVENT_NAME_CCM_SEND_SUCCESS,
+        name = EVENT_NAME_CCM_SENT_SUCCESS,
```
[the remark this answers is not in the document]
ricott1: fixed

proposals/lip-0045.md
ricott1: I would leave it as is for the moment then.

gkoumout (approved):
Ready to merge!

gkoumout (approved):

AndreasKendziorra:
I spotted this by chance: it should be `messageWitnessHashes` instead of `siblingHashes` in [this line](https://github.com/LiskHQ/lips/blob/4dfe5ef853277dbae3b6c925fa2e98ca20e6460e/proposals/lip-0053.md?plain=1#L61). Maybe it can be added here.

proposals/lip-0045.md
```diff
@@ -134,6 +134,7 @@ In this section, we specify the substores that are part of the Interoperability
 | `CHAIN_ID_MAINCHAIN` | bytes | 0x00000000 | Chain ID of the Lisk mainchain in the mainnet. |
 | `CHAIN_NAME_MAINCHAIN` | string | "lisk-mainchain" | Name of the Lisk mainchain. |
 | **Interoperability Store** | | | |
+| `STORE_PREFIX_INTEROPERABILITY` | bytes | 0x83ed0d25 | Prefix of the interoperability store. |
```
janhack, line 137: This corresponds to the integer 2213350693. What is the reason for this specific choice?

proposals/lip-0045.md
```diff
@@ -1105,105 +1222,25 @@ The data property follows the `terminatedOutboxAccountSchema` schema given [abov
 
 #### send
 
-The `send` function is used to create and add a message to the outbox of a partner chain.
+The `send` function is used to create and add a message to the outbox of a partner chain. The ccm status is automatically set to `CCM_STATUS_OK`.
```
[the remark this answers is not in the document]
janhack: This issue is re-introduced again with a later PR. The constant should be `CCM_STATUS_CODE_OK` as in the table.

proposals/lip-0045.md
janhack: So we can resolve the comment?

janhack (approved):
I have one comment regarding a small bugfix, otherwise the PR looks good.

I did not check the logic and parameters of the Sidechain Terminated and Channel Terminated message as I do not have the context of LIP 54 and I think it would be best if someone who is also reviewing that LIP checks these things.

proposals/lip-0045.md
ricott1: The rationale is explained in LIP40. We want to separate the interop store from other stores to minimize the witness length in CCUs.

proposals/lip-0045.md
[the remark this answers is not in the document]
gkoumout: I still see it `CCM_STATUS_OK`. Replace it with `CCM_STATUS_CODE_OK` everywhere.

gkoumout (changes requested):
Looks good, just some minor suggestions. 

- In the commands section of LIP45 add the new Livenesss termination command and link to its description in LIP 54. Also, I realized in the list of commands of LIP45, there is one entry "cross chain update command" although in reality we have two such commands in LIP53, the mainchain and the sidechain one. Perhaps it would be good to update this also here. 

- LIP 54 has a minor modification in this PR and also there is another open PR. Be careful for consistent

proposals/lip-0045.md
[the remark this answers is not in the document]
ricott1: fixed :)

proposals/lip-0045.md
ricott1: yes, resolving.

gkoumout (approved):

karmacoma (approved):

PR #177, opened 2022-10-04
Change specifications according to consensus-application separation

janhack:
The BFT module is not implemented as a module any more, but instead moved to the engine. We therefore update the application layer in order to remove any information flow from engine to state machine. This implies the following changes:

## Update of LIP 0044

- Removed the rationale section about block slots and moved it to LIP 0058 instead.
- The Validators module now stores the certificate threshold, precommit threshold and for every validator in a round the address, BFT weight, generator key and BLS key. This way the Validators module can forward this information to the engine and other modules can request this information for the current round if needed.
- The Generator List Substore is renamed to the Validators Params Substore which stores the information mentioned above.
- The function to set and get the generator list is renamed to `setValidatorParams` and `getValidatorParams` and generalized.
- The function `getSlotTime` is removed as it is not used any more.
- The functions `getGeneratorAtTimestamp` and `getSlotNumber` are moved to LIP 0058 as they are used in the engine during the block processing and not used in the state machine any more.
- Remove endpoint `getValidatorParams` as this information will be provided by the engine.

## Update of LIP 0050

- Change Validator method call from `getValidatorAccount` to `getValidatorKeys`. 

## Update of LIP 0055

- We add the property `impliesMaxPrevotes` to the block header. This property will be validated by the engine and can be used by the Rewards module to determine whether the block reward should be reduced. Therefore the call to the method `impliesMaximalPrevotes` is not needed any more.
- Update references to the functions `getGeneratorAtTimestamp` and `getSlotNumber`, which are now defined in LIP 0058.

## Update of LIP 0056

- Update links.
- Remove any reference to "BFT module".

## Update of LIP 0058

- Added the section about block slots that was previously in LIP 0044.
- Removed any reference that the BFT logic is implemented as a module.
- Use terminology "consensus domain" and "application domain" instead of replication layer or state machine.
- Add Genesis Data Substore for storing the genesis block timestamp.
- The function `setBFTParameters` is changed as follows:
    - The array `validators` given as input is not sorted and stored in the BFT Parameters store in unchanged order. This array determines the block generation list.
    - The BFT weight of the validators is allowed to be 0.
    - Any calls to the Validators module can be removed as the `validators` array already contains the generator and BLS keys. 
- Fix function `impliesMaximalPrevotes` to work for the first block after the genesis block.
- Remove function `getCurrentValidators` as it is not needed any more.
- Add function `getSlotNumber` (moved from LIP 0044)
- Add function `getGeneratorAtTimestamp` (moved from LIP 0044)

## Update of LIP 0059

- Use the property `b.header.aggregateCommit.height` instead of requesting the current value of `maxHeightCertified` from the BFT store.

## Update of LIP 0060

- Add `b.header.impliesMaxPrevotes == True` to static validation.
- Add `eventRoot` property to result verification stage.
- Change `assetsRoot` to `assetRoot`.


proposals/lip-0044.md
```diff
@@ -127,29 +130,76 @@ The validator account holds the generator and BLS keys of a registered validator
...
 
 ##### Substore prefix, Store Key, and Store Value
 
-* The substore prefix is set to `SUBSTORE_PREFIX_GENERATOR_LIST`.
+* The substore prefix is set to `SUBSTORE_PREFIX_VALIDATORS_PARAMS`.
 * The store key is set to empty bytes.
-* The store value is set to the _generator list_ data structure, serialized using the JSON schema `generatorListSchema`, presented [below](#json-schema-1).
-* Notation: For the rest of this proposal let `generatorList` be the entry in the generator list substore (for notational convenience, we refer directly to the array of addresses, without explicitly writing `generatorList.addresses`).
+* The store value is the serialization of an object following the JSON schema `validatorParamsSchema` defined below.
+* Notation: For the rest of this proposal let `validatorsParams` be the entry in the generator list substore.
```
ricott1, line 140: "... in the validator params substore."

proposals/lip-0044.md
```diff
@@ -596,29 +646,71 @@ def getGeneratorsBetweenTimestamps(startTimestamp: uint32, endTimestamp: uint32)
...
+
+##### Execution
 
 ```python
-def setGeneratorList(generatorAddresses: list[Address]) -> None:
-    for address in generatorAddresses:
-        if no entry in the validators data substore exist with storeKey == address:
-            raise Exception(f'Invalid address in generator list. The address {address.hex()} is not registered as validator.')
-    generatorList = generatorAddresses
+setValidatorParams(precommitThreshold: uint64, certificateThreshold: uint64, validatorList: list[object]):
```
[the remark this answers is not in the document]
janhack: I tried to put the function from the BFT module here 1-1 and also store exactly the same information as in the BFT Params store, but of course we could change it. I was thinking it would be nice to do the validation here so if in the application layer this method is called with invalid parameters, an exception is raised directly. Additionally, I though it is nice to have all consensus-related properties here in one place as they can then be also requested via an endpoint. However, in terms of protocol it would not be necessary to store them. I am open to suggestions here.

proposals/lip-0055.md
```diff
@@ -599,16 +609,22 @@ def verifyEventRoot():
...
 ```python
 def verifyBFTProperties():
   if block.header.maxHeightPrevoted != bft.getBFTHeights().maxHeightPrevoted:
     return False
 
-  return not bft.isHeaderContradictingChain(block.header)
+  if bft.isHeaderContradictingChain(block.header):
+    return False
+
+  if block.header.impliesMaxPrevotes != bft.impliesMaximalPrevotes(block.header)
```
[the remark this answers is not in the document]
janhack: Isn't the whole function `verifyBFTProperties()` part of the engine now so all information would be available?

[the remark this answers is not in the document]
shuse2: yes, all the information is available. I guess I just need to slightly modify the `impliesMaximalPrevotes` function to handle genesis block case, but it can be handled.

proposals/lip-0044.md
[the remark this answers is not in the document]
shuse2: I think it would be nicer if we can avoid validators module to know too much about BFT protocol, so I would prefer to keep it as proxy to reply to the engine which stores the value as it is.

Validation wise it will be the same in most cases. Even if we don't throw in the application, it will be thrown in the engine and the block will be invalid.
Maybe only case where it could be different is if there is integration test without engine, developer will not notice, but I think that's probably true for other replies for engine (like invalid event schema etc), also this can be resolved by mock ABI implementation.

Currently, there is `consensus` endpoint exposed from the engine, so BFT related properties can be fetched from the endpoint (for certificate etc)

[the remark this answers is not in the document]
janhack: Okay, then in terms of specifications, the function `setValidatorParams` is a simple getter without validation, but in the BFT LIP I keep the full function with full validation. I will also remove the `prevoteThreshold` and `validatorsHash` property from the store.

proposals/lip-0044.md
janhack: Fixed.

proposals/lip-0044.md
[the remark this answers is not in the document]
janhack: This comment can be resolved from my side.

proposals/lip-0044.md
```diff
@@ -127,32 +124,69 @@ The validator account holds the generator and BLS keys of a registered validator
...
+            "dataType": "bytes",
+            "length": ADDRESS_LENGTH,
+            "fieldNumber": 1
+        },
+        "bftWeight": {
+            "dataType": "uint64",
+            "fieldNumber": 2
+        },
+        "generatorKey": {
+            "dataType": "uint64",
```
ricott1, line 179:
```suggestion
            "dataType": "bytes",
```

proposals/lip-0059.md
```diff
@@ -41,16 +41,16 @@ lastHeightOfRound(h + 2 * ROUND_LENGTH) <= maxHeightCertified.
...
 - `maxHeightCertified` is the maximum height of a block for which the chain contains a certificate.
 
 ## Specification
 
 ### Notation
 
-- We use the constant `ROUND_LENGTH` defined in the [LIP "Define state and state transitions of DPoS module"][research:dpos-module] to denote the round length in the DPoS chain, i.e., the number of active plus randomly selected standby delegates.
-- We write `bftModule.fct` for a call to the function `fct` defined in the [BFT module][research:bft-module].
+- We use the constant `ROUND_LENGTH` defined in [LIP 0057][lip-0057] to denote the round length in the DPoS chain, i.e., the number of active plus randomly selected standby delegates.
+- We write `bftModule.fct` for a call to the function `fct` defined in the [BFT module][lip-0058].
```
ricott1, line 53: We should not refer to it as a module, right?

proposals/lip-0058.md
```diff
@@ -626,26 +674,50 @@ The function allows to obtain the next larger key in the BFT Parameters substore
...
+    for val in bft.validators:
+        if val.address = address:
+            return val
+    raise Exception()
+```
+
 #### setBFTParameters
 
-This function allows to set the thresholds, validators and associated BFT weights to be used from the next height onward and updates the BFT module store accordingly.
+This function allows to set the thresholds, validators and associated BFT weights to be used from the next height onward and updates the BFT module store accordingly. The function should be called with the parameters forwarded to the application layer from the state machine every time the function [setValidatorParams][lip-0044#setValidatorParams] of the Validators module is called.
```
ricott1, line 705:
```suggestion
This function allows to set the thresholds, validators and associated BFT weights to be used from the next height onward and updates the BFT store accordingly. The function should be called with the parameters forwarded to the application layer from the state machine every time the function [setValidatorParams][lip-0044#setValidatorParams] of the Validators module is called.
```

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+The Lisk-BFT consensus protocol introduced in [LIP 0014](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0014.md) specifies two new block properties, `maxHeightGenerated` (renamed from `maxHeightPreviouslyForged`) and `maxHeightPrevoted`, which are validated as part of the block processing. As part of the [new state architecture][lip-0069] and stricter separation between state machine layer and replication layer, the majority of the block processing now happens in the state machine layer and is logically separated into separate modules. The consensus-related logic and verification of the block header properties is performed in the replication layer, however. That is why the consensus participants, their BFT weights, keys and the current state of consensus votes is also maintained as part of the replication layer. In this LIP we therefore define the BFT store for this purpose and specify how this information is updated as part of the block processing. Additionally, we define the concept of block slots and the logic for determining which validator is assigned to a specific block slot.
```
ricott1, line 24: I would also mention the property `impliesMaximalPrevotes` (although it is not introduced in LIP14), since it is handled by the BFT store

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+We call the first four inputs above the **BFT parameters**. We assume that these BFT parameters are forwarded from the state machine layer to the replication layer whenever the [setValidatorParams][lip-0044#setValidatorParams] function in the Validators module of the state machine layer is called. For a DPoS chain, the DPoS module would need to call this function with the correct parameters, in particular, notifying the replication layer about any changes of validators. Similarly, for a PoA chain, it is the responsibility of the PoA module to provide the correct parameters as input to `setValidatorParams` function. The block header information can be obtained during the block processing in the replication layer.
```
ricott1, line 42: I have some problems parsing the last sentence. I would find it clearer to just say smtg like: "The block header information is always available to the replication layer." or even be more specific about what is considered the 'engine-state'.

proposals/lip-0058.md
```diff
@@ -268,9 +292,29 @@ bftVotesSchema = {
...
-### Commands
+#### Genesis Data Substore
 
-The BFT module does not define any commands.
+##### Substore prefix, Store Key, and Store Value
+
+* The substore prefix is set to `SUBSTORE_PREFIX_GENESIS_DATA`.
+* The store key is set to empty bytes.
+* The store value is set to a data structure holding the timestamp of the genesis block of the chain, serialized using the `genesisDataSchema` schema presented below.
+* Notation: For the rest of this proposal let `genesisData` be the entry in the genesis data substore.
```
ricott1, line 302: you are a bit inconsistent with the capitalization of substore names in the LIP

proposals/lip-0058.md
```diff
@@ -401,7 +445,7 @@ The BFT parameters valid at the height `h`.
 ##### Execution
 
 ```python
-getBFTParametersInternal(h):
+def getBFTParametersInternal(h):
```
ricott1, line 448: Here and later, I would still add the typing

proposals/lip-0058.md
```diff
@@ -487,7 +531,7 @@ The function returns `True` if and only if there is an entry in the BFT Paramete
 ##### Execution
 
 ```python
-existBFTParameters(h):
+def existBFTParameters(h):
```
ricott1, line 534: Does `h` represent an height here? what type is it? if it is uint we still need to convert it internally

proposals/lip-0058.md
```diff
@@ -487,7 +531,7 @@ The function returns `True` if and only if there is an entry in the BFT Paramete
...
     if exists entry in BFT Parameters store with key h:
```
ricott1, line 535:
```suggestion
    if exists entry in BFT Parameters store with store key h:
```

proposals/lip-0058.md
```diff
@@ -694,29 +765,28 @@ setBFTParameters(precommitThreshold, certificateThreshold, validators):
...
   * `bftVotes.activeValidatorsVoteInfo` is an empty array.
+* The Genesis Data substore is initialized by setting `genesisData.timestamp = h.header.timestamp`.
 
 #### Genesis State Finalization
 
-In this step the BFT module does not execute any logic. However, the exposed function `setBFTParameters` is expected to be called by the PoA module or DPoS module as part of the genesis state finalization of these modules.
+In this step, the function [setValidatorParams][lip-0044#setValidatorParams] of the Validators module should be called by the PoA module or DPoS module.
+These parameters are forwarded to the replication layer and the BFT parameters store is updated accordingly using the function `setBFTParameters` defined above.
 
 ### Block Processing
```
ricott1, line 791: I think this section should be removed. As this is not a module anymore, the state machine hooks would not be called. If the logic is still processed, we should clarify when that happens (probably in the replication layer).

proposals/lip-0058.md
```diff
@@ -736,21 +806,23 @@ Before the transactions of a block `b` are executed, the following logic is exec
 
 ### Block Creation
```
ricott1, line 807: Also here, it is not clear to me when would this logic be called.

ricott1 (changes requested):
In general, it looks good. I still have some comments, in particular related to the (ex)BFT module hooks.

proposals/lip-0044.md
```diff
@@ -127,32 +124,69 @@ The validator account holds the generator and BLS keys of a registered validator
...
 
 ##### Substore prefix, Store Key, and Store Value
 
-* The substore prefix is set to `SUBSTORE_PREFIX_GENERATOR_LIST`.
+* The substore prefix is set to `SUBSTORE_PREFIX_VALIDATORS_PARAMS`.
 * The store key is set to empty bytes.
-* The store value is set to the _generator list_ data structure, serialized using the JSON schema `generatorListSchema`, presented [below](#json-schema-1).
-* Notation: For the rest of this proposal let `generatorList` be the entry in the generator list substore (for notational convenience, we refer directly to the array of addresses, without explicitly writing `generatorList.addresses`).
+* The store value is the serialization of an object following the JSON schema `validatorParamsSchema` defined below.
+* Notation: For the rest of this proposal let `validatorsParams` be the entry in the validators params substore.
```
AndreasKendziorra, line 134: sometimes it's "validatorsParams" and sometimes "validatorParams"

proposals/lip-0059.md
```diff
@@ -61,33 +61,33 @@ Here
 ```python
 lastHeightOfRound(h):
     r = roundNumber(h)
-    return r * ROUND_LENGTH + genesisHeight
+    return r * ROUND_LENGTH + genesisDataStore.height
 ```
 
-Here `roundNumber` is a function defined in the [LIP "Define state and state transitions of DPoS module"][research:dpos-module#roundnumber] and `genesisHeight` denotes the height of the genesis block as stored in the [DPoS module][research:dpos-module].
+Here `roundNumber` is a function defined in [LIP 0057][lip-0057#roundnumber] and `genesisDataStore.height` denotes the height of the genesis block as stored in the [DPoS module][lip-0057].
```
AndreasKendziorra, line 67: > `roundNumber` is a function defined in [LIP 0057]

this is maybe a bit misleading because this function is not public. So maybe we should define this internal function here as well or just replace `roundNumber(h)` by the formula.

AndreasKendziorra (changes requested):
The term "BFT module" is still used in the LIPs 0042, 0047, 0056, 0057, 0061

proposals/lip-0044.md
janhack: Fixed.

proposals/lip-0059.md
janhack: Fixed, I removed this bullet point.

proposals/lip-0058.md
janhack: Fixed.

proposals/lip-0058.md
janhack: Changed to "Genesis Data substore" here and in other places in LIP 58 for consistency.

proposals/lip-0058.md
janhack: The input parameters is described in the section Parameters above. Changed the line below to
```
if exists entry in BFT Parameters store with key uint32be(h):
```

proposals/lip-0058.md
janhack: Fixed.

proposals/lip-0058.md
janhack: I would prefer not to do it as I wanted to restrict the PR to necessary specification changes, we have all the "Parameters" section which already describe the input parameters, it would not help to save additional validations and there are also a lot of complex objects as input parameters. I think we also said that introducing type hinting for published LIPs is optional.

proposals/lip-0058.md
janhack: As concluded in our discussions, I now have a section "Before Application Processing" and "After Application Processing" for the genesis block processing.

proposals/lip-0058.md
janhack: I improved this section and reference the concrete block generation stage defined in LIP 55. Note that as far as I understand both the header initialization and header finalization stage are in the engine domain.

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: Updated to

> The function is called in the after application processing stage of the genesis block processing and general block processing as defined below.

In this LIP I use the more general academic terminology (state machine, replication layer) instead of the concrete Lisk SDK architecture terminology (application domain, engine domain). Not sure if this is the best approach, I will raise this point on Slack.

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: Fixed both points. I would keep "logic" as not only verification is performed in the replication layer.

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: As concluded on Slack, I now only use the terms application domain and consensus domain as introduced in LIP 69.

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: - I removed the part "notifying the replication layer about any changes of validators".
- Changed to "input to the setValidatorParams function" as suggested.
- As concluded on Slack, I now only use the terms application domain and consensus domain as introduced in LIP 69.
- The DPoS module should already be updated on LIPs staging. For the PoA module, I would suggest that Sergey includes this change as he is already a PR for updating the module. I will send a message to Sergey about this point.

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
-* Functions that allow to verify the Lisk-BFT related properties in the block header, namely `validatorsHash`, `maxHeightGenerated` and `maxHeightPrevoted`, see also [LIP 0055][lip-0055].
-* Functions for other modules to obtain information regarding the current state of the BFT module. For instance, the DPoS module requires the value of `maxHeightCertified` for the [unlock transaction][lip-0059].
+* Functions that allow to verify the Lisk-BFT related properties in the block header, namely `validatorsHash`, `maxHeightGenerated`, `maxHeightPrevoted` and `impliesMaximalPrevotes`, see also [LIP 0055][lip-0055].
+
+### Block Slots
+
+In the Lisk protocol, time is divided into intervals of fixed length. These intervals are called _block slots_ and at most one block can be added to the blockchain during each slot. The length of a block slot is a constant called _block time_. In this LIP we define how to determine whether which generator is eligible to generate a block in a certain block slot.
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+| `SUBSTORE_PREFIX_GENESIS_DATA`| bytes | `0x8000` | Substore prefix of the genesis data substore. |
 | `LSK_BFT_BATCH_SIZE`          |`uint32`  | configurable per chain | This constant determines the value of `MAX_LENGTH_BLOCK_BFT_INFOS` and thereby the range of prevotes and precommits. Additionally, it is used in the block synchronization mechanism and fast chain switching mechanism defined in [LIP 0014](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0014.md). |
 | `MAX_LENGTH_BLOCK_BFT_INFOS`  |`uint32`  | `3*LSK_BFT_BATCH_SIZE` | The maximum length of the array `blockBFTInfos` in the BFT Votes substore. |
+| `BLOCK_TIME`                  |`uint32`  | 10 (value for the Lisk mainchain) | Block time (in seconds) set in the chain configuration.         |
+| `ADDRESS_LENGTH`              |`uint32`  | 20                     | Length in bytes of an address.                                             |
+| `ED25519_PUBLIC_KEY_LENGTH`   |`uint32`  | 32                     | Length in bytes of a Ed25519 public keys.                                  |
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+| `BLS_PUBLIC_KEY_LENGTH`       |`uint32`  | 48                     | Length in bytes of BLS public key.                                         |
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0044.md
```diff
@@ -594,32 +595,45 @@ def getGeneratorsBetweenTimestamps(startTimestamp: uint32, endTimestamp: uint32)
...
+        newValidator.address =  validator.address
+        newValidator.bftWeight = validator.bftWeight
+        validatorKeys = getValidatorKeys(validator.address)
+        newValidator.blsKey =  validatorKeys.blsKey
+        newValidator.generatorKey =  validatorKeys.generatorKey
+        newValidators.append(newValidator)
+    params.validators = newValidators
+
+    # write new validator params to state store
+    validatorParams = params
```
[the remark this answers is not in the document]
janhack: My suggestion would then be to use "validatorsParamsStore" for the state store notation, i.e., always have "store" as part of the name.

proposals/lip-0058.md
```diff
@@ -74,18 +82,18 @@ As explained in [Section "Comparison to the Lisk-BFT paper" in LIP 0056][lip-005
...
 
-### BFT Module Store
+### BFT Store
 
-The BFT module store is organized in the two substores described below.
+The BFT store is organized in the three substores described below. Note that the BFT Store is not part of the state store of the state machine and therefore does not impact the state root.
 
 #### BFT Parameters
 
 ##### Store Prefix, Store Key, and Store Value
```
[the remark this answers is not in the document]
janhack: - I added the sentence "Each substore of the BFT store is a simple key-value store similar to the substores in the application domain."
- I agree that we don't necessarily need these exact substore prefixes, but we would still need different prefixes for different substores. I think it is also fine to define them in the LIP although the protocol would be the same. @shuse2 Can you give more details on how the substore prefixes are used?

proposals/lip-0058.md
```diff
@@ -136,6 +133,33 @@ bftParametersSchema = {
...
+            "dataType": "bytes",
+            "length": ADDRESS_LENGTH,
+            "fieldNumber": 1
+        },
+        "bftWeight": {
+            "dataType": "uint64",
+            "fieldNumber": 2
+        },
+        "generatorKey": {
+            "dataType": "uint64",
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0058.md
```diff
@@ -136,6 +133,33 @@ bftParametersSchema = {
...
+            "length": ED25519_PUBLIC_KEY_LENGTH,
+            "fieldNumber": 3
+        },
+        "blsKey": {
+            "dataType": "uint64",
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0058.md
```diff
@@ -147,17 +171,17 @@ All properties below are valid starting from the height given as key:
...
+* `validators`: Each element in the array `validators` corresponds to a validator and stores its address, generator key, BLS key and BFT weight property. The order in the array determines the block generation order.
 * `validatorsHash`: A 32-byte value authenticating the BLS keys and BFT weights of the validators and the certificate threshold of the same store entry.
 
 #### BFT Votes
 
 ##### Store Prefix, Store Key, and Store Value
 
-* The store prefix is `STORE_PREFIX_BFT_VOTES`.
+* The substore prefix is `SUBSTORE_PREFIX_BFT_VOTES`.
 * The store key is empty bytes.
```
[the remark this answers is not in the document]
janhack: As far as I know we only have one persistent storage that we use in the Lisk SDK, the RocksDB database which is basically one giant key-value store. I think for consistency it makes sense to also store this information there and to distinguish it from other substores, it makes sense to have a certain prefix.

proposals/lip-0058.md
```diff
@@ -546,7 +596,7 @@ The function returns `True` if the block header `blockHeader` provided as input
 ##### Execution
```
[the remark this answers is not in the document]
janhack: Changed to "Update ... to the new _store_ structure".

proposals/lip-0044.md
```diff
@@ -594,32 +595,45 @@ def getGeneratorsBetweenTimestamps(startTimestamp: uint32, endTimestamp: uint32)
     return result
 ```
 
-#### setGeneratorList
+#### setValidatorParams
 
-This function updates the value of the generator list to the input `generatorAddresses`. It checks if all elements of `generatorAddresses` have a registered validator account in the validators data substore. If this is not the case, it throws an error. Otherwise, it updates the value of `generatorList`.
+This function allows to set the precommit threshold, certificate threshold, validators and associated BFT weights to be used from the next height onward. The information is not validated and simply stored in the Validators module. The information is further forwarded to the engine where all validations are performed.
```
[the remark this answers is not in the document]
janhack: I added the following line to the function:
```
forward params to consensus domain
```

What happens in the engine if the validation fails, is now defined in the section "After Application Processing" in LIP 58.

proposals/lip-0058.md
```diff
@@ -626,26 +674,50 @@ The function allows to obtain the next larger key in the BFT Parameters substore
...
+
+```python
+def getSlotNumber(timestamp):
+    if timestamp < genesisData.timestamp:
+        raise Exception('Input timestamp cannot be smaller than genesis block timestamp.')
+    elapsedTime = timestamp - genesisData.timestamp
+    return floor(elapsedTime / BLOCK_TIME)
+```
+
+#### getValidator
```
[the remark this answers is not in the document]
janhack: I removed the function.

proposals/lip-0058.md
```diff
@@ -694,29 +765,28 @@ setBFTParameters(precommitThreshold, certificateThreshold, validators):
...
 * The BFT Votes substore is initialized as follows:
-  * `bftVotes.maxHeightPrevoted = b.height`,
-  * `bftVotes.maxHeightPrecommitted = b.height`,
-  * `bftVotes.maxHeightCertified = b.height`,
+  * `bftVotes.maxHeightPrevoted = b.header.height`,
+  * `bftVotes.maxHeightPrecommitted = b.header.height`,
+  * `bftVotes.maxHeightCertified = b.header.height`,
   * `bftVotes.blockBFTInfos` is an empty array,
   * `bftVotes.activeValidatorsVoteInfo` is an empty array.
+* The Genesis Data substore is initialized by setting `genesisData.timestamp = h.header.timestamp`.
```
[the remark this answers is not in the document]
janhack: Fixed.

proposals/lip-0044.md
```diff
@@ -511,46 +545,13 @@ def getAddressFromBLSKey(blsKey: PublicKeyBLS) -> Address:
...
-
-This function returns the current generator list `generatorList`.
-
-```python
-def getGeneratorList() -> list[Address]:
-    return generatorList
-```
-
-#### getSlotNumber
+#### getValidatorParams
```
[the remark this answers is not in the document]
janhack: This is introduced so that other modules have the flexibility to obtain this information from the Validators module. However, for our modules we don't need it so far.

proposals/lip-0044.md
```diff
@@ -395,15 +429,15 @@ def registerValidatorWithoutBLSKey(validatorAddress: Address, generatorKey: Publ
     )
 ```
 
-#### getValidatorAccount
+#### getValidatorKeys
```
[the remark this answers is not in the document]
janhack: I fixed the 1 occurrence in LIP 55 and the 1 occurrence in LIP 50. For all other LIPs there is an in-progress LIP update and I added a point on the Trello card "Follow up tasks after engine-application separation".

proposals/lip-0044.md
janhack: It should now be "validatorParams" everywhere.

proposals/lip-0059.md
janhack: This LIP defines some additional functionality of the DPoS module and would therefore be implemented as part of the DPoS module. Therefore, I assume here that I can use internal functions of the DPoS module.

proposals/lip-0058.md
```diff
@@ -591,27 +641,25 @@ The BFT parameters valid at the height `h`.
...
-##### Returns
-
-An array of objects, each with a property `address` of type bytes and a property `bftWeight` of type 64-bit unsigned integer.
-
-##### Execution
+This function returns the address of the generator active at the input timestamp. Notice that if the input timestamp corresponds to a time slot before or after the current round, this function may not return the correct generator address.
 
 ```python
-getCurrentValidators()
+def getGeneratorAtTimestamp(timestamp):
```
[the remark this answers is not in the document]
janhack: I added an additional input parameter "currentHeight" and updated the function description.

proposals/lip-0058.md
```diff
@@ -655,15 +727,14 @@ setBFTParameters(precommitThreshold, certificateThreshold, validators):
...
 
     # Compute the validators hash
     activeValidators = []
     for validator in validators:
         activeValidator = object with property bftWeight and blsKey
         activeValidator.bftWeight = validator.bftWeight
-        activeValidator.blsKey =  validatorModule.getValidatorAccount(validator.address).blsKey
+        activeValidator.blsKey =  validator.blsKey
         add activeValidator to activeValidators
     sort elements in activeValidators lexicographically by blsKey property
```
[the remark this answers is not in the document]
janhack: Very good point. I added the check.

janhack:
> Should we also update LIP0042 (reward) to use the block heade impliesMaximalPrevotes property in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0042.md#execution ?

@ricott1 I also fixed this point.

> The term "BFT module" is still used in the LIPs 0042, 0047, 0056, 0057, 0061

@AndreasKendziorra I removed the references in LIP 42 and 56. For the LIPs 47, 57 and 61 there is a pending update so I did not change it here, but instead added these points to a Trello card.

janhack:
I addressed all review comments and re-requested your reviews.

proposals/lip-0044.md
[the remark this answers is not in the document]
AndreasKendziorra: ok. It would be good to reference the sections from LIP 58 as it's very hard for a reader of this section to see where in LIP 58 this data is processed /validated.
Maybe it's also not very obvious that the consensus domain has to keep the forwarded data in memory until it a certain stage (after application processing) is reached and will not process/validate it directly.

One idea: LIP 0058 introduces an auxiliary data structure `aux_BFT_params` that is kept in memory and that is filled by a function `set_aux_BFT_params` which is called by `setValidatorParams`. Then in  After Application Processing `setBFTParameters`is called with `aux_BFT_params` if `aux_BFT_params` not empty. At the end of After Application Processing is cleared such that it is empty once the next block is processed.

proposals/lip-0058.md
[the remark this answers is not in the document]
AndreasKendziorra: Fine for me. I would leave it to @shuse2 if more changes are needed here.

proposals/lip-0044.md
```diff
@@ -127,32 +124,69 @@ The validator account holds the generator and BLS keys of a registered validator
...
+        "blsKey": {
+            "dataType": "uint64",
+            "length": BLS_PUBLIC_KEY_LENGTH,
+            "fieldNumber": 4
+        }
+    }
+}
 ```
 
 ##### Properties and Default values
```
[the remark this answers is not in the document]
AndreasKendziorra: this was overlooked?

AndreasKendziorra (changes requested):

proposals/lip-0042.md
```diff
@@ -206,6 +206,5 @@ TBA
 [creative]: https://creativecommons.org/publicdomain/zero/1.0/
 [block-schema-lip]: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md
 [bft-module]: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md
```
ricott1, line 208: I think this link can be removed

proposals/lip-0058.md
ricott1: sounds good.

ricott1 (changes requested):
Iresolved most of my comments, an extra one:

- We still need to add `impliesMaxPrevotes` to the section "New Block Header Properties" of LIP0055.

proposals/lip-0060.md
```diff
@@ -41,17 +41,21 @@ Another distinctive specification of the genesis block is its processing. The bl
```
ricott1, line 41: For consistency we could also just name this section "Static validation"

ricott1: Also, in the previous paragraph we should amend "In particular the processing of the genesis block happens in **four** steps described in the sections below." --> "In particular the processing of the genesis block happens in **five** steps described in the sections below." (not sure if 'block storage should be included, if so, it should be 6)

proposals/lip-0058.md
ricott1: I think you still have inconsistencies in the LIP

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
 ## Specification
 
 ### Constants
 
 | **Name**                      | **Type** | **Value**              | **Description**                                                            |
 |-------------------------------|----------|------------------------|----------------------------------------------------------------------------|
-| `MODULE_ID_BFT`               |`uint32`  | TBD                    | The module ID of the BFT module.                                           |
-| `STORE_PREFIX_BFT_PARAMETERS` |`bytes`   | `0x0000`               | The store prefix of the BFT Parameters substore.                           |
-| `STORE_PREFIX_BFT_VOTES`      |`bytes`   | `0x8000`               | The store prefix of the BFT Votes substore.                                |
+| `SUBSTORE_PREFIX_BFT_PARAMETERS` |`bytes`   | `0x0000`               | The store prefix of the BFT Parameters substore.  |
```
ricott1, line 67: It should be 'substore prefix'

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+| `SUBSTORE_PREFIX_BFT_VOTES`   |`bytes`   | `0x4000`               | The store prefix of the BFT Votes substore.                                |
```
ricott1, line 68: It should be 'substore prefix'

proposals/lip-0058.md
```diff
@@ -13,58 +13,66 @@ Requires: 0055, 0056, 0061
...
+| `SUBSTORE_PREFIX_GENESIS_DATA`| bytes | `0x8000` | Substore prefix of the Genesis Data substore. |
```
ricott1, line 69:
```suggestion
| `SUBSTORE_PREFIX_GENESIS_DATA`| bytes | `0x8000` |  The substore prefix of the Genesis Data substore. |
```

proposals/lip-0058.md
ricott1: I think we can still improve by referencing exactly in which stage this happens.

proposals/lip-0058.md
ricott1: I was looking at the outdated version, resolving.

proposals/lip-0058.md
ricott1: same problem, now it's fine, resolving.

proposals/lip-0044.md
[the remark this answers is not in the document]
janhack: I added some links to the function description to make it more clear what happens.

I would not add any auxiliary data structure for temporary storage or get very detailed regarding the implementation. @shuse2  was explaining that the transfer of information basically works as follows:
```
const result = abi.afterTransactionsExecute();
```
If I understand correctly here the application is called from the consensus domain via the application-blockchain-interface. What "forward params to consensus domain" is to add the BFT parameters to the return value of the function `afterTransactionsExecute`. The `result` value is then used in the after application processing stage within the consensus domain.

proposals/lip-0044.md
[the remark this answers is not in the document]
janhack: Sorry, missed it. I fixed this section and moved the reference to `getGeneratorAtTimestamp` to the parameter description of the function `setValidatorParams`.

proposals/lip-0044.md
```diff
@@ -124,14 +124,14 @@ The validator account holds the generator and BLS keys of a registered validator
 * `generatorKey`: The public key whose corresponding private key is used to sign blocks generated by the validator.
 * `blsKey`: The validator BLS key is the public BLS key whose corresponding private key is used to sign certificates.
 
-#### Validators Params Substore
+#### Validator Params Substore
 
 ##### Substore prefix, Store Key, and Store Value
 
 * The substore prefix is set to `SUBSTORE_PREFIX_VALIDATORS_PARAMS`.
```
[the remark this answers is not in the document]
janhack: Good point, fixed.

proposals/lip-0058.md
```diff
@@ -21,7 +21,7 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
...ain/proposals/lip-0014.md) specifies two block properties, `maxHeightGenerated` (renamed from `maxHeightPreviouslyForged`) and `maxHeightPrevoted`, which are validated as part of the block processing. As part of the [new state architecture][lip-0069] and stricter separation into network domain, consensus domain and application domain, the majority of the block processing now happens application domain and is logically separated into separate modules. The consensus-related logic and verification of the block header properties are performed in the consensus domain, however. That is why the consensus participants, their BFT weights, keys and the current state of consensus votes are also maintained as part of the consensus domain. In this LIP, we therefore define the BFT store for this purpose and specify how this information is updated as part of the block processing. Additionally, we define the concept of block slots and the logic for determining which validator is assigned to a specific block slot. Further, we specify how the new block header properties `impliesMaxPrevotes`, `validatorsHash` and `aggregateCommit` are validated as part of the block processing in the consensus domain.
```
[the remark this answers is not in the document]
janhack: Fixed, thanks.

proposals/lip-0042.md
janhack: Done.

proposals/lip-0058.md
janhack: Fixed.

proposals/lip-0058.md
janhack: Fixed.

proposals/lip-0058.md
janhack: Changed as suggested.

proposals/lip-0058.md
```diff
@@ -681,28 +694,16 @@ This function returns the slot number corresponding to the input timestamp.
...
-```python
-def getValidator(address, height):
-    bftParams = getBFTParameters(height)
-    for val in bft.validators:
-        if val.address = address:
-            return val
-    raise Exception()
-```
-
 #### setBFTParameters
```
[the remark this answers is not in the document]
janhack: I would agree and moved the function to the section "Internal Functions". @shuse2 I hope this is correct.

proposals/lip-0058.md
```diff
@@ -561,7 +578,7 @@ def impliesMaximalPrevotes(blockHeader):
             return False
     heightCurrentTip = bftVotes.blockBFTInfos[0].height
```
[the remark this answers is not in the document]
janhack: I would not do more refactoring now to also keep this diff for this PR small and easier to implement for the developers.

proposals/lip-0060.md
```diff
@@ -70,7 +74,7 @@ As introduced in [the Rationale section](#processing-of-the-genesis-block), the
 * _Static validation of the genesis block_: The stateless checks that ensure the structure of the genesis block are performed. Also, the properties in the block header that do not require access to the state store are checked. These checks are defined in the sections below.
 * _Genesis state initialization_: The genesis state initialization logic for the registered modules is executed.
```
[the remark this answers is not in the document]
janhack: Good catch, I added it.

proposals/lip-0060.md
janhack: - I changed all occurrences of "Static validation of the genesis block" to "Static validation".
- Changed it from four to five steps.

ricott1: I would add the 'block storage' step to the description (you can copy/paste it from LIP 0055), since I put it in the diagram, and update from 'five' to 'six'.

proposals/lip-0058.md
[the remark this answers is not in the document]
AndreasKendziorra: can be resolved from my side, but I leave it open for @shuse2 .

proposals/lip-0058.md
```diff
@@ -268,9 +292,29 @@ bftVotesSchema = {
   * `minActiveHeight`: The height from which the validator has been continuously active.
   * `largestHeightPrecommit`: The largest height for which a block generated by the validator implied a precommit.
 
-### Commands
+#### Genesis Data Substore
```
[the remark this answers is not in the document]
shuse2: I think it's fine to introduce it as notation and function.
Currently, genesis block is saved as any other blocks. At the beginning of starting the node, it checks if the genesis block exists and match with the genesis block given through the config.

there the genesis block header is passed to slot calculation object (Slot) and maintained in the consensus and used for syncing and forkchoice rule

proposals/lip-0058.md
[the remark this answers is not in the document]
shuse2: prefix are used to split the data in the database, so i think it's good to keep it there.
how it's currently structured is

`[STATE_PREFIX, SUB_STORE_PREFIX, KEY]` similar to state db in the application

proposals/lip-0058.md
```diff
@@ -664,7 +726,7 @@ def getBFTParameters(h):
 
 #### getGeneratorAtTimestamp
 
-This function returns the address of the generator active at the timestamp `timestamp` for the block at height `currentHeight` where `currentHeight` where `timestamp` and `currentHeight` are given as input. Notice that if the input timestamp corresponds to a time slot before or after the current round, this function may not return the correct generator address.
+This function returns the address of the generator active at the timestamp `timestamp` for the block at height `currentHeight + 1` where `currentHeight` where `timestamp` and `currentHeight` are given as input. Notice that if the input timestamp corresponds to a time slot before or after the current round, this function may not return the correct generator address.
```
AndreasKendziorra, line 729:
```suggestion
This function returns the address of the generator active at the timestamp `timestamp` for the block at height `currentHeight + 1` where `timestamp` and `currentHeight` are given as input. Notice that if the input timestamp corresponds to a time slot before or after the current round, this function may not return the correct generator address.
```

proposals/lip-0060.md
```diff
@@ -31,13 +31,13 @@ The new genesis block format and processing is specified with the rationale of h
...
+Another distinctive specification of the genesis block is its processing. The blockchain starts with an empty key-value store and in particular, all module stores are initially empty. As part of the genesis block processing modules can add key-value entries to their module store to initialize their state. They should also check the consistency of their state independently and against information provided by other modules. Also, before this is done, the framework layer has to validate the block header and the block format itself. Hence, this LIP defines steps specific to the processing of the genesis block: a step to initialize the state store per each module and the second step to verify this initial state. In particular the processing of the genesis block happens in five steps described in the sections below.
 
 <img src="lip-0060/stages.png" width="70%">
```
AndreasKendziorra, line 36: @ricott1 the legend is broken: no text next to the red square

proposals/lip-0060.md
```diff
@@ -31,13 +31,13 @@ The new genesis block format and processing is specified with the rationale of h
...
+Another distinctive specification of the genesis block is its processing. The blockchain starts with an empty key-value store and in particular, all module stores are initially empty. As part of the genesis block processing modules can add key-value entries to their module store to initialize their state. They should also check the consistency of their state independently and against information provided by other modules. Also, before this is done, the framework layer has to validate the block header and the block format itself. Hence, this LIP defines steps specific to the processing of the genesis block: a step to initialize the state store per each module and the second step to verify this initial state. In particular the processing of the genesis block happens in five steps described in the sections below.
```
AndreasKendziorra, line 34: > in five steps described in the sections below.

not a big deal: the diagram visualizes 6 steps. Also tagging @ricott1. Feel free to ignore this comment.

AndreasKendziorra (changes requested):

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: Thanks for the clarification. I think this comment can be resolved then.

proposals/lip-0058.md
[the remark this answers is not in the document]
janhack: Okay. I removed the Genesis Data substore and added the two global variables as suggested above.

proposals/lip-0058.md
janhack: Fixed, thanks.

proposals/lip-0060.md
janhack: I removed concrete counts of the number of steps from this LIP.

proposals/lip-0060.md
janhack: I removed the exact number of steps.

ricott1 (approved):
Regarding the block storage stage, I think we should align on what is currently doing in the SDK, therefore I delegate the decision to @shuse2

proposals/lip-0058.md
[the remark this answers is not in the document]
shuse2: it should be fine, I think it is used in consensus and generator but it's not open for any other modules

proposals/lip-0060.md
janhack: I added the block storage to the description in LIP 60.

proposals/lip-0060.md
janhack: The diagram is also fixed.

janhack:
I added the block storage step to the description in LIP 60 and also updated the diagram.

shuse2 (approved):

proposals/lip-0058.md
```diff
@@ -756,9 +730,9 @@ This function returns the slot number corresponding to the input timestamp.
 
 ```python
 def getSlotNumber(timestamp):
-    if timestamp < genesisDataStore.timestamp:
+    if timestamp < genesisTimestamp:
         raise Exception('Input timestamp cannot be smaller than genesis block timestamp.')
-    elapsedTime = timestamp - genesisDataStore.timestamp
+    elapsedTime = timestamp - genesisTimestamp
```
AndreasKendziorra, line 735: oh, that's something I realized just now after having a look at LIP 0063 again. The timestamp of the genesis block could be any timepoint in its time slot. See [here](https://github.com/LiskHQ/lips/blob/ce9b653f7210f0d997a599486243ec0d031ed30b/proposals/lip-0063.md?plain=1#L332). Therefore, this calculation is not necessarily doing what it intended here. We could fix this here or in LIP 0063. I'm open for both.

For `getGeneratorAtTimestamp` I'm not sure if this is an issue. I guess not.

AndreasKendziorra (approved):

karmacoma (approved):

PR #179, opened 2022-10-13
Update LIP 0052 (NFT Module) - identifiers, events, error-handling, alignment to Token module

mjerkov:
Proposed changes:
- Update to use type 'bytes' for identifiers in JSON schemas.
- Add events and error events.
- Check and fix all links.
- Rewrite and improve verification and execution of functions
- Change of module ID, command ID etc. to string 
- Redesign module substores (Align NFT module with Token module)

Relevant links:
- [Update LIPs for identifiers, errors, events, and extraCommandFee](https://trello.com/c/iRgewmJP/1191-update-lips-for-identifiers-errors-events-and-extracommandfee)
- [Change of module ID, command ID to string and change to store/substore prefix](https://trello.com/c/kaD6XQfl/1294-change-of-module-id-command-id-to-string-and-change-to-store-substore-prefix)
- [Align NFT module with Token module](https://trello.com/c/vEQm5Dcb/1397-align-nft-module-with-token-module)


proposals/lip-0052.md
```diff
@@ -94,50 +95,98 @@ This function is used to transfer ownership of NFTs across chains in the Lisk ec
...
+| Name                                   | Type   | Value                |
+| -------------------------------------- | ------ | -------------------- |
+| **Interoperability Constants**         |        |                      |
+| `MIN_RETURN_FEE`                       | uint64 | 1000                 |
+| **NFT Module Constants**               |        |                      |
+| `MODULE_NAME_NFT`                      | string | "nft"                |
+| `COMMAND_NAME_TRANSFER`                | string | "transfer"           |
+| `COMMAND_NAME_CROSS_CHAIN_TRANSFER`    | string | "transferCrossChain" |
+| `CROSS_CHAIN_COMMAND_NAME_TRANSFER`    | string | TBD                  |
+| `CHAIN_ID_ALIAS_NATIVE`                | bytes  | 0x 00 00 00 00       |
```
ricott1, line 127: I leave here the general comment about the 'chain ID alias' concept. I think we should follow the same approach we took for the Token module and remove this concept from the protocol. NFT would then just be identified by the ID of the chain also on the native chain. The reason why we introduced this alias was that before a chain not yet connected to the network did not have a chain ID at all. With the new changes, there is already a concept of chain ID independent from Interoperability (see https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#chain-identifiers). 

In practice, this would simply mean that we remove `CHAIN_ID_ALIAS_NATIVE` from the LIP (taking care that everything works out)

proposals/lip-0052.md
```diff
@@ -151,39 +200,61 @@ The function `uint64be(x)` returns the big endian uint64 serialization of an int
...
+
+All NFTs in the ecosystem are identified by using the following three values:
+
+- `chainID`, always the chain ID of the chain that created the NFT,
+- `collection`, an integer specified at NFT creation,
+- `index`, assigned at NFT creation to the next available index in the collection.
+
+In this LIP, the NFT identifier `nftID` is a 16 bytes long concatenation of the `chainID` of the NFT native chain and the serializations of `collection` and `index`: `nftID = chainID + uint32be(collection) + uint64be(index)`. This is for example used in all input formats for the module's exposed logics. This allows the exposed logic interfaces to be simple and uniform.
+
+#### NFT ID and Native NFT
```
ricott1, line 217: so for instance this section could be removed completely (see previous comment about aliasing native NFTs)

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+```python
+def verify(trs: Transaction) -> None:
 
-The `params` property of a cross-chain NFT transfer transaction is valid if:
+    chainID = nftID[:CHAIN_ID_LENGTH]
 
-* `recipientAddress` is a byte array of length `ADDRESS_LENGTH`,
-* `tokenID.chainID` is either `CHAIN_ID_ALIAS_NATIVE` or `receivingChainID`,
+    if chainID not in [CHAIN_ID_ALIAS_NATIVE, receivingChainID]:
+        raise Exception("NFT must be native to either to the mainchain or receiving chain.")
```
ricott1, line 469:
```suggestion
        raise Exception("NFT must be native to either to the sending or receiving chain.")
```

ricott1: I am not sure if we want to allow to transfer mainchain NFTs directly from sidechain to sidechain. For the moment this is not really an issue as the mainchain does not have any NFT

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
-        newAttributes = oldAttributes
-        newRecipientAddress = senderAddress
+ownChainID = Interoperability.getOwnChainAccount().ID
 
-    nftID = canonicalNFTID(nftID)
-    NFTStore(nftID).owner = newRecipienAddress
+if chainID not in [ownChainID, sendingChainID]
+    raise Exception("NFT is not native to either the sending chain or the receiving chain.")
+
+if chainID == ownChainID and NFTStore(nftID).owner != senderAddress
```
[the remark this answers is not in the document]
ricott1: Otherwise I do not understand the check

proposals/lip-0052.md
```diff
@@ -975,103 +1509,479 @@ transferCrossChain(senderAddress,
...
+def setAttributes(
+    attributes: bytes,
+    nftID: NFTID
+) -> None:
+
+    if NFTStore(nftID) does not exist:
+        emitFailedsetAttributesEvent(attributes, nftID, RESULT_NFT_DOES_NOT_EXIST)
+        raise Exception("NFT substore entry does not exist.")
+
+    if length(attributes) > MAX_BYTE_LENGTH_ATTRIBUTES:
```
ricott1, line 1682: minor point: I would make this check before, so in some cases we do not even have to read from the state

proposals/lip-0052.md
```diff
@@ -882,90 +1287,219 @@ getNextAvailableCollection():
...
     collectionStore(collection).nextAvailableIndex += 1
+
+    def emitFailedCreateEvent(
+        address: Address,
+        collection: uint32,
+        attributes: bytes,
+        result: uint32
+    ) -> None:
+
+    emitPersistentEvent(
```
sergeyshemyakov, line 1316: Here and in other places where you define `emitFailedEvent` functions, the indentation is wrong. The `emitPersistentEvent` function should be moved right by 1 tab.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+        }
+    }
+}
+```
+
+#### create
+
+This event has `name = EVENT_NAME_CREATE`, and is emitted when the create function is called.
+
+##### Topics
```
sergeyshemyakov, line 880: Here you could put the address as a topic.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+        }
+    }
+}
+```
+
+#### destroy
+
+This event has `name = EVENT_NAME_DESTROY`, and is emitted when the destroy function is called.
+
+##### Topics
```
sergeyshemyakov, line 915: Maybe also would be nice to have the address as a topic.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+This event has `name = EVENT_NAME_INITIALIZE_COLLECTION`, and is emitted when the initializeCollection function is called.
+
+##### Topics
+
+##### Data
+
+```java
+initializeCollectionEventDataSchema = {
+    "type": "object",
+    "required": [],
```
sergeyshemyakov, line 948: Add `"collection"` and `"result"` parameters, you use then when you emit this event. I think, no topics are needed here and you can just remove Topics subsection, or say that there are no topics for this event.

proposals/lip-0052.md
```diff
@@ -882,90 +1287,219 @@ getNextAvailableCollection():
...
 
     if collection >= nextAvailableCollection:
         nextAvailableCollection = collection + 1
     return collection
+
+    emitEvent(
+        module = MODULE_NAME_NFT,
+        name = EVENT_NAME_INITIALIZE_COLLECTION,
+        data = {
+            "collection": collection,
```
sergeyshemyakov, line 1397: Add `"result"` to the data.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+    "properties": {
+    }
+}
+```
+
+#### lock
+
+This event has `name = EVENT_NAME_LOCK`, and is emitted when the lock function is called.
+
+##### Topics
```
sergeyshemyakov, line 958: I think, no topics here is fine. The same for `unlock`, `setAttributes` and `recover` events.

proposals/lip-0052.md
```diff
@@ -151,39 +200,61 @@ The function `uint64be(x)` returns the big endian uint64 serialization of an int
...
     "type": "object",
     "required": [
         "owner",
-        "lockingModuleID",
+        "lockingModule",
         "attributes"
     ],
     "properties": {
         "owner": {
             "dataType": "bytes",
```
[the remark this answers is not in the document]
mjerkov: In the current design, it is either of ADDRESS_LENGTH or CHAIN_ID_LENGTH (in the case of an escrowed NFT, the owner becomes the native chain). Therefore, I stated it without conditions on the bytes length.

proposals/lip-0052.md
```diff
@@ -151,39 +200,61 @@ The function `uint64be(x)` returns the big endian uint64 serialization of an int
 
 In this LIP, `length(byteSequence)` returns the length in bytes of `byteSequence`.
 
-#### Functions from Other Modules
+#### Logic from Other Modules
+
+Calling a function `fct` from the [Interoperability module][lip-0045] is represented by `Interoperability.fct(required inputs)`.
+
+### NFT Identification
```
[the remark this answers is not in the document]
mjerkov: I agree, sometimes repetition does have a reasoning behind it.

proposals/lip-0052.md
mjerkov: Yes, I was initially confused about the chain ID alias, and will be glad to remove it now that it's not necessary anymore.

proposals/lip-0052.md
mjerkov: What do you mean by before?

proposals/lip-0052.md
```diff
@@ -882,90 +1287,219 @@ getNextAvailableCollection():
 This function creates an NFT.
 
 ```python
-create(address, collection, attributes):
-    if (length(attributes) > MAX_BYTE_LENGTH_ATTRIBUTES bytes
-        or collectionStore(collection) does not exist):
-        create fails
+def create(
```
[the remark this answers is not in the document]
mjerkov: Good point; definitely something to be discussed.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+### Internal Functions
+
+#### createNFTEntry
 
 ```python
-getAttributes(address, nftID):
-    nftID = canonicalNFTID(nftID)
-    if NFTStore(nftID) exists:
-        return NFTStore(nftID).attributes
+def createNFTEntry(nftID: NFTID, address: Address, attributes: bytes) -> None:
```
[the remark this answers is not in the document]
mjerkov: Ah, I suppose hat's what Alessandro probably had in mind when he mentioned that the check on the attributes length should have been made "before"...

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
 }
 ```
 
-##### Parameters Validity
+##### Verification
+
+```python
```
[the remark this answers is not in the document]
mjerkov: Yes! Thanks... I have those checks in both functions in "Protocol logic for other modules" section, but not here. I will add it to the cross-chain transfer command.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
+        value = encode(
+            schema = NFTStoreSchema,
+            object = {
+                "owner": address,
+                "lockingModule": NFT_NOT_LOCKED,
+                "attributes": attributes
+            }
+        )
+
+    emitEvent(
```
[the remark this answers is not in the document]
mjerkov: Yes, I will remove all the events related to substore entry creations and deletions and add a single event related to NFT creation and deletion, respectively. I will leave this comment as unresolved so we can have a final check when the time comes.

proposals/lip-0052.md
```diff
@@ -492,378 +458,817 @@ crossChainTransferParams = {
...
-        return NFTStore(nftID).owner
-    else:
-        return entry does not exist
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT substore entry does not exist.")
+
+    return NFTStore(nftID).owner
 ```
 
 #### getNextAvailableIndex
```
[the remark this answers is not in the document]
mjerkov: Yes, after we reach an agreement on the alignment (NFT -> Token), I will look at the Token LIP updated @gkoumout made; this one is only one of them.

proposals/lip-0052.md
mjerkov: Very good, thanks!

proposals/lip-0052.md
mjerkov: Right! Do you also think it should be added to event "data"? For now I only have "nftID" and "result" properties in the case of NFT destruction

proposals/lip-0052.md
```diff
@@ -882,90 +1287,219 @@ getNextAvailableCollection():
...
 #### initializeCollection
 
 This function creates a new collection substore entry.
 
 ```python
-initializeCollection(collection):
-    if collectionStore(collection) exists:
-        initializeCollection fails
+def initializeCollection(
+    collection: uint32
```
ricott1, line 1249:
```suggestion
    collection: bytes
```

proposals/lip-0052.md
```diff
@@ -975,103 +1509,479 @@ transferCrossChain(senderAddress,
...
+            "module": module,
+            "nftID": nftID,
+            "result": RESULT_SUCCESSFUL
+        },
+        topics = []
+    )
+
+    def emitFailedLockEvent(
+        module: Module,
+        nftID: NFTID
```
ricott1, line 1482:
```suggestion
        nftID: NFTID,
```

proposals/lip-0052.md
```diff
@@ -975,103 +1509,479 @@ transferCrossChain(senderAddress,
...
+            "module": module,
+            "nftID": nftID,
+            "result": RESULT_SUCCESSFUL
+        },
+        topics = []
+    )
+
+    def emitFailedUnlockEvent(
+        module: Module,
+        nftID: NFTID
```
ricott1, line 1530:
```suggestion
        nftID: NFTID,
```

proposals/lip-0052.md
```diff
@@ -975,103 +1509,479 @@ transferCrossChain(senderAddress,
...
+            "attributes": attributes,
+            "nftID": nftID,
+            "result": RESULT_SUCCESSFUL
+        },
+        topics = []
+    )
+
+    def emitFailedSetAttributesEvent(
+        attributes: bytes,
+        nftID: NFTID
```
ricott1, line 1579:
```suggestion
        nftID: NFTID,
```

proposals/lip-0052.md
```diff
@@ -492,480 +395,825 @@ crossChainTransferParams = {
...
+
+    if chainID == ownChainID:
+        if ccm.status == CCM_STATUS_OK:
+            newAttributes = getNewAttributes(collection, storedAttributes, receivedAttributes)
+            newOwner = recipientAddress
+        else:
+            newAttributes = storedAttributes
+            newOwner = senderAddress
+
+        NFTStore(nftID).owner = newOwner
```
sergeyshemyakov, line 526: Here the escrow should also be updated: you have received the nft from escrow.

mjerkov:
> Generally I think it would be good to understand the cross-chain communication of the Token module (including ccu hooks `verifyCrossChainMessage`, `beforeCrossChainCommandExecution` and `beforeCrossChainMessageForwarding`). I think, there should be many similarities with the escrow checks and updates, possibly you will have better understanding of the cross-chain modification of attributes as well.

Yes, I will definitely need a talk with you or Alessandro (or Greg) about this, since unfortunately this part is still beyond my scope of current understanding.

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
+The NFT store contains entries for all NFTs present on the native chain, as well as entries for all native NFTs that have been sent cross-chain to a foreign chain. Each entry contains two properties:
 
-The locking module ID stores the information regarding the locking status of the NFT. If the NFT is unlocked, this property will have the value `NFT_NOT_LOCKED`, whereas if the NFT is locked, this property will store the ID of the locking module.
-
-Lastly, the NFT stores an attribute property which can be used by custom applications to store information about the NFT, or modify interactions with the NFT.
+- The `owner` property can either be a `ADDRESS_LENGTH` bytes long user address or a `CHAIN_ID_LENGTH` bytes long chain ID. In the latter case, the NFT is a native NFT that has been sent cross-chain to a foreign chain and is escrowed.
```
ricott1, line 48: I think another option that was proposed was to store the `owner` property in the `attributes` array, under the NFT module. I think this option is fine as well, and at the moment I don't see a specific reason for either one. Maybe it could be simpler to store everything under `attributes` for when we mutate the NFT off-chain?

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
-| `EMPTY_BYTES`                         | bytes   | ""              |
-| `ADDRESS_LENGTH`                      | uint32  | 20              |
-
-#### uint32be Function
-
-The function `uint32be(x)` returns the big endian uint32 serialization of an integer `x`, with `0 <= x < 2^32`. This serialization is always 4 bytes long.
+| Name                                | Type   | Value                   |
+| ----------------------------------- | ------ | ----------------------- |
+| **Interoperability Constants**      |        |                         |
+| `MAINCHAIN_ID`                      | bytes  | 0x 00 00 00 00          |
```
ricott1, line 122: This is not used anywhere and could be removed. On the other hand, maybe we should use it to handle mainchain NFTs in the future

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    messageFee = trs.params.messageFee
 
-##### Execution
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT store entry does not exist.")
+
+    if len(getNFTOwner(nftID)) == CHAIN_ID_LENGTH:
+        raise Exception("NFT is already escrowed to another chain.")
+
+    if getChainID(nftID) not in [OWN_CHAIN_ID, receivingChainID]:
```
ricott1, line 520: I think we concluded to also handl the mainchain case here, see: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#verification-1


Basically we also add `MAINCHAIN_ID` to the allowed receivinChainIDs

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
+The default value for the `lockingModule` property is `NFT_NOT_LOCKED`.
 
-* The store prefix is set to `STORE_PREFIX_AVAILABLE_COLLECTION`.
-* Each store key is the empty bytes.
-* Each store value is the serialization of an object following `availableCollectionStoreSchema`.
+#### Escrow Substore
+
+- The substore prefix is set to `SUBSTORE_PREFIX_ESCROW`.
+- Each store key is the identifier of the chain to which the NFTs are escrowed, and the NFT ID of the escrowed NFT: `escrowedChainID + nftID`.
+- Each store value is the serialization of an object following `escrowStoreSchema`.
```
ricott1, line 309: Here I think we should use the empty schema for the store value (basically we store nothing). We indicate that the NFT is escrowed simply by the existence of the entry in the substore, which we should delete once the NFT is unescrowed.

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
-```python
-canonicalNFTID(nftID):
-    if nftID.chainID == interoperability.getOwnChainAccount().ID
-        return {"chainID": CHAIN_ID_ALIAS_NATIVE, "collection": nftID.collection, "index": nftID.index}
-    else:
-        return nftID
-```
-
-This function will return the input `nftID` in the case `interoperabiliby.getOwnChainAccount()` cannot be called.
+If not all NFTs are supported, but all NFTs from a chain with `chainID` are supported, the substore contains an entry for key `chainID` with an empty array as value.
```
ricott1, line 358: So in my understanding if an entry exist for key `chainID` with a non-empty array as value, it means that we support only the specified NFTs from that chain? I am wondering if this level of granularity is really required and if we shouldn't just support or not NFTs from a certain chain. On the other hand, if we want something more granular, I would rather include the `collectionID` in the store value, so that we specify that we support all NFTs of a certain collection from a certain chain, by specifying a non-empty store value.

In all cases, I would add a line to explain this last example (entry existing with for key `chainID` with a non-empty array as value).

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
-The `params` property of an NFT transfer transaction is valid if:
+    senderAddress = sha256(trs.senderPublicKey)[:ADDRESS_LENGTH]
+    nftID = trs.params.nftID
 
-* `recipientAddress` must be a byte array of length `ADDRESS_LENGTH`.
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT store entry does not exist.")
 
-##### Execution
+    if getNFTOwner(nftID) == OWN_CHAIN_ID:
```
ricott1, line 428: This check seems superfluous given the following one. I would remove it.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+        raise Exception("NFT is already escrowed to another chain.")
+
+    if getNFTOwner(nftID) != senderAddress:
+        raise Exception("Transfer not initiated by the NFT owner.")
+
+    if getLockingModule(nftID) != NFT_NOT_LOCKED:
+        raise Exception("NFT is locked.")
```
ricott1, line 435: I would slightly improve, maybe something like: "Locked NFTs cannot be transferred"

proposals/lip-0052.md
ricott1: Also, I think the check is wrong, shouldn't we check that the length of the owner matches a chain ID length (so it could be any chain)?

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    receivingChainID = trs.params.receivingChainID
+    recipientAddress = trs.params.recipientAddress
+    data = trs.params.data
+    messageFee = trs.params.messageFee
 
-##### Execution
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT store entry does not exist.")
+
+    if len(getNFTOwner(nftID)) == CHAIN_ID_LENGTH:
```
ricott1, line 517: As before, I think this check is superfluous given the following one on the senderAddress. Maybe it is fine to keep it to give a more granular error message? In that case you should also add it to the normal transfer function.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+        raise Exception("NFT must be native to either the sending or the receiving chain.")
 
-When executing a cross-chain NFT transfer transaction `trs`, the following is done:
+    if getNFTOwner(nftID) != senderAddress:
+        raise Exception("Transfer not initiated by the NFT owner.")
 
-* Derive `senderAddress` from `trs.senderPublicKey`.
-* Execute the same logic as the function
+    if getLockingModule(nftID) != NFT_NOT_LOCKED:
+        raise Exception("NFT is locked.")
```
ricott1, line 527: Same as above about error message

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+def execute(trs: Transaction) -> None:
+
+    senderAddress = sha256(trs.senderPublicKey)[:ADDRESS_LENGTH]
+    nftID = trs.params.nftID
+    receivingChainID = trs.params.receivingChainID
+    recipientAddress = trs.params.recipientAddress
+    data = trs.params.data
+    messageFee = trs.params.messageFee
+
+    if getChainID(nftID)== OWN_CHAIN_ID and escrowStore(receivingChainID, nftID) does not exist:
```
ricott1, line 546: I would move this to the `transferCrossChainInternal` function.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+##### Verification
+
+```python
+def verify(
+    ccu: Transaction,
+    ccm: CCM
+) -> None:
+
+    nftID = ccm.params.nftID
+    senderAddress = ccm.params.senderAddress
```
ricott1, line 610: some of this variables are not used and should be removed

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    recipientAddress = ccm.params.recipientAddress
+    data = ccm.params.data
+    sendingChainID = ccm.sendingChainID
+
+    if getChainID(nftID) not in [OWN_CHAIN_ID, sendingChainID]:
```
ricott1, line 615: same as before, we should add support to mainchain NFTs

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+
+    if isNFTSupported(nftID) == False:
+        emitPersistentEvent(
+            module = MODULE_NAME_NFT,
+            name = EVENT_NAME_CCM_TRANSFER,
+            data = {
+                "senderAddress": senderAddress,
+                "recipientAddress": recipientAddress,
+                "nftID": nftID,
+                "receivingChainID": OWN_CHAIN_ID,
```
ricott1, line 643: Why do we add the receivingChainID to the data and, say, not the sendingChainID? I would consistently add both orr neither (maybe neither is better?) Notice that bot values can be recovered from the ccm anyway (and the ccm ID is by default one of the event topics)

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+            "dataType": "bytes",
+            "length": ADDRESS_LENGTH,
+            "fieldNumber": 2
+        },
+        "nftID": {
+            "dataType": "bytes",
+            "length": NFT_ID_LENGTH,
+            "fieldNumber": 3
+        },
+        "receivingChainID": {
```
ricott1, line 810: see previous comment, I would maybe remove this

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
+| **NFT Module Constants**            |        |                         |
+| `MODULE_NAME_NFT`                   | string | "nft"                   |
+| `COMMAND_NAME_TRANSFER`             | string | "transfer"              |
+| `COMMAND_NAME_CROSS_CHAIN_TRANSFER` | string | "transferCrossChain"    |
+| `CROSS_CHAIN_COMMAND_NAME_TRANSFER` | string | TBD                     |
+| `CCM_STATUS_OK`                     | uint32 | 0                       |
+| `CCM_STATUS_NFT_NOT_SUPPORTED`      | uint32 | 64                      |
+| `CCM_STATUS_PROTOCOL_VIOLATION`     | uint32 | 65                      |
+| `NFT_NOT_LOCKED`                    | string | `MODULE_NAME_NFT`       |
+| `MAX_BYTE_LENGTH_ATTRIBUTES`        | uint32 | 9 \* 1024               |
```
ricott1, line 132: Why did we choose 9kb here? Notice that previously 9kb was the limit for the whole array, while now we are limiting only one entry, so this should probably be smaller (or we should rather check the whole array size)

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    # This may fail if there is not enough fee left.
+    Fee.payFee(ESCROW_ENTRY_INITIALIZATION_FEE)
+
+    create substore entry with
+        substorePrefix = SUBSTORE_PREFIX_ESCROW
+        key = receivingChainID + nftID
+        value = encode(
+            schema = escrowStoreSchema,
+            object = {
+                "escrow": True
```
ricott1, line 1210: As mentioned above, I would rather use the empty schema here

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
         createUserEntry(recipientAddress, nftID)
-    else:
-        createNFTEntry(nftID,
-                       senderAddress,
-                       receivedAttributes)
-        createUserEntry(senderAddress, nftID)
+
+    if getChainID(nftID) == OWN_CHAIN_ID:
+        NFTStore(nftID).owner = recipientAddress
+        escrowEntry(sendingChainID, nftID).escrow = False
```
ricott1, line 659: I think we should rather delete the entry for consistency. If the entry does not exist, it is not escrowed, otherwise we also have to check the value.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    module: Module,
+    nftID: NFTID,
+    attributes: Attributes
+) -> None:
+
+    if NFTStore(nftID) does not exist:
+        emitFailedSetAttributesEvent(module, nftID, attributes, RESULT_NFT_DOES_NOT_EXIST)
+        raise Exception("NFT substore entry does not exist.")
+
+    if getChainID(nftID) != OWN_CHAIN_ID:
```
ricott1, line 1601: Why are we limiting modifying attributes only to native chains? Isn't it an important use case to modify non-native NFTs? Maybe @MaximeGagnebin can chim in?

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+}
+```
+
+#### setAttributes
+
+This event has `name = EVENT_NAME_SET_ATTRIBUTES`, and is emitted when the [setAttributes](#setattributes-2) function is called.
+
+##### Topics
+
+- `module`: name of the module that sets the attributes.
```
ricott1, line 990: I don't think this is a useful topic and I would remove it. I don't see why someone should query all events related to a module.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    senderAddress: Address,
+    recipientAddress: Address,
+    nftID: NFTID
+) -> None:
+
+    if NFTStore(nftID) does not exist:
+        emitFailedTransferEvent(senderAddress, recipientAddress, nftID, RESULT_NFT_DOES_NOT_EXIST)
+        raise Exception("NFT store entry does not exist.")
+
+    if getNFTOwner(nftID) == OWN_CHAIN_ID:
```
ricott1, line 1654: Same as above, this check is superfluous but we could keep it for extra error granularity, in which case we should fix the check `len(getNFTOwner(nftID)) == LENGTH_CHAIN_ID`

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+
+    if len(data) > MAX_DATA_LENGTH:
+        emitFailedCrossChainTransferEvent(senderAddress, nftID, amount, receivingChainID, recipientAddress, RESULT_DATA_TOO_LONG)
+        raise Exception("Data field is too long.")
+
+    if NFTStore(nftID) does not exist:
+        emitFailedTransferCrossChainEvent(senderAddress, recipientAddress, nftID, receivingChainID, RESULT_NFT_DOES_NOT_EXIST)
+        raise Exception("NFT store entry does not exist.")
+
+    if getNFTOwner(nftID) == OWN_CHAIN_ID:
```
ricott1, line 1710: same as previous comment

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+        emitFailedTransferCrossChainEvent(senderAddress, recipientAddress, nftID, receivingChainID, RESULT_NFT_ESCROWED)
+        raise Exception("NFT is escrowed to another chain.")
+
+    if getChainID(nftID) not in [OWN_CHAIN_ID, receivingChainID]:
```
ricott1, line 1714: We should add `MAINCHAIN_ID`

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    if getLockingModule(nftID) != NFT_NOT_LOCKED:
+        emitFailedTransferCrossChainEvent(senderAddress, recipientAddress, nftID, receivingChainID, RESULT_NFT_LOCKED)
+        raise Exception("NFT is locked.")
+
+    messageFeeTokenID = Interoperability.getMessageFeeTokenID(receivingChainID)
+    if Token.getAvailableBalance(senderAddress, messageFeeTokenID) < messageFee:
+        emitFailedTransferCrossChainEvent(senderAddress, recipientAddress, nftID, receivingChainID, RESULT_INSUFFICIENT_BALANCE)
+        raise Exception("Insufficient balance for the message fee.")
+
+    if getChainID(nftID)== OWN_CHAIN_ID and escrowStore(receivingChainID, nftID) does not exist:
```
ricott1, line 1731: I would move it to `transferCrossChainInternal`

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+#### createEscrowEntry
+
+```python
+def createEscrowEntry(
+    receivingChainID: ChainID,
+    nftID: NFTID
+) -> None:
+
+    # This may fail if there is not enough fee left.
+    Fee.payFee(ESCROW_ENTRY_INITIALIZATION_FEE)
```
ricott1, line 1202: I am not sure that we should pay a fee here. If the NFT is transferred cross-chain, we create an escrow entry but we also delete a user entry, so the state does not increase in size.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
 
-This function returns the owner of an NFT.
+```python
+def createUserEntry(
+    address: Address,
+    nftID: NFTID
+) -> None:
+
+    # This may fail if there is not enough fee left.
+    Fee.payFee(USER_ENTRY_INITIALIZATION_FEE)
```
ricott1, line 1180: I am not sure that we should pay a fee here. This entry is created when a new NFT is minted (in which case we already pay that fee for the state increase and we could just pay it once in the createNFT function) or when an NFT is transferred from another user or from an escrow. If we delete the escrow account as I suggest, in both cases the total state would not increase, so we can avoid paying the fee.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+            "senderAddress": senderAddress,
+            "recipientAddress": recipientAddress,
+            "nftID": nftID,
+            "receivingChainID": receivingChainID,
+            "result": RESULT_SUCCESSFUL
+        },
+        topics = [senderAddress, recipientAddress, receivingChainID]
+    )
+
+    Interoperability.send(
```
ricott1, line 1274: We modified interop.send and now the params input must be of type bytes (you just need to encode it before sending it)

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+```python
+def getChainID(nftID: NFTID) -> ChainID:
+
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT substore entry does not exist.")
+
+    return nftID[:CHAIN_ID_LENGTH]
+```
+
+#### getAttributes
```
ricott1, line 1331: I would rename this `getAttributesArray` and have another function to just get the attributes for a certain module `getAttributes(nftID: NFTID, module: str) -> bytes`

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+#### getLockingModule
+
+This function returns the locking status of an NFT.
 
 ```python
-getNFTowner(nftID):
-    nftID = canonicalNFTID(nftID)
-    if NFTStore(nftID) exists:
-        return NFTStore(nftID).owner
+def getLockingModule(nftID: NFTID) -> string:
```
ricott1, line 1349:
```suggestion
def getLockingModule(nftID: NFTID) -> str:
```

to conform to python notation

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+#### getNFTOwner
 
-This function returns the max index of a collection.
+This function returns the owner of an NFT.
 
 ```python
-getNextAvailableIndex(collection):
-    if collectionStore(collection) does not exist:
-        return collection does not exist
+def getNFTOwner(nftID: NFTID) -> Address:
```
ricott1, line 1365:
```suggestion
def getNFTOwner(nftID: NFTID) -> Address | ChainID:
```

proposals/lip-0052.md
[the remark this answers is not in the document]
ricott1: I think we now pay the fee in `createNFTEntry`

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
-
-    address = NFTStore(nftID).owner
-    deleteNFTEntry(nftID)
-    deleteUserEntry(address, nftID)
+def destroy(
+    address: Address,
+    nftID: NFTID
+) -> None:
+
+    if getChainID(nftID) != OWN_CHAIN_ID:
```
ricott1, line 1428: Is there a reason why we don't allow to destroy non-native NFTs?

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+
+    chainID = storeKey[:CHAIN_ID_LENGTH]
+    nftID = storeKey
+    nftValue = decode(schema = NFTStoreSchema, object = storeValue)
+
+    if getChainID(nftID) != OWN_CHAIN_ID:
+        emitFailedRecoverEvent(terminatedChainID, nftID, RESULT_INITIATED_BY_NONNATIVE_CHAIN)
+        raise Exception("Recovery called by non-native chain.")
+
+    if NFTStore(nftID).owner != terminatedChainID or escrowStore(terminatedChainID, nftID).escrow == False:
```
ricott1, line 1787:
```suggestion
    if NFTStore(nftID).owner != terminatedChainID:
```

ricott1: If we follow my suggestion to just delete escrow accounts

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+        emitFailedRecoverEvent(terminatedChainID, nftID, RESULT_NFT_NOT_ESCROWED)
+        raise Exception("NFT was not escrowed to terminated chain.")
+
+    if len(nftValue.owner) != ADDRESS_LENGTH:
+        emitFailedRecoverEvent(terminatedChainID, nftID, RESULT_INVALID_ACCOUNT)
+        raise Exception("Invalid account information.")
+
+    NFTStore(nftID).owner = nftValue.owner
+    createUserEntry(nftValue.owner, nftID)
+    escrowStore(terminatedChainID, nftID).escrow = False
```
ricott1, line 1797: Would just delete

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    emitEvent(
+        module = MODULE_NAME_NFT,
+        name = EVENT_NAME_REMOVE_SUPPORT_ALL_NFTS_FROM_CHAIN,
+        data = {"chainID": chainID},
+        topics = [chainID]
+    )
 ```
 
-#### recover
+#### supportNftID
```
ricott1, line 1927: To follow up on previous comment, I would think that supporting a certain collection rather than a specific nftID would be more useful

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+
+During the genesis state initialization stage, the following steps are executed. If any step fails, the block is discarded and has no further effect.
+
+Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets for the NFT module and let `genesisBlockAssetObject` be the deserialization of `genesisBlockAssetBytes` according to the `genesisNFTStoreSchema` schema, given above.
+
+- Initial checks on the properties of `genesisBlockAssetObject`:
+
+  - Across all elements of the `NFTSubstore` array, all values given for `nftID` must be unique.
+  - For all elements of the `NFTSubstore` array, values given for `owner` must have either length `ADDRESS_ID_LENGTH` bytes (representing a user address) or `CHAIN_ID_LENGTH` bytes (representing a chain ID).
+  - The `NFTSubstore` must be in lexicographical order of `owner`.
```
ricott1, line 2151: maybe this is a remnant of a previous version, but i think it should be in lexicographical order of `nftID`.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    - For a given `owner`, the entries must be in ascending order of `nftID`.
```
ricott1, line 2152: This should be deleted

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+    - The `attributesArray` must be in lexicographic order of `module` name.
+  - Across all elements of the `userSubstore` array, the pairs `address`, `nftID` must be unique: each pair appears at most once, although a given `address` can appear multiple times with different unique `nftID`.
```
ricott1, line 2154: I think `nftID` should also be independently unique

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+  - The `userSubstore` array must be in lexicographical order of `address`. For a given `address`, the entries must be in lexicographic order of `nftID`.
+  - Across all elements of the `escrowSubstore` array, the pairs `escrowedChainID`, `nftID` must be unique: each pair appears at most once, although a given `escrowedChainID` can appear multiple times with different unique `nftID`.
```
ricott1, line 2156: same as above, `nftID` should be independently unique

ricott1: Just to confirm: we do not check that all entries in NFT substore have a corresponding entry in either user or escrow substore because in principle the NFT could have been destroyed?

Even in this case, I think we should check that all entries in user and escrow substore together must have a corresponding entry in the NFT substore.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+  storeValue = encode(
+      schema = NFTStoreSchema,
+      object = {
+          "owner": NFTEntry.owner,
+          "attributesArray": NFTEntry.attributesArray
+      }
+  )
+  ```
+
+  Furthermore, if `NFTEntry.owner` has length `ADDRESS_ID_LENGTH` bytes, create an entry in the user substore with:
```
ricott1, line 2172: We do not support a genesis state with locked NFTs. This is fine for me, but in principle we could quite easily allow for that.

ricott1 (changes requested):
I left many comments but several are just redundant.

To summarize my main points:

- delete escrow accounts instead of setting value to False. Use empty schema for store value.
- Remove user and escrow initialization fees
- some general comments on small mistakes in some checks
- Some improvements on logic regarding supported NFTs

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+  - The `NFTSubstore` must be in lexicographical order of `owner`.
+    - For a given `owner`, the entries must be in ascending order of `nftID`.
+    - The `attributesArray` must be in lexicographic order of `module` name.
+  - Across all elements of the `userSubstore` array, the pairs `address`, `nftID` must be unique: each pair appears at most once, although a given `address` can appear multiple times with different unique `nftID`.
+  - The `userSubstore` array must be in lexicographical order of `address`. For a given `address`, the entries must be in lexicographic order of `nftID`.
+  - Across all elements of the `escrowSubstore` array, the pairs `escrowedChainID`, `nftID` must be unique: each pair appears at most once, although a given `escrowedChainID` can appear multiple times with different unique `nftID`.
+  - The `supportedNFTsSubstore` array must be ordered lexicographically by `chainID`. For each entry of this array, the `supportedNftIDArray` should be in lexicographical order.
+
```
sergeyshemyakov, line 2158: I would add a check that escrow information is consistent across NFT and escrow substores: for every entry in `NFTSubstore` array with owner being another chain, check that there is an entry in the `escrowSubstore` with the same `nftID` (and `escrow = True` if we keep it like that).

sergeyshemyakov (changes requested):
I think we still need to discuss couple of points, but it is surely getting closer!

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+            topics = [senderAddress, recipientAddress]
+        )
+        raise Exception("Non-supported NFT.")
+
+    if ccm.status != CCM_STATUS_OK:
+        recipientAddress = senderAddress
+
+    if NFTStore(nftID) does not exist:
+        createNFTEntry(recipientAddress, nftID)
         createUserEntry(recipientAddress, nftID)
```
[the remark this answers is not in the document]
ricott1: Agree, and I think we should even have a check in the verify. If the NFT is native, then the `NFTStore(nftID)` must exist. If it's not, then it must not exist

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
 
-This function is used to lock an NFT to a module ID. A locked NFT cannot be transferred (within the chain or across chains). This can be useful, for example, when the NFT is used as a deposit for a service. A module ID is specified when locking the NFT and this ID has to be specified when unlocking the NFT. This avoids NFTs being accidentally locked and unlocked by different modules.
+This function is used to lock an NFT to a module. A locked NFT cannot be transferred (within the chain or across chains). This can be useful, for example, when the NFT is used as a deposit for a service. Module is specified both when locking and unlocking the NFT, thus preventing NFTs being accidentally locked and unlocked by different modules.
 
 #### unlock
 
-This function is used to unlock an NFT that was locked to a module ID.
+This function is used to unlock an NFT that was locked to a module.
 
 #### setAttributes
```
ricott1, line 97: I left a comment about this later on. I think it would be nice to explain how this method could be used to process NFTs coming back to the native chain in a ccm with possibly updated attributes. The idea then would be that the NFT module does not update the attributes, but rather the custom module should call this method from the `afterCrossChainMessageExecute` hook.

proposals/lip-0052.md
```diff
@@ -30,329 +31,345 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
+When an NFT is sent to another chain, the `attributes` property of the NFT can be modified according to specifications set on the receiving chain. For this reason, custom modules specifying an NFT collection must also implement the behavior to adopt when an NFT is returned with a modified `attributes` property. This custom behavior will compare the returned attributes with the ones stored with the escrowed NFT. If the returned NFT has an empty attributes, the native chain will restore the attributes as stored, this can be used to save on cross-chain messages size when returning non-modified NFTs to their native chains.
```
ricott1, line 75: I think we shpould drop the last sentence or so, as I mention in other comments

proposals/lip-0052.md
mjerkov: Yes, I put it like that for new because the cross-chain modification of attributes is something I would like to discuss with you. Actually, the same goes with all modification of attributes, even on native chain, since we don't have implemented the distinction between 'fixed' and 'mutable' attributes. 

As I see it, 'fixed' attributes would be the ones set by the collectionID (or particular NFTs, to be decided) creator, and could not be mutated by anyone. The NFT creator would then be able to decide whether he allows for the existence of 'mutable' attributes that other modules (both on native and foreign chains) can then mutate.

In my view, something to be discussed on an STM before (potentially) implementing it in the protocol. So, this setAttributes function is really unfinished business for now.

proposals/lip-0052.md
mjerkov: Agree. I don't yet have a perfect understanding of events, so thanks for this comment because it helps me understand the logic behind events.

proposals/lip-0052.md
mjerkov: I don't have a specific opinion about this, either. I would leave it as it is for now, and we can touch upon it in the STM.

proposals/lip-0052.md
mjerkov: I agree and have changed escrowing accordingly.

proposals/lip-0052.md
mjerkov: This the supporting logic was copy-pasted from the Token LIP. I was wandering the same, namely if for NFTs it makes sense to support the individual NFTs instead of the whole collection. So, yes, I'm also in favor of supporting the whole collections instead of individual NFTs and I will change the logic accordingly.

proposals/lip-0052.md
mjerkov: Added. The same was noticed by @ricott1.

proposals/lip-0052.md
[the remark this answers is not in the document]
mjerkov: Agreed.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
+        createNFTEntry(recipientAddress, nftID)
         createUserEntry(recipientAddress, nftID)
-    else:
-        createNFTEntry(nftID,
-                       senderAddress,
-                       receivedAttributes)
-        createUserEntry(senderAddress, nftID)
+
+    if getChainID(nftID) == OWN_CHAIN_ID:
+        NFTStore(nftID).owner = recipientAddress
```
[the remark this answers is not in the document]
mjerkov: This is something I would like to discuss live, when time allows.

proposals/lip-0052.md
mjerkov: Again, cross-chain modification of attributes is one of the topics that I would like to further discuss live.

proposals/lip-0052.md
mjerkov: Right. We can also discuss this briefly live.

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
-    createNFTEntry(nftID, address, attributes)   
+def create(
+    address: Address,
+    collectionID: Collection
+) -> None:
+
+    index = getNextAvailableIndex(collectionID)
+    nftID = OWN_CHAIN_ID + collectionID + uint64be(index)
+
+    createNFTEntry(address, nftID)
```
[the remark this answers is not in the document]
ricott1: I am not 100% sold on this. For instance, when we receive the ccm we forgot to pay the fee now, which could have been avoided if the fee was paid in the createNFTEntry

proposals/lip-0052.md
```diff
@@ -362,712 +364,1852 @@ The module provides the following commands to modify the NFT store.
...
+        )
+        raise Exception("Non-supported NFT")
+
+    if ccm.status != CCM_STATUS_OK:
+        recipientAddress = senderAddress
+
+    createUserEntry(recipientAddress, nftID)
+
+    if NFTStore(nftID) does not exist:
+        createNFTEntry(recipientAddress, nftID)
```
ricott1, line 649: I think here we should call `Fee.payFee(FEE_CREATE_NFT)`

proposals/lip-0052.md
ricott1: These attributes would be stored also in the native chain. In a sense, I think only the attributes stored in the native chain should be considered the 'real' ones.

proposals/lip-0052.md
ricott1: I think it is fine like it is, keeping the `owner` property separate.

proposals/lip-0052.md
sergeyshemyakov: Hm, now I don't think that a setAttributes event should have `address` as a topic, since there could be no particular address initiating the attribute state change (e.g. a DeFi module automatically updates a status of an NFT). 
So it is fine as it is form my side, feel free to resolve.

proposals/lip-0052.md
sergeyshemyakov: Okay, I call off my comment about having the address. Probably, we don't need it here for the same reason as above: `nftID` is already given and the destroyer might not have an address (could be some DeFi module).

proposals/lip-0052.md
```diff
@@ -362,712 +379,1832 @@ The module provides the following commands to modify the NFT store.
...
-initializeCollection(collection):
-    if collectionStore(collection) exists:
-        initializeCollection fails
+def lock(
+    module: Module,
+    nftID: NFTID
+) -> None:
+
+    if NFTStore(nftID) does not exist:
+        emit FailedLockEvent(module, nftID, RESULT_NFT_DOES_NOT_EXIST)
```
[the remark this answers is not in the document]
sergeyshemyakov: `emit FailedLockEvent` -> `emitFailedLockEvent`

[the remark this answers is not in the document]
mjerkov: I see now, thanks!

proposals/lip-0052.md
```diff
@@ -362,712 +365,1901 @@ The module provides the following commands to modify the NFT store.
...
+    receivedAttributes = ccm.params.attributes
+    data = ccm.params.data
+    receivingChainID = ccm.receivingChainID
+    sendingChainID = ccm.sendingChainID
+
+    if getChainID(nftID) == OWN_CHAIN_ID: # execution on the native chain
+        storedAttributes = NFTStore(nftID).attributesArray
+        if ccm.status == CCM_STATUS_OK:
+            NFTStore(nftID).owner = recipientAddress
+            NFTStore(nftID).attributesArray = getNewAttributes(nftID, storedAttributes, receivedAttributes)
```
sergeyshemyakov, line 650: This logic of CCM processing will not allow modules to define custom handling of the NFT attributes. It will be just centrally decided from the NFT module (in particular, by the `getNewAttributes` function).

I think, the following could be a solution: every module that works with NFT attributes should define `afterCrossChainCommandExecution` CCM hook (see https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#apply). In this hook, the module will have access to the CCM with the attributes from another chain and `nftID`, to the `getAttributes` and `setAttributes` methods of the NFT module. Then they can implement any custom logic, for example update some attributes and discard some other ones.

Note that the default behavior for all modules that have not implemented `afterCrossChainCommandExecution` will be correct: the received attributes will be discarded. 

I am also interested to hear from @ricott1 about this.

proposals/lip-0052.md
```diff
@@ -30,329 +31,331 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
 
-When an NFT is sent to another chain, the attributes property of the NFT can be modified according to specifications set on the receiving chain. For this reason, custom modules specifying an NFT collection must also implement the behavior to adopt when an NFT is returned with a modified attributes property. This custom behavior will compare the returned attributes with the ones stored with the escrowed NFT. If the returned NFT has an empty attribute, the native chain will restore the attributes as stored, this can be used to save on cross-chain messages size when returning non-modified NFTs to their native chains.
+When an NFT is sent to another chain, the `attributes` properties of the NFT can be modified according to specifications set on the receiving chain, i.e., in the processing done after the cross-chain NFT transfer. For this reason, custom modules specifying an NFT must also implement the behavior to adopt when an NFT is returned with a modified `attributes` properties. By default, an NFT returning back to its native chain will have the attributes as saved in the escrow account.
```
sergeyshemyakov, line 75: Once the topic is completely finalized (see my comment in CCM execution), it would be nice to explicitly say what a module should do in order to specify the cross-chain attribute processing logic.

E.g. for my proposal add something like: "Once an NFT is received back at its native chain, any module can perform custom attribute processing according to the logic specified in `afterCrossChainCommandExecution`. In particular, a module might update all the attributes to the values received cross-chain, or keep the old values, or implement more complicated processing."

sergeyshemyakov (changes requested):
We still need to finalize the cross-chain attribute receiving functionality. See my comment with details.

proposals/lip-0052.md
ricott1: Since you are returning, you can just raise at the end and remove the `attributes_exist` variable all together

proposals/lip-0052.md
```diff
@@ -30,329 +31,331 @@ Therefore, providing a unified module to handle, transfer and modify NFTs is a n
...
-This function is used to destroy NFTs. The NFT will be removed from the NFT store and cannot be retrieved. The use of this function is limited to destroying native NFTs.
-
-#### transfer
-
-This function is used to transfer ownership of NFTs within one chain.
-
-#### transferCrossChain
-
-This function is used to transfer ownership of NFTs across chains in the Lisk ecosystem.
+This function is used to destroy NFTs. The NFT will be removed from the NFT substore and cannot be retrieved. The use of this function is limited to destroying native NFTs.
```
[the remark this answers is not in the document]
mjerkov: Right!

proposals/lip-0052.md
```diff
@@ -362,712 +365,1901 @@ The module provides the following commands to modify the NFT store.
...
+
+    senderAddress = sha256(trs.senderPublicKey)[:LENGTH_ADDRESS]
+    nftID = trs.params.nftID
+    receivingChainID = trs.params.receivingChainID
+    recipientAddress = trs.params.recipientAddress
+    data = trs.params.data
+    messageFee = trs.params.messageFee
+    includeAttributes = trs.params.includeAttributes
+
+    if includeAttributes:
```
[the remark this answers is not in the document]
mjerkov: Of course :)... thanks

proposals/lip-0052.md
```diff
@@ -362,712 +365,1901 @@ The module provides the following commands to modify the NFT store.
...
             "dataType": "bytes",
+            "length": LENGTH_ADDRESS,
             "fieldNumber": 2
         },
         "recipientAddress": {
             "dataType": "bytes",
+            "length": LENGTH_ADDRESS,
             "fieldNumber": 3
         },
         "attributes": {
```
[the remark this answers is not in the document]
mjerkov: Yes, I was aware that it should be the array, just didn't quite know how to write it in short version, so I suppose it should be written in total, as in the NFT substore schema specification. Thanks.

proposals/lip-0052.md
```diff
@@ -362,712 +365,1901 @@ The module provides the following commands to modify the NFT store.
...
+    if NFTStore(nftID) does not exist:
+        raise Exception("NFT substore entry does not exist")
 
-When executing a cross-chain NFT transfer transaction `trs`, the following is done:
+    if len(getNFTOwner(nftID)) == LENGTH_CHAIN_ID:
+        raise Exception("NFT is escrowed to another chain")
 
-* Derive `senderAddress` from `trs.senderPublicKey`.
-* Execute the same logic as the function
+    if getChainID(nftID) not in [OWN_CHAIN_ID, receivingChainID, MAINCHAIN_ID]:
```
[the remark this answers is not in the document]
mjerkov: Removed!

proposals/lip-0052.md
mjerkov: Noted!

proposals/lip-0052.md
mjerkov: Noted.

proposals/lip-0052.md
```diff
@@ -1478,7 +1496,7 @@ def create(
             "collectionID": collectionID,
             "result": RESULT_SUCCESSFUL
         },
-        topics = [address, nftID]
+        topics = [nftID]
```
ricott1, line 1499: Why did we remove the address from this event? Sorry if this was from a previous review comment that I missed.

mjerkov: My bad, I got confused while reading https://github.com/LiskHQ/lips/pull/179/files/73ac49ad2f4341097be5f723b1dc115d1f98478f#diff-75b2968202aef053f6535bf1d9c00465475cf0ef69c2909f498569b1994227d4. I will return it back to the topics.

proposals/lip-0052.md
```diff
@@ -673,9 +669,10 @@ def execute(
             NFTStore(nftID).attributesArray = getNewAttributes(nftID, storedAttributes, receivedAttributes)
             createUserEntry(recipientAddress, nftID)
             delete entry escrowEntry(sendingChainID, nftID) from the escrow substore
-        else: # message is returning to the non-native chain
-            createNFTEntry(senderAddress, nftID, receivedAttributes)
+        else: # message is returning to the native chain
```
ricott1, line 685: I was thinking that this handling is correct only if the ccm status is one set byy Interoperability (so that it is guaranteed that sending and receiving chains have been swapped). I think we should explicitly only handle these cases, so I would change `else` to `elif ccm.status <= MAX_RESERVED_ERROR_STATUS` (MAX_RESERVED_ERROR_STATUS is defined in LIP0045).

ricott1, line 685:
```diff
elif ccm.status <= MAX_RESERVED_ERROR_STATUS 
```

proposals/lip-0052.md
```diff
@@ -691,12 +688,12 @@ def execute(
             )
             raise Exception("Non-supported NFT")
         if ccm.status == CCM_STATUS_OK:
+            Fee.payFee(FEE_CREATE_NFT)
             createNFTEntry(recipientAddress, nftID, receivedAttributes)
             createUserEntry(recipientAddress, nftID)
-        else: # message is returning to the native chain
-            NFTStore(nftID).owner = senderAddress
+        else: # message is returning to the non-native chain
```
ricott1: same as above

proposals/lip-0052.md
```diff
@@ -1394,7 +1391,7 @@ def getCollectionID(nftID: NFTID) -> CollectionID:
 This function returns the `attributesArray` of an NFT.
 
 ```python
-def getAttributesArray(nftID: NFTID) -> List[{Module, Attributes}]:
+def getAttributesArray(nftID: NFTID) -> List[{Module, bytes}]:
```
ricott1, line 1407:
```suggestion
def getAttributesArray(nftID: NFTID) -> List[dict[Module, bytes]]:
```

ricott1: If you use capital `List` you have to import it frm typing, while from python 3.5 I think u can just use `list`

ricott1 (changes requested):
I left a couple minor comments. Almost done!

proposals/lip-0052.md
mjerkov: Changed as suggested.

mjerkov: Changed as suggested.

proposals/lip-0052.md
mjerkov: Done!

proposals/lip-0052.md
mjerkov: Done.

proposals/lip-0052.md
mjerkov: Done.

proposals/lip-0052.md
```diff
@@ -147,6 +147,7 @@ The following constants are used throughout the document:
 | `LENGTH_COLLECTION_ID`              | uint32 | 4                       |
 | `LENGTH_TOKEN_ID`                   | uint32 | 8                       |
 | `MAX_LENGTH_DATA`                   | uint32 | 64                      |
+| `MAX_RESERVED_ERROR_STATUS`         | uint64 | 63                      |
```
ricott1, line 150: This constant and `CCM_STATUS_OK` should go under the 'Interoperability constants' section

ricott1 (changes requested):
I left another small comment about the table. Regarding the creation fee, I think I wouldn't change it by introducing another fee.

ricott1 (approved):
Great job!

proposals/lip-0052.md
ricott1: can be resolved

proposals/lip-0052.md
ricott1: can be resolved

proposals/lip-0052.md
ricott1: can be resolved

proposals/lip-0052.md
```diff
@@ -362,712 +377,1926 @@ The module provides the following commands to modify the NFT store.
...
+    if getChainID(nftID) != OWN_CHAIN_ID and NFTStore[nftID] entry exists:
+        raise Exception("NFT substore entry already exists")
 
-* Create an entry in the available collection substore with:
-  ```python
-  storeKey = EMPTY_BYTES
-  storeValue = {"nextAvailableCollection": genesisBlockAssetObject.availableCollectionSubstore}
-               serialized using availableCollectionStoreSchema.
-  ```
+    if ccm.status > MAX_RESERVED_ERROR_STATUS:
```
ricott1, line 661: I would do this check first, so in case of errors we do not need to access the state

proposals/lip-0052.md
```diff
@@ -362,712 +377,1926 @@ The module provides the following commands to modify the NFT store.
...
+            )
+            raise Exception("Non-supported NFT")
+        if ccm.status == CCM_STATUS_CODE_OK:
+            Fee.payFee(FEE_CREATE_NFT)
+            createNFTEntry(recipientAddress, nftID, receivedAttributes)
+            createUserEntry(recipientAddress, nftID)
+        else: # return the NFT to the sender
+            recipientAddress = senderAddress
+            createNFTEntry(recipientAddress, nftID, receivedAttributes)
+            createUserEntry(senderAddress, nftID)
```
ricott1, line 716:
```suggestion
            createUserEntry(recipientAddress, nftID)
```

mjerkov: Yes, a mistake that will be corrected. Thanks!

proposals/lip-0052.md
mjerkov: OK, I will move to the beginning.

ricott1 (approved):

karmacoma (approved):

PR #180, opened 2022-10-24
Fix assets verification in LIP 0055 and 0060

AndreasKendziorra:
The PR includes the following changes: 
- **LIP 0055**: Checking that for each `asset` in `block.assets` there exists a registered module with name `asset.module` is done in the _assets verification_ stage. However, the LIP is [wrongly stating](https://github.com/LiskHQ/lips/blob/4dfe5ef853277dbae3b6c925fa2e98ca20e6460e/proposals/lip-0055.md?plain=1#L248) that it done in the _static validation_ stage. See also [this discussion](https://docs.google.com/spreadsheets/d/1enrrdHS9K2vXboIpEoW7T9-qCjrYbxjXC19NAPAUGK0/edit?disco=AAAAWEUqPJ0). This is fixed in this PR
 - **LIP 0060**: Assets verification is also fixed here. Moreover, the validation rules are named explicitly instead of linking to LIP 0055.


shuse2 (approved):

ricott1 (approved):
It looks good to me, maybe one point we could consider is to order assets per module registration order rather than alphabetically. It does not really matter, and maybe alphabetical is more intuitive, but we could still consider this as it is maybe the more natural order (hooks are called in the registration order, so maybe assets verification should be called in the same order)

AndreasKendziorra:
> It looks good to me, maybe one point we could consider is to order assets per module registration order rather than alphabetically. It does not really matter, and maybe alphabetical is more intuitive, but we could still consider this as it is maybe the more natural order (hooks are called in the registration order, so maybe assets verification should be called in the same order)

@ricott1 I just wonder if the registration order is guaranteed to be same for every node. Especially if a node would uses a software NOT created by our SDK. So, probably not a good idea?

ricott1:
The module registration order has to be the same by protocol, otherwise hooks would not be called in the same order and hence possibly events could be emitted in a different order (result in a different eventsRoot).

proposals/lip-0055.md
```diff
@@ -242,14 +242,16 @@ assetSchema = {
...
 - **Static validation**:
-  - Check that each entry in the assets array has `module` corresponding to a module registered in the chain, while the `data` property has size at most equal to `MAX_ASSET_DATA_SIZE_BYTES`.
+  - Check that for each entry in the assets array the `data` property has size at most equal to `MAX_ASSET_DATA_SIZE_BYTES`.
   - Each module can insert at most one entry in the block assets. Hence, check that each entry has a distinct `module` property.
   - Check that the entries are sorted by lexicographical order of `module`.
+- **Assets verification**:
+  - Check that each entry in the assets array has `module` corresponding to a module registered in the chain.
 
-These validations are performed before the block is processed and without accessing the state. The individual entries are then checked in the assets verification verification stage.
+These validations are performed before the block is processed and without accessing the state. The individual entries are then also checked in the assets verification verification stage.
```
janhack, line 254:
```suggestion
These validations are performed before the block is processed and without accessing the state. The individual entries are then also checked in the assets verification stage.
```

janhack: Above in the LIP we write

> Each module verifies the respective entry in the block assets. If any check fails, the block is discarded and has no further effect.

Would it make sense to just add this here as well as one bullet point and reformulate the sentence above?

janhack:
Regarding

> Checking that for each asset in block.assets there exists a registered module with name asset.module

Do we also need to check this in LIP 60?

janhack (changes requested):
Just some minor point.

proposals/lip-0055.md
AndreasKendziorra: I changed this section. I'm not 100%ly if this is what you suggested.

AndreasKendziorra:
> Regarding
> 
> > Checking that for each asset in block.assets there exists a registered module with name asset.module
> 
> Do we also need to check this in LIP 60?

I think so. I'm still waiting for feedback from @shuse2 in which stage this has to be done ;)

AndreasKendziorra:
> The module registration order has to be the same by protocol, otherwise hooks would not be called in the same order and hence possibly events could be emitted in a different order (result in a different eventsRoot).

@ricott1 I see. Then we also need to add the registration order to LIP 0063. 
@shuse2 If you see any advantage in changing the asset order from lexicographical order to registration order, either from SDK developer or SDK user side, we could change it. Otherwise I would leave it as it is.

ricott1 (approved):
LGTM, maybe to be nit-picking we could still explicitly state that for the genesis block there is no limitation on the size of the `data` property of every asset (just to be super clear).

proposals/lip-0055.md
janhack: Looks all good.

proposals/lip-0060.md
```diff
@@ -100,7 +100,11 @@ The asset schema is the same as defined in [LIP 0055][lip-0055#asset-schema].
 The block assets property is validated as follows:
 
 * _Static validation of the genesis block_:
```
janhack, line 102: I think it would be nice to still mention:

> In contrast to the block processing in [LIP 0055][lip-0055#asset-validation], for the genesis block there is no limitation on the size of the `data` property of every entry.

janhack (approved):
The PR looks good, I just have one (optional) suggestion.

shuse2 (approved):

proposals/lip-0060.md
AndreasKendziorra: done

AndreasKendziorra:
> LGTM, maybe to be nit-picking we could still explicitly state that for the genesis block there is no limitation on the size of the `data` property of every asset (just to be super clear).

@ricott1 done.

proposals/lip-0060.md
```diff
@@ -106,6 +106,8 @@ The block assets property is validated as follows:
   * Check that each entry in the assets array has `module` corresponding to a module registered in the chain.
   * The individual entries are checked by the respective modules.
 
+Note that in contrast to the block processing in [LIP 0055][lip-0055#asset-validation], for the genesis block there is no limitation on the size of the data property of every entry.
```
ricott1, line 109:
```diff
Note that in contrast to the block processing in [LIP 0055][lip-0055#asset-validation], for the genesis block there is no limitation on the size of the data property of every entry.
```

ricott1: Furthermore, I would add the sentence to the first bullet point of the static validation (since the difference is actually there)

ricott1 (changes requested):
Left one comment

proposals/lip-0060.md
AndreasKendziorra: done

ricott1 (approved):

janhack (approved):
Looks all good.

karmacoma (approved):

PR #183, opened 2022-11-09
Update lip 0063

AndreasKendziorra:
The following changes are done on LIP 0063:

- update the asset for the DPoS module in the snapshot block due to reward sharing
- update the asset for the Token module in the snapshot block due to several changes in the token module LIP
- remove references to "BFT module" due to [the separation of application and consensus domain](https://github.com/LiskHQ/lips/pull/177)
- add Dynamic Block Reward module to mainchain
- define module registration order for mainchain
- fix a statement regarding asset usage on mainachain (the random module will add asset values, not the reward module)
- update the configurable constants
- fix statement about asset order
- remove outdated statement about upcoming changes in DPoS (the changes are included in the DPoS LIP by now). 
- fix grammar mistake


proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+  <tr>
+   <td><code>Q96_ZERO</code>
+   </td>
+   <td>bytes
+   </td>
+   <td><code>''</code>
+   </td>
+   <td>The empty byte string, which represents the zero in Q96 representation.
+   </td>
+  </tr>
```
janhack, line 230: In the table above, the constant `LOCAL_ID_LSK` can be removed as it does not seemed to be used in this LIP any more.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
-- [Random](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0046.md)
-- [Fee](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md)
-- [Interoperability](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md)
-- [Token](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md)
-- [Legacy](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md)
+1. [Auth](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0041.md)
+2. [Validators](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0044.md)
+3. [Token](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md)
+4. [Fee](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md)
+5. [Interoperability](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md)
```
janhack, line 243: I remember that Validators and Interoperability have a special status (system modules?) because they cannot be replaced and also that their hooks are always called first. Not sure if this is the case. Would this mean that these modules have to be registered first @shuse2 ?

shuse2: after moving to application/engine, notion `system modules` were removed.
We can register modules in particular order, but we currently don't have any assumption from the implementation side

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+--------------------------------------------------------------------------------------------------------------- | ----------------------------------- | -----------------------------------------------------------------------------------------------------------------------------------------------
+[LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#constants)                            | `LSK_BFT_BATCH_SIZE`                | 103
+[LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants)               | `MAX_TRANSACTIONS_SIZE_BYTES`       | 15,360
+[LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants)               | `MAX_ASSET_DATA_SIZE_BYTES`         | 18 (The Random Module is the only module adding an entry to the assets property whereby the size of the data property will not exceed 18 bytes)
+[LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants)               | `CHAIN_ID`                          | `0x 00 00 00 00`
```
janhack, line 258: This constant is special in that it would be different for the Betanet, Testnet and Mainnet mainchain. All other constants would be the same for the different networks as far as I see.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `LOCKING_PERIOD_VOTES`              | 26,000
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `LOCKING_PERIOD_SELF_VOTES`         | 260,000
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `PUNISHMENT_WINDOW_VOTES`           | 260,000
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `PUNISHMENT_WINDOW_SELF_VOTES`      | 780,000
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `POM_LIMIT_BANNED`                  | 5
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `ROUND_LENGTH`                      | 103
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `BFT_THRESHOLD`                     | 68
```
janhack, line 274: This constant is not used any more in the DPoS module.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `MIN_WEIGHT`                        | 1000 x (10^8)
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `NUMBER_ACTIVE_DELEGATES`           | 101
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `NUMBER_STANDBY_DELEGATES`          | 2
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `TOKEN_ID_DPOS`                     | `TOKEN_ID_LSK = 0x 00 00 00 00 00 00 00 00`
```
janhack, line 278: I would avoid the redundancy in the value definition here which will just create extra maintenance in case the value changes and put  `TOKEN_ID_LSK`.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+[DPoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants)            | `DELEGATE_REGISTRATION_FEE`         | 10 x (10^8)
+[Random Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0046.md#constants-and-config-parameters) | `MAX_LENGTH_VALIDATOR_REVEALS`      | 206
+[Fee Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md#notation-and-constants)             | `MIN_FEE_PER_BYTE`                  | 1000
+[Fee Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md#notation-and-constants)             | `TOKEN_ID_FEE`                      | `TOKEN_ID_LSK = 0x 00 00 00 00 00 00 00 00`
```
janhack, line 282: Same as above.

proposals/lip-0063.md
janhack: Further note that `TOKEN_ID_LSK` is different for the Betanet, Testnet and Mainnet mainchain.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
+[Token Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations)          | `USER_ACCOUNT_INITIALIZATION_FEE`   | 5000000
+[Token Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations)          | `ESCROW_ACCOUNT_INITIALIZATION_FEE` | 5000000
+[Token Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations)          | `FEE_CCM_INIT_USER_ACCOUNT`         | 5000000
+[Token Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations)          | `TOKEN_ID_STORE_INITIALIZATION`     | `TOKEN_ID_LSK = 0x 00 00 00 00 00 00 00 00`
```
janhack, line 286: Same as above.

janhack (changes requested):
I have a few points regarding some constants, otherwise it looks good.

sergeyshemyakov (changes requested):
Nice, just a couple of remarks concerning constants from my side.

proposals/lip-0063.md
```diff
@@ -393,7 +412,7 @@ def createUserSubstoreArray():
             userSubstore.append(userObj)
     # append the legacy reserve account separately
     userSubstore.append(createLegacyReserveAccount())
-    sort userSubstore in the ascending order of (userObj.address + userObj.tokenID)
+    sort userSubstore in the lexicographical order of (userObj.address + userObj.tokenID)
```
sameersubudhi, line 415: Could you please confirm, if the `userObj.address` here is in bytes or Lisk32 format? And, would the `userObj.tokenID` to be considered for the sorting only when there is a clash in the `userObj.address`?

proposals/lip-0063.md
```diff
@@ -374,10 +396,7 @@ def addTokenModuleEntry():
     tokenObj.userSubstore = createUserSubstoreArray()
     tokenObj.supplySubstore = createSupplySubstoreArray()
     tokenObj.escrowSubstore = []
-    tokenObj.availableLocalIDSubstore = {}
-    tokenObj.availableLocalIDSubstore.nextAvailableLocalID = nextLexicographicalOrder(LOCAL_ID_LSK)
-    # Where the function nextLexicographicalOrder returns the next byte value, in lexicographical order
-    # this is equivalent to doing localID + 1 if the byte value is viewed as a big endian integer.
+    tokenObj.supportedTokensSubstore = []
     data = serialization of tokenObj using genesisTokenStoreSchema
```
sameersubudhi, line 400: We do not need to explicitly serialize the data for the genesis assets within the migrator anymore. The `genesis-block create` command seems to take care of it when generating the genesis block blob. We should adjust this accordingly.

Ref implementations:
-  https://github.com/LiskHQ/lisk-sdk/blob/74f59e1704c8c02db76d8f30671b91a0bb114bcc/examples/dpos-mainchain/config/default/genesis_assets.json
- https://github.com/LiskHQ/lisk-core/blob/1a01bae01bc876d4c5a9fe2006463d9488d2571e/config/devnet/genesis_assets.json

proposals/lip-0063.md
AndreasKendziorra: @shuse2 could you answer this one?

proposals/lip-0063.md
AndreasKendziorra: yes, but I would leave this point open until [this discussion](https://github.com/LiskHQ/discussions/discussions/63) is solved. As this issue is spread over many LIPs, I would also prefer to fix it in a separate PR in which we fix all LIPs at the same time.

proposals/lip-0063.md
AndreasKendziorra: Right, thanks. But is this threshold constant now defined in another LIP? I guess this threshold must be set somewhere.

proposals/lip-0063.md
```diff
@@ -218,50 +218,72 @@ ffffff
...
-[Random Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0046.md#constants-and-config-parameters) | `MAX_LENGTH_VALIDATOR_REVEALS` | 206
-[Fee Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md#notation-and-constants)             | `MIN_FEE_PER_BYTE`             | 1000
-[Fee Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md#notation-and-constants)             | `TOKEN_ID_FEE`                 | `TOKEN_ID_LSK = 0x 00 00 00 00 00 00 00 00`
+**Module/LIP**                                                                                                  | **Constant Name**                   | **Value**
+--------------------------------------------------------------------------------------------------------------- | ----------------------------------- | -----------------------------------------------------------------------------------------------------------------------------------------------
+[LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#constants)                            | `LSK_BFT_BATCH_SIZE`                | 103
```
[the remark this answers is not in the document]
AndreasKendziorra: I'm not sure how you mean this. Do you mean removing this line? This is constant is still a configurable constant, despite not belonging to a module anymore.
Note that I was assuming the changes of [this PR](https://github.com/LiskHQ/lips/pull/177).

proposals/lip-0063.md
AndreasKendziorra: `userObj.address` is in bytes. And yes, sorting by `tokenID` is only done if there are two `userObj` instances with the same address.

proposals/lip-0063.md
AndreasKendziorra: So the implementation of the migrator does need to strictly follow the implementation details of the pseudo code of  this LIP. The pseudo code is only here to specify how the snapshot looks like. But how you get there is not important.

proposals/lip-0063.md
AndreasKendziorra: I removed the two constants. 

Regarding `TOKEN_ID_LSK`: I would leave this point open until https://github.com/LiskHQ/discussions/discussions/63 is solved. As this issue is spread over many LIPs, I would also prefer to fix it in a separate PR in which we fix all LIPs at the same time.

proposals/lip-0063.md
AndreasKendziorra: Good point. Changed

proposals/lip-0063.md
AndreasKendziorra: changed

proposals/lip-0063.md
AndreasKendziorra: changed

sameersubudhi (approved):
LGTM 👍

sergeyshemyakov (approved):
Looks good now.

proposals/lip-0063.md
janhack: Okay, sounds good.

proposals/lip-0063.md
janhack: The comment can be resolved from my side.

proposals/lip-0063.md
janhack: Okay, can be resolved then.

proposals/lip-0063.md
janhack: It is instead computed in the DPoS module, see the computation before the call of `setValidatorParams`. The comment can be resolved from my side.

janhack (approved):
Looks good. All my remaining comments can be resolved.

karmacoma (approved):

PR #184, opened 2022-11-15
Update LIP 61 due application-engine separation

janhack:
In this PR makes the following changes:

- Any reference to the BFT module in LIP 0061 is removed (note that due to the application-engine separation, there is no BFT module any more). Previous function calls `bftModule.fct` are changed to `fct`. A list of all functions from [LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md) that are used in the specifications is given at the beginning of the specification section. 
- Any call to the Validator module is removed. Instead, the necessary information regarding validators is obtained from the BFT store via functions defined in [LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md) . This is possible as the BFT store now also stores validator keys.
- The network identifier is removed from LIP 0061 and replaced with the chain ID, see [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md) for background on chain identfiers.
- There is now two types of objects for certificates, a unsigned certificate (without properties `aggregationBits` and `signature`) and a signed certificate. All properties in both associated schema are required in accordance with [LIP 0064](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0064.md). All functions are updated with the respective return types.
- The block-level logic is defined more clearly, adjusting for the update of [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md) and [LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md) and the new stages "Before Application Processing" and "After Application Processing". In particular, it is made more clear at what step in the block creation the function `selectAggregateCommit` should be called and at what step in the block processing the function `verifyAggregateCommit` should be called. For this, some additional clarity is also added to logic executed in "Before Application Processing" and "After Application Processing" in LIP 0058.
- Approach 2 for computing certificates from single commits is updated and now relies on past BFT parameters instead of calls to the Validators module.
- Type hinting and the appropriate types are introduced and some syntax is simplified.

### DIFF
--- a/proposals/lip-0061.md
+++ b/proposals/lip-0061.md
@@ -61,7 +61,7 @@ It further is important that the protocol provides enough incentives for validat
 
 The goal of the P2P gossip mechanism for single commits is to  ensure that the validators have the necessary data to generate aggregate commits which satisfy the chain of trust (see section below) as quickly and reliably as possible. In particular, we want to share single commits as fast as possible while being robust in case of adverse network conditions, such as a temporary network split, a longer time without finalized blocks or a significant number of validators not sharing single commits. The design decisions and choice of parameters explained in the list below are based on this overarching goal.
 
-- For the `COMMIT_RANGE_STORED+1` most recent blocks of height at most `getBFTHeights().maxHeightPrecommitted` ([getBFTHeights][lip-0058#getbftheights] is defined in LIP 0058) we store and gossip all single commits, while for older blocks we only store single commits for blocks for which the `validatorsHash` property is different from that of its parent block. The reason is that certificates or aggregate commits are only required to be generated for blocks that authenticate a change of [BFT parameters][lip-0058], see Section "Chain of Trust" for details. For more recent blocks we want to create and gossip single commits as soon as possible so we store and gossip them for all blocks.
+- For the `COMMIT_RANGE_STORED+1` most recent blocks of height at most `getBFTHeights().maxHeightPrecommitted` ([`getBFTHeights`][lip-0058#getbftheights] is defined in LIP 0058) we store and gossip all single commits, while for older blocks we only store single commits for blocks for which the `validatorsHash` property is different from that of its parent block. The reason is that certificates or aggregate commits are only required to be generated for blocks that authenticate a change of [BFT parameters][lip-0058], see Section "Chain of Trust" for details. For more recent blocks we want to create and gossip single commits as soon as possible so we store and gossip them for all blocks.
 - For blocks of height at most `getBFTHeights().maxHeightCertified` we do not need to store any single commits, as an aggregate commit has been already generated for the height `getBFTHeights().maxHeightCertified`.
 - We gossip single commits every `BLOCK_TIME/2` seconds. This means single commits are shared at a higher frequency than blocks so that ideally after a block is finalized, an aggregate commit for that block is included in the chain only one or two blocks later.
 - An aggregate commit requires single commits by a certain subset of the active validators, depending on the value of the certificate threshold described in the next section. One P2P message can contain up to `2*numActiveValidators` single commits, where `numActiveValidators` is the number of active validators at that height. Such a message can therefore contain enough single commits to create two aggregate commits. Together with the gossiping frequency of `BLOCK_TIME/2` seconds this allows to share single commits four times faster than blocks are created, allowing for a significant buffer.
@@ -77,7 +77,7 @@ During the commit phase an additional threshold is used, called **certificate th
 * Commit messages are aggregated to aggregate commits, which are then included in blocks. The sum of BFT weights of the validators signing an aggregate commit has to be at least the certificate threshold value. Here the validators are those that are active at the height of the certificate and the BFT weights are the corresponding weights at that height.
 * When submitting a certificate via a cross-chain update transaction, the weights of all signers have to be above the certificate threshold value that is currently known to the other chain, see [LIP 0053][lip-0053] for details.
 
-The certificate threshold is stored as part of the BFT store in the consensus domain, see [LIP 0058][lip-0058] for details. The value of this parameters is set by the DPoS or PoA module via the [setValidatorParams][lip-0044#setValidatorParams] of the Validators module and then forwarded to the consensus domain. Both for [Lisk DPoS][lip-0057] and [Lisk PoA][lip-0047], we propose to use the same values for the certificate threshold as for the precommit threshold. This means that the same threshold is required for finality as for cross-chain certification.
+The certificate threshold is stored as part of the BFT store in the consensus domain, see [LIP 0058][lip-0058] for details. The value of this parameters is set by the DPoS or PoA module via the [`setValidatorParams`][lip-0044#setValidatorParams] of the Validators module and then forwarded to the consensus domain. Both for [Lisk DPoS][lip-0057] and [Lisk PoA][lip-0047], we propose to use the same values for the certificate threshold as for the precommit threshold. This means that the same threshold is required for finality as for cross-chain certification.
 
 In general, for a height `h`, the certificate threshold could be chosen within the following range:
 
@@ -106,9 +106,9 @@ Intuitively, the chain of trust property means that for any validator change a s
 
 _Figure 1: Example of a sequence of three certificates satisfying the chain of trust._
 
-For maintaining interoperability via certification, it is therefore crucial that the chain of trust is always maintained. Every block and also any certificate derived from a block has a `validatorsHash` property. This property is computed from the BLS keys of the active validators, their BFT weights and the certificate threshold after the block is applied, see the function [computeValidatorsHash][lip-0058#computevalidatorshash] in LIP 0058 for details. As the active validators are the same within one round, only blocks which are the last block of a round may have a different `validatorsHash` property than their parent block. This means that the chain of trust property is satisfied if the certificate generation mechanism ensures that a certificate is generated for all blocks for which the `validatorsHash` property is distinct from the `validatorsHash` property of the parent block. This way there is a certificate authenticating any validator transition that happened in the chain.
+For maintaining interoperability via certification, it is therefore crucial that the chain of trust is always maintained. Every block and also any certificate derived from a block has a `validatorsHash` property. This property is computed from the BLS keys of the active validators, their BFT weights and the certificate threshold after the block is applied, see the function [`computeValidatorsHash`][lip-0058#computevalidatorshash] in LIP 0058 for details. As the active validators are the same within one round, only blocks which are the last block of a round may have a different `validatorsHash` property than their parent block. This means that the chain of trust property is satisfied if the certificate generation mechanism ensures that a certificate is generated for all blocks for which the `validatorsHash` property is distinct from the `validatorsHash` property of the parent block. This way there is a certificate authenticating any validator transition that happened in the chain.
 
-The certificate generation specified in this LIP therefore generates an aggregate commit for any block at height `h` for which  `existBFTParameters(h+1)` returns `True`, where [existBFTParameters][lip-0058#existbftparameters] is defined in LIP 0058. This means that at height `h+1` the [BFT store][lip-0058#bft-parameters] contains new and possibly different BFT parameters that need to be authenticated by the block at height `h`. As already mentioned, from the on-chain data, i.e., the aggregate commit and referenced block header, the corresponding certificates can be computed. Hence, the mechanism specified in this LIP guarantees that the generated certificates satisfy the chain of trust property.
+The certificate generation specified in this LIP therefore generates an aggregate commit for any block at height `h` for which  `existBFTParameters(h+1)` returns `True`, where [`existBFTParameters`][lip-0058#existbftparameters] is defined in LIP 0058. This means that at height `h+1` the [BFT store][lip-0058#bft-parameters] contains new and possibly different BFT parameters that need to be authenticated by the block at height `h`. As already mentioned, from the on-chain data, i.e., the aggregate commit and referenced block header, the corresponding certificates can be computed. Hence, the mechanism specified in this LIP guarantees that the generated certificates satisfy the chain of trust property.
 
 Note that if possible, also aggregate commits for blocks that are not at the end of the round are created and added to blocks so that a certificate is created as soon as a block is finalized and it is not required to wait for a change of BFT parameters.
 
@@ -118,10 +118,10 @@ Note that if possible, also aggregate commits for blocks that are not at the end
 
 In this LIP, we will frequently use the following functions specified in LIP 0058 without reference to the function definition:
 
-- [existBFTParameters][lip-0058#existbftparameters]
-- [getBFTHeights][lip-0058#getbftheights]
-- [getBFTParameters][lip-0058#getbftparameters]
-- [getNextHeightBFTParameters][lip-0058#getnextheightbftparameters]
+- [`existBFTParameters`][lip-0058#existbftparameters]
+- [`getBFTHeights`][lip-0058#getbftheights]
+- [`getBFTParameters`][lip-0058#getbftparameters]
+- [`getNextHeightBFTParameters`][lip-0058#getnextheightbftparameters]
 
 ### Constants
 
@@ -663,7 +663,7 @@ def getNextCertificateFromAggregateCommits(lastCertifiedHeight: uint32) -> Certi
 ```
 
 ```python
-def checkChainOfTrust(lastValidatorsHash: bytes, blsKeyToBFTWeight: dict[bytes, uint64], lastCertificateThreshold: uint32, aggregateCommit: AggregateCommit) -> bool:
+def checkChainOfTrust(lastValidatorsHash: bytes, blsKeyToBFTWeight: dict[bytes, uint64], lastCertificateThreshold: uint64, aggregateCommit: AggregateCommit) -> bool:
     blockHeader = block header at height aggregateCommit.height - 1
     # Certificate signers and certificate threshold for aggregateCommit are those authenticated by the last certificate
     if lastValidatorsHash == blockHeader.validatorsHash:
--- a/proposals/lip-0061.md
+++ b/proposals/lip-0061.md
@@ -136,10 +136,11 @@ In this LIP, we will frequently use the following functions specified in LIP 005
 
 | Name               | Type   | Validation | Description |
 |--------------------|--------|------------|-------------|
+| `AggregateCommit`  | object | Must follow `aggregateCommitSchema` schema. | An object representing an aggregate commit. |
 | `BlockHeader`      | object | Must follow `blockHeaderSchema` schema defined in [LIP 0055][lip-0055]. | An object representing a block header. |
 | `Certificate`      | object | Must follow `certificateSchema` schema. | An object representing a certificate. |
 | `SingleCommit`     | object | Must follow `singleCommitSchema` schema. | An object representing a single commit. |
-| `AggregateCommit`  | object | Must follow `aggregateCommitSchema` schema. | An object representing an aggregate commit. |
+| `Validator`        | object | Must follow `validatorSchema` schema defined in [LIP 0058][lip-0058]. | An object representing a validator with properties `address`, `bftWeight`, `generatorKey` and `blsKey`. |
 
 ### Certificate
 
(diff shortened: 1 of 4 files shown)


proposals/lip-0061.md
```diff
@@ -61,12 +61,12 @@ It further is important that the protocol provides enough incentives for validat
...
-- For blocks of height at most `bftModule.getBFTHeights().maxHeightCertified` we do not need to store any single commits, as an aggregate commit has been already generated for the height `bftModule.getBFTHeights().maxHeightCertified`.
+- For the `COMMIT_RANGE_STORED+1` most recent blocks of height at most `getBFTHeights().maxHeightPrecommitted` ([getBFTHeights][lip-0058#getbftheights] is defined in LIP 0058) we store and gossip all single commits, while for older blocks we only store single commits for blocks for which the `validatorsHash` property is different from that of its parent block. The reason is that certificates or aggregate commits are only required to be generated for blocks that authenticate a change of [BFT parameters][lip-0058], see Section "Chain of Trust" for details. For more recent blocks we want to create and gossip single commits as soon as possible so we store and gossip them for all blocks.
```
ricott1, line 64:
```suggestion
- For the `COMMIT_RANGE_STORED+1` most recent blocks of height at most `getBFTHeights().maxHeightPrecommitted` ([`getBFTHeights`][lip-0058#getbftheights] is defined in LIP 0058) we store and gossip all single commits, while for older blocks we only store single commits for blocks for which the `validatorsHash` property is different from that of its parent block. The reason is that certificates or aggregate commits are only required to be generated for blocks that authenticate a change of [BFT parameters][lip-0058], see Section "Chain of Trust" for details. For more recent blocks we want to create and gossip single commits as soon as possible so we store and gossip them for all blocks.
```

proposals/lip-0061.md
```diff
@@ -77,7 +77,7 @@ During the commit phase an additional threshold is used, called **certificate th
...
 
-The certificate threshold is stored as part of the BFT Parameters substore of the BFT module, see [LIP 0058][lip-0058] for details. The value of this parameters is set by the DPoS or PoA module via the `setBFTParameters` functions. Both for [Lisk DPoS][lip-0057] and [Lisk PoA][lip-0047], we propose to use the same values for the certificate threshold as for the precommit threshold. This means that the same threshold is required for finality as for cross-chain certification.
+The certificate threshold is stored as part of the BFT store in the consensus domain, see [LIP 0058][lip-0058] for details. The value of this parameters is set by the DPoS or PoA module via the [setValidatorParams][lip-0044#setValidatorParams] of the Validators module and then forwarded to the consensus domain. Both for [Lisk DPoS][lip-0057] and [Lisk PoA][lip-0047], we propose to use the same values for the certificate threshold as for the precommit threshold. This means that the same threshold is required for finality as for cross-chain certification.
```
ricott1, line 80:
```suggestion
The certificate threshold is stored as part of the BFT store in the consensus domain, see [LIP 0058][lip-0058] for details. The value of this parameters is set by the DPoS or PoA module via the [`setValidatorParams`][lip-0044#setValidatorParams] of the Validators module and then forwarded to the consensus domain. Both for [Lisk DPoS][lip-0057] and [Lisk PoA][lip-0047], we propose to use the same values for the certificate threshold as for the precommit threshold. This means that the same threshold is required for finality as for cross-chain certification.
```

proposals/lip-0061.md
```diff
@@ -106,25 +106,41 @@ Intuitively, the chain of trust property means that for any validator change a s
...
+For maintaining interoperability via certification, it is therefore crucial that the chain of trust is always maintained. Every block and also any certificate derived from a block has a `validatorsHash` property. This property is computed from the BLS keys of the active validators, their BFT weights and the certificate threshold after the block is applied, see the function [computeValidatorsHash][lip-0058#computevalidatorshash] in LIP 0058 for details. As the active validators are the same within one round, only blocks which are the last block of a round may have a different `validatorsHash` property than their parent block. This means that the chain of trust property is satisfied if the certificate generation mechanism ensures that a certificate is generated for all blocks for which the `validatorsHash` property is distinct from the `validatorsHash` property of the parent block. This way there is a certificate authenticating any validator transition that happened in the chain.
```
ricott1, line 109:
```suggestion
For maintaining interoperability via certification, it is therefore crucial that the chain of trust is always maintained. Every block and also any certificate derived from a block has a `validatorsHash` property. This property is computed from the BLS keys of the active validators, their BFT weights and the certificate threshold after the block is applied, see the function [`computeValidatorsHash`][lip-0058#computevalidatorshash] in LIP 0058 for details. As the active validators are the same within one round, only blocks which are the last block of a round may have a different `validatorsHash` property than their parent block. This means that the chain of trust property is satisfied if the certificate generation mechanism ensures that a certificate is generated for all blocks for which the `validatorsHash` property is distinct from the `validatorsHash` property of the parent block. This way there is a certificate authenticating any validator transition that happened in the chain.
```

proposals/lip-0061.md
```diff
@@ -106,25 +106,41 @@ Intuitively, the chain of trust property means that for any validator change a s
...
+The certificate generation specified in this LIP therefore generates an aggregate commit for any block at height `h` for which  `existBFTParameters(h+1)` returns `True`, where [existBFTParameters][lip-0058#existbftparameters] is defined in LIP 0058. This means that at height `h+1` the [BFT store][lip-0058#bft-parameters] contains new and possibly different BFT parameters that need to be authenticated by the block at height `h`. As already mentioned, from the on-chain data, i.e., the aggregate commit and referenced block header, the corresponding certificates can be computed. Hence, the mechanism specified in this LIP guarantees that the generated certificates satisfy the chain of trust property.
```
ricott1, line 111:
```suggestion
The certificate generation specified in this LIP therefore generates an aggregate commit for any block at height `h` for which  `existBFTParameters(h+1)` returns `True`, where [`existBFTParameters`][lip-0058#existbftparameters] is defined in LIP 0058. This means that at height `h+1` the [BFT store][lip-0058#bft-parameters] contains new and possibly different BFT parameters that need to be authenticated by the block at height `h`. As already mentioned, from the on-chain data, i.e., the aggregate commit and referenced block header, the corresponding certificates can be computed. Hence, the mechanism specified in this LIP guarantees that the generated certificates satisfy the chain of trust property.
```

proposals/lip-0061.md
```diff
@@ -106,25 +106,41 @@ Intuitively, the chain of trust property means that for any validator change a s
...
 
 Note that if possible, also aggregate commits for blocks that are not at the end of the round are created and added to blocks so that a certificate is created as soon as a block is finalized and it is not required to wait for a change of BFT parameters.
 
 ## Specification
 
 ### Notation
 
-In this LIP, we write `bftModule.fct` for a call to the function `fct` defined in the [BFT module][lip-0058] and analogously `validatorsModule.fct` for a call to the function `fct` defined in the [Validators module][lip-0044].
+In this LIP, we will frequently use the following functions specified in LIP 0058 without reference to the function definition:
+
```
ricott1, line 120: I think we should add missing ` to all function names, but since you didn't now I wonder if it was a deliberate choice?

proposals/lip-0061.md
```diff
@@ -213,11 +233,12 @@ The certificate signature as byte array.
 ###### Execution
 
 ```python
-signCertificate(sk, networkIdentifier, c):
-    remove the aggregationBits and signature property from c
-    message = serialization of c according to LIP 0027
+def signCertificate(sk: bytes, chainID: bytes, c: Certificate) -> bytes:
+    c.aggregationBits = EMPTY_BYTES
+    c.signature = EMPTY_BYTES
+    message = encode(certificateSchema, c)
```
ricott1, line 239: We are changing the protocol here, as now the serialization still contains the empty field from the schema. I think this should be fine.

proposals/lip-0061.md
```diff
@@ -140,7 +156,9 @@ certificateSchema = {
         "height",
         "timestamp",
         "stateRoot",
-        "validatorsHash"
+        "validatorsHash",
```
ricott1, line 159: So if I understant correctly, rather than having a new type 'unsigned certificate', you set `aggregationBits` and `signature` to empty bytes. I guess this is fine in this case, since there is no concept of ID and the signature should still be fine, but one drawback is that we can not impose the length of the signature in the schema (which we do in other places) and that we are changing the protocol.

ricott1 (changes requested):
Minor formatting comments on setting font of function names. 
One comment about using `EMPTY_BYTES` for the certificate fields that were previously non-required. Another option, as discussed, could be to introduce another schema without the required fields, similar to what we do for block headers.

proposals/lip-0061.md
```diff
@@ -534,15 +561,9 @@ verifyAggregateCommit(aggregateCommit):
     c = computeCertificateFromBlockHeader(blockHeader1)
     c.aggregationBits = aggregateCommit.aggregationBits
     c.signature = aggregateCommit.certificateSignature
-    networkIdentifier = network identifier of the chain
-    bftParams = bftModule.getBFTParameters(aggregateCommit.height)
-    validatorKeys = list of BLS public keys of validators in bftParams.validators obtained via
-                    validatorsModule.getValidatorAccount
-    sort validatorKeys lexicographically
-    weights = array of validator BFT weights obtained from bftParams.validators such that
-              weights[i] is the BFT weight of the validator with the BLS key validatorKeys[i]
```
ishantiw, line 543: Does it make any difference if we sort `validatorKeys` lexicographically here or we do it in `verifyAggregateCertificateSignature`? or is it just more suitable to be in `verifyAggregateCertificateSignature`

ishantiw (approved):
LGTM, a small question though

proposals/lip-0061.md
```diff
@@ -642,7 +663,7 @@ getNextCertificateFromAggregateCommits(lastCertifiedHeight):
 ```
 
 ```python
-checkChainOfTrust(lastValidatorsHash, blsKeyToBFTWeight, lastCertificateThreshold, aggregateCommit):
+def checkChainOfTrust(lastValidatorsHash: bytes, blsKeyToBFTWeight: dict[bytes, uint64], lastCertificateThreshold: uint32, aggregateCommit: AggregateCommit) -> bool:
```
AndreasKendziorra, line 666:
```suggestion
def checkChainOfTrust(lastValidatorsHash: bytes, blsKeyToBFTWeight: dict[bytes, uint64], lastCertificateThreshold: uint64, aggregateCommit: AggregateCommit) -> bool:
```

proposals/lip-0061.md
```diff
@@ -690,17 +711,18 @@ The function returns a valid certificate object that can be submitted to blockch
 For readability, we describe the logic using two functions, the main function `getNextCertificateFromSingleCommits` and the auxiliary function `computeEligibleSingleCommits`.
 
 ```python
-getNextCertificateFromSingleCommits(lastCertifiedHeight):
+def getNextCertificateFromSingleCommits(lastCertifiedHeight: uint32) -> Certificate:
     lastValidatorsHash  = validatorsHash property of block header at height lastCertifiedHeight
     lastCertifiedValidators = validatorsHashPreimage[lastValidatorsHash].validators
```
AndreasKendziorra, line 716: `lastCertifiedValidators` is not used anymore

proposals/lip-0061.md
janhack: Fixed.

proposals/lip-0061.md
janhack: Fixed.

proposals/lip-0061.md
janhack: Fixed.

proposals/lip-0061.md
janhack: Fixed.

proposals/lip-0061.md
janhack: Yep, makes sense. I fixed it everywhere.

proposals/lip-0061.md
janhack: This is the same comment as https://github.com/LiskHQ/lips/pull/184#discussion_r1026155967 below, I assume.

proposals/lip-0061.md
janhack: In terms of protocol, it is the same. The reason I changed it in the specifications is that in `aggregateSingleCommits` we also do the sorting within the function and I think it is more robust to sort within the function than to assume that the input parameter is sorted in a certain way.

ishantiw (commented):
Need to update this to chainID
https://github.com/LiskHQ/lips/blob/c864dc8d108efe63e2835b10648cadd6bc826bd6/proposals/lip-0061.md?plain=1#L257

https://github.com/LiskHQ/lips/blob/c864dc8d108efe63e2835b10648cadd6bc826bd6/proposals/lip-0061.md?plain=1#L283

proposals/lip-0061.md
janhack: As discussed, I now introduce an unsigned certificate schema and type.

proposals/lip-0061.md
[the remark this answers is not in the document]
janhack: Fixed both.

proposals/lip-0061.md
```diff
@@ -180,17 +198,19 @@ certificateSchema = {
...
     c.validatorsHash = blockHeader.validatorsHash
+    c.aggregationBits = EMPTY_BYTES
+    c.signature = EMPTY_BYTES
     return c
 ```
 
-Note that the two properties `aggregationBits` and `signature` are not required in the schema and are not set in the above function because these properties are removed when signing certificates (see section below).
+Note that as all certificate properties are required, we initialize `aggregationBits` and `signature` with the default value (empty bytes).
 
 #### Signature Computation and Validation
```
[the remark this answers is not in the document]
janhack: Good point, I updated the reference.

proposals/lip-0061.md
```diff
@@ -264,13 +285,17 @@ The functions returns `True` if and only if the aggregate certificate signature
 ###### Execution
 
 ```python
-verifyAggregateCertificateSignature(keysList, weights, threshold, networkIdentifier, c):
+def verifyAggregateCertificateSignature(validatorList: list[Validator], threshold: uint64, chainID: bytes, c: Certificate) -> bool:
```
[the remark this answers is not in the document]
janhack: 1. @ricott1 Could you incorporate it in the Interoperability updates?
2. Note that this is a stateless function that could be implemented as a library. I think in terms of specifications, it is fine to just reference the function definition. In the implementation it could mean that some code is duplicated or written as a library.

proposals/lip-0061.md
```diff
@@ -264,13 +285,17 @@ The functions returns `True` if and only if the aggregate certificate signature
...
     aggregateSignature = c.signature
     aggregationBits = c.aggregationBits
-    remove the aggregationBits and signature property from c
-    message = serialization of c according to LIP 0027
+    c.aggregationBits = EMPTY_BYTES
+    c.signature = EMPTY_BYTES
+    message = encode(certificateSchema, c)
     tag = MESSAGE_TAG_CERTIFICATE
-    return verifyWeightedAggSig(keysList, aggregationBits, aggregateSignature, tag, networkIdentifier, weights, threshold, message) == VALID
+    return verifyWeightedAggSig(keysList, aggregationBits, aggregateSignature, tag, chainID, bftWeights, threshold, message)
```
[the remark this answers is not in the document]
janhack: Those return values have been removed and those function now return boolean values.

proposals/lip-0061.md
```diff
@@ -493,38 +519,39 @@ selectAggregateCommit():
...
+    aggregateCommit.aggregationBits =  EMPTY_BYTES
+    aggregateCommit.certificateSignature =  EMPTY_BYTES
     return aggregateCommit
 ```
 
-#### Block Verification
+### Block Processing
 
-As part of the verification of a block header `blockHeader` as described in [LIP 0055][lip-0055], the property `aggregateCommit` has to be verified, i.e., the following function has to return `True`.
+As described in the ["before application processing" stage of the block processing in LIP 0055][lip-0055#aggregate-commit], the property block header property `aggregateCommit` is verified using the function `verifyAggregateCommit` defined below. Note that it is important that the validation defined below is performed before the BFT store is updated, i.e., before the "before application processing" logic defined in [LIP 0058][lip-0058#before-application-processing-1] is executed.
```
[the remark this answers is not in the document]
janhack: Note that actually LIP 61 only defines functions that are used in these steps, but references to other LIPs to when these functions are called. At the moment the split is as follows:

- LIP 55:  all block level verifications
- LIP 58: logic to update the BFT store

proposals/lip-0061.md
```diff
@@ -493,38 +519,39 @@ selectAggregateCommit():
...
 
 ```python
-verifyAggregateCommit(aggregateCommit):
+def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
```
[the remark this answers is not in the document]
janhack: Good observation, fixed it.

proposals/lip-0061.md
janhack: Good point. After thinking about approach 2 and the required data more, I now only use the BFT parameters here and in the bullet point list about the required data, I am more explicit about the fact that past BFT parameters are needed.

janhack: @ishantiw As this turned out to be a bit of a bigger change for this function and the required data, it would be nice if you could also have another look.

proposals/lip-0061.md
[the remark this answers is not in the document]
ricott1: yes.

ricott1 (approved):

proposals/lip-0061.md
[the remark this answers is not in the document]
AndreasKendziorra: Fine for me (although I personally prefer when PRs are complete and do not break anything). I leave it open for @ricott1

proposals/lip-0061.md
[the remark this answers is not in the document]
AndreasKendziorra: ah, true. I missed that.

proposals/lip-0061.md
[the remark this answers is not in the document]
AndreasKendziorra: Ok, I see.

proposals/lip-0061.md
```diff
@@ -689,8 +720,7 @@ In this section, we describe how to compute the certificate of largest height th
...
-- A function to obtain the address of a validator in blockchain *A* from its BLS key. We assume that that this functionality is provided by the function `getAddressFromBLSKey` of the [Validators module][lip-0044].
+- In Step 7 of the ["before application processing" stage][lip-0058#before-application-processing-1] defined in LIP 0058, no longer needed entries in the [BFT Parameters substore][lip-0058#bft-parameters] are deleted. However, for the approach described here, we may need BFT parameters that would be deleted in this stage. Hence, the implementation should be able to provide past BFT parameters such that it is always possible to obtain them via `getBFTParameters(lastCertifiedHeight + 1)`. For this is is sufficient if from a height `h` onwards, where `h` is the largest integer such that `h <= lastCertifiedHeight + 1`, all BFT parameters in the BFT parameters substore are maintained. In particular, the implementation could allow to maintain all past BFT parameters for nodes that allow to obtain certificates with the approach described here.
```
AndreasKendziorra, line 723:
```suggestion
- In Step 7 of the ["before application processing" stage][lip-0058#before-application-processing-1] defined in LIP 0058, no longer needed entries in the [BFT Parameters substore][lip-0058#bft-parameters] are deleted. However, for the approach described here, we may need BFT parameters that would be deleted in this stage. Hence, the implementation should be able to provide past BFT parameters such that it is always possible to obtain them via `getBFTParameters(lastCertifiedHeight + 1)`. For this it is sufficient if from a height `h` onwards, where `h` is the largest integer such that `h <= lastCertifiedHeight + 1`, all BFT parameters in the BFT parameters substore are maintained. In particular, the implementation could allow to maintain all past BFT parameters for nodes that allow to obtain certificates with the approach described here.
```

proposals/lip-0061.md
AndreasKendziorra: can be resolved from my side. But I leave it open for @ishantiw

proposals/lip-0061.md
AndreasKendziorra: can be resolved

proposals/lip-0061.md
ishantiw: Looks good to me, looks more simplified now

ishantiw (approved):

proposals/lip-0061.md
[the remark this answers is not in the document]
janhack: I will resolve this comment then.

ricott1 (approved):

AndreasKendziorra (approved):

karmacoma (approved):

PR #182, opened 2022-11-02
Improve fee system

ricott1:
Improve the fee system to handle fee payment and assignment consistently within the Fee module.

- Update LIP0048 to include new (cross-chain) command processing stages and `payFee` method.
- Update LIP0043 to remove extra registration fee and use `payFee`.
- Update LIP0049 to include methods to serialize/deserialize CCMs.
- Update LIP0053 to set the ccm processing context and initialize the relayer token account.
- Update LIP0051 to remove initialization fees and automatically initialize non-existing accounts and use `payFee`.

PR #185, opened 2022-11-16
Add LIP "Introduce reward sharing mechanism"

gkoumout:
This pull request introduces reward sharing mechanism in the Lisk protocol. 

## New LIP

The LIP "Introduce reward sharing mechanism" is proposed. The draft has been posted in the research forum for 3 weeks, see [here](https://research.lisk.com/t/introduce-reward-sharing-mechanism/386)

## Updates in DPoS Module

The reward sharing mechanism is part of the DPoS module. Therefore, to make it functional, this PR modifies the DPoS module store and commands to support reward sharing. Moreover, to keep the DPoS module updated/consistent with recent protocol changes, the following edits are performed:

 - **Make the DPoS module compatible with the [new fee system](https://github.com/LiskHQ/lips/pull/182).** In particular, this requires remove the `delegateRegistrationFee` parameter from the delegate registration command params. Instead, during execution, the function`Fee.payFee()` is called. 
 - **Make the DPoS module compatible with the [recent updates on engine/application separation](https://github.com/LiskHQ/lips/pull/177),** where BFT module does not exist anymore and some of its functionalities are included in the validators module. In particular, this includes removing any references to BFT module and updating all calls to functions of BFT module or Validators module that are renamed.
 - **Make the DPoS compatible with the new [dynamic block rewards system](https://github.com/LiskHQ/lips-staging/pull/148),** which was designed in parallel with the reward sharing mechanism.  In particular this includes handling BFT weights differently: instead of bft weight of 1 for active and 0 for standby delegates, the BFT weight of active delegates is now proportional to the delegate weight. Moreover, to ensure that no delegate gets too much power, a capping function is applied to ensure that no delegate has more than X% of the total BFT weight. 
 - **Updates suggested from development teams**. These include:
     - Remove dependency from reward module, by assigning the PoM reward in `TOKEN_ID_DPOS` instead of `TOKEN_ID_REWARD` and get this amount from the locked self-vote amount of the delegate who misbehaved.
     - Add an endpoint `getDPoSTokenID()` which returns the token used for voting.
     - Do not include `height` as a parameter in any event, since it is included in event params.
     - Make the event names consistent.
     - Minor technical changes/typos.


proposals/lip-0057.md
```diff
@@ -1298,6 +1453,20 @@ def getDelegate(address: Address)-> DelegateStoreObject:
...
+
+Removes the banning for a delegate, given that the reason of banning was not due to misbehaviors. For example, this applies in cases the delegate got banned due to inactivity or not registering valid BLS keys.
+
+```python
+def unbanDelegate(address: Address)-> None:
+    if there does not exist an entry in delegate substore with key == address:
+        raise Exception('No registered delegate with the specified address.')
+    if  delegateStore(address).isBanned == False:
+        raise Exception('The specified delegate is not banned.')
+    if len(delegateStore(address).pomHeights) < POM_LIMIT_BANNED and Validators.getValidatorAccount(address).blsKey != INVALID_BLS_KEY:
```
janhack, line 1466: What is the reason for checking `Validators.getValidatorAccount(address).blsKey != INVALID_BLS_KEY` here? Is the reasoning that you should only be able to unban if you previously updated the BLS keys?

In a sense, what we want is also that you can only unban yourself if beforehand your BLS key was `INVALID_BLS_KEY`. This is not checked here, but in the legacy module. 

I would maybe try to avoid this coupling and have rather a simple, clean unbanning API that assumes it is used responsibly and where the conditions are checked outside the scope of this function. In particular, no knowledge of the regenesis is then required here.

proposals/lip-0057.md
```diff
@@ -160,10 +173,10 @@ The store keys and values of the DPoS store are set as follows:
 
 ##### Store Prefix, Store Key, and Store Value
 
-* The store prefix is set to `STORE_PREFIX_VOTER`.
+* The store prefix is set to `SUBSTORE_PREFIX_VOTER`.
 * Each store key is a `ADDRESS_LENGTH`-byte `address`, representing a user address.
 * Each store value is the serialization of an object following `voterStoreSchema`.
-* Notation: For the rest of this proposal let `voterStore(address)` be the object stored in the voter substore with store key `address`.
+* Notation: For the rest of this proposal let `voterStore(address)` be the object stored in the voter substore with store key `address`, deserialized as a dictionary.
```
janhack, line 179: As also raised on Slack, I am a bit confused with this changed store notation.

proposals/lip-introduce-reward-sharing
```diff
@@ -0,0 +1,549 @@
...
+                module = MODULE_NAME_DPOS,
+                name = EVENT_NAME_REWARDS_ASSIGNED,
+                data={
+                    "voterAddress": voterAddress,
+                    "delegateAddress": vote.delegateAddress,
+                    "tokenID": tokenID,
+                    "amount": reward
+                },
+                topics = [voterAddress]
+            )
```
janhack, line 345: As far as I see we now move the update of the vote sharing coefficient outside this function, so it needs to happen outside of this function after/before this function is called. Is there any specific reason for this specification change?

proposals/lip-0057.md
```diff
@@ -1840,7 +1852,6 @@ if the chain is the mainchain and b is the snapshot block for the migration:
             validator.address,
             validator.generatorKey
         )
-        updateDelegateEligibility(validator.address, 0)
```
janhack: This is removed as we have the implicit assumption that all delegates without BLS key satisfy `isBanned == True`? To me it seems cleaner to not make any assumptions about BLS keys within DPoS and always initialize the eligibility for all delegates.

janhack (changes requested):
Looks good overall. I just have a few small questions and a comment on the notation.

proposals/lip-introduce-reward-sharing
gkoumout: It needs to happen after the function is called actually.

This change would not make sense from my point of view; since this has to be done *every* time after transferring the rewards to the voter, it is less error prone to include it inside the function. However the SDK team preferred to keep separate the store update from this function and just perform the unlock and transfer here. Then, we should update the state store every time after calling this function. The change happened to align the LIP with the actual implementation.

ricott1 (approved):

proposals/lip-0057.md
gkoumout: Replied to this discussion, will take into account on the next push.

proposals/lip-0057.md
gkoumout: The checks included here were the ones that somehow are not necessarily performed before for registered delegates(either on delegate registration or genesis data). 

There is no coupling, exactly because we don't check if before the key was valid or not. Therefore, I don't see any benefit on removing this check to be honest, but could do it if you insist ;-)

proposals/lip-0057.md
gkoumout: The reason to remove this is that we know that `updateDelegateEligibility` will not include them in the eligible delegates substore, so by having this line, you perform useless checks for all delegates registered in mainchain..

Moreover, I don't see how you avoid to make assumptions about BLS keys in DPoS, since we already have this "if" condition, whose only reason of existence is exactly the corner case of mainchain validators with no BLS key.

proposals/lip-introduce-reward-sharing
janhack: Okay, let us keep it as is even though I also agree that the previous specs were cleaner.

In general, I think it is fine if implementation and specifications differ in how functions are structures as long as the protocol is the same in the end. Therefore, we should not feel obliged to change it.

proposals/lip-0057.md
janhack: Okay, sounds good to remove it here as this logic is anyways for the Lisk Mainchain only.

proposals/lip-introduce-reward-sharing
gkoumout: ok, it is easy (and perhaps better) for me to revert it back.

proposals/lip-0057.md
janhack: Looks good now, I resolve this comment.

proposals/lip-introduce-reward-sharing
janhack: To avoid another review, I would keep it as is now.

proposals/lip-0057.md
gkoumout: Ok I  see now: in your example, we should allow unbanning a delegate which is already banned from mainchain, even if they don't have BLS keys. Makes sense.  Removed the check.

janhack: Looks good.

proposals/lip-introduce-reward-sharing
janhack: If you prefer to make the change, feel free to do so. I will review again :-)

janhack (approved):
Nice work!

janhack (approved):

proposals/lip-introduce-reward-sharing.md
```diff
@@ -0,0 +1,551 @@
...
+
+    # calculate reward
+    reward = mul_96(amount, sub_96(delegateSharingCoefficient, voteSharingCoefficient))
+
+    return Q_96_ToInt(reward)
+```
+
+Note that here `i` specifies the index of entries for token with id `tokenID` in the sharing coefficients array in both for the delegate and the voter substore. The fact that both arrays have the same `tokenID` in the same position is guaranteed from the fact that both arrays are sorted in lexicographic order. For the delegate substore array this is done in the [`updateSharedRewards`](#updatesharedrewards) function and for the voter substore in the [assignVoteRewards](#assignvoterewards) function.
+
+#### assignVoteRewards
```
shuse2, line 301: Should we align the pseudocode with the implementation?

gkoumout: We decided finally not to do the update in the LIP (I made the change and then revert it back to the previous state) since the specifications look much cleaner if it is clear that the state change should happen at every call of this function.  I think it is fine if for implementation reasons you slightly modify this, as long as the logic is the same.

shuse2 (approved):

ricott1 (approved):

karmacoma (approved):

PR #188, opened 2022-11-18
Fix timeslot and timestamp issues in LIP 0044 and LIP 0058

AndreasKendziorra:
Currently, slots are counted from the genesis block onward, i.e., the genesis block has slot number 0. Moreover, slots are computed like the following:

```
def getSlotNumber(timestamp):
    elapsedTime = timestamp - genesisTimestamp
    return floor(elapsedTime / BLOCK_TIME)
```
This computation assumes that the genesis timestamp is at the very beginning of its timeslot, i.e., it must be a multiple of 10 seconds. This requirement is, however, not necessarily fulfilled for every genesis block. Hence, the slot computation can return incorrect results. This PR fixes this issue by computing a slot like the following:
```
def getSlotNumber(timestamp):
    return floor(timestamp / BLOCK_TIME)
```
With this, slots are counted form the UNIX epoch onward and are independent of the genesis timestamp. Besides fixing the issue described above, this makes the computation simpler and also eliminates the need to store the genesis timestamp.

In summary, this PR does the following:
- update the slot computation everywhere (slots are only computed in LIP 0044 and LIP 0058)
- remove the storage of the genesis timestamp in LIP 0044 and LIP 0058
- fix some unrelated outdated statements in LIP 0044


proposals/lip-0058.md
```diff
@@ -730,10 +728,7 @@ This function returns the slot number corresponding to the input timestamp.
 
 ```python
 def getSlotNumber(timestamp):
-    if timestamp < genesisTimestamp:
-        raise Exception('Input timestamp cannot be smaller than genesis block timestamp.')
-    elapsedTime = timestamp - genesisTimestamp
-    return floor(elapsedTime / BLOCK_TIME)
+    return floor(timestamp / BLOCK_TIME)
```
ricott1, line 731: Notation: should we use the `//` integer division (as done in other LIPs)?

proposals/lip-0044.md
```diff
@@ -447,13 +419,11 @@ def getValidatorKeys(address: Address) -> ValidatorKeys:
...
-
 #### setValidatorBLSKey
 
-This function sets the BLS key of a validator account. This update is only possible if the current BLS key of the validator is equal to a predefined constant invalid BLS key. It checks that there exists a validator account registered for the input `validatorAddress`, that the input BLS key has not been registered before in the chain, and that the input proof of possession `proofOfPossession` for the BLS key is valid. Finally, it updates the BLS key of the validator account to the input `blsKey`.
+This function sets the BLS key of a validator account. It checks that there exists a validator account registered for the input `validatorAddress`, that the input BLS key has not been registered before in the chain, and that the input proof of possession `proofOfPossession` for the BLS key is valid. Finally, it updates the BLS key of the validator account to the input `blsKey`. 
+
+**Note**: For blockchains build with the SDK v6, this function should be not be called except from the [legacy module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys) used on the Lisk Mainchain. Otherwise, the security of interoperability may brake.
```
ricott1, line 426:
```suggestion
**Note**: For blockchains build with the SDK v6, this function should be not be called except from the [Legacy module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys) used on the Lisk Mainchain. Otherwise, the security of interoperability may brake.
```

ricott1: The last sentence it's a bit vague, maybe we could improve it a bit. I reckon the security breaks because currently we don't support updating BLS keys (maybe the reason for it it's a bit out of scope here), but what exactly would be the problem for Interoperability?

ricott1 (changes requested):
Minor formatting comments.

shuse2 (approved):

proposals/lip-0044.md
```diff
@@ -570,13 +540,10 @@ def getGeneratorsBetweenTimestamps(startTimestamp: uint32, endTimestamp: uint32)
...
         raise Exception('End timestamp cannot be smaller than start timestamp.')
 
-    if startTimestamp < genesisDataStore.timestamp:
-        raise Exception('Start timestamp cannot be smaller than genesis block timestamp.')
-
     result = {}
 
-    startSlotNumber = floor((startTimestamp - genesisDataStore.timestamp) / BLOCK_TIME) + 1
-    endSlotNumber = floor((endTimestamp - genesisDataStore.timestamp) / BLOCK_TIME) - 1
+    startSlotNumber = floor(startTimestamp / BLOCK_TIME) + 1
```
janhack, line 545: We should avoid using `floor` and instead use integer division (`//`).

proposals/lip-0044.md
janhack: I did not spend much time to think about it, but I would not see a security issue and think the protocol in principle should be equipped to do BLS key updates.

proposals/lip-0058.md
```diff
@@ -704,8 +703,7 @@ This function returns the address of the generator active at the timestamp `time
 
 ```python
 def getGeneratorAtTimestamp(timestamp, currentHeight):
-    elapsedTime = timestamp - genesisTimestamp
-    slotIndex = floor(elapsedTime / BLOCK_TIME) % len(generatorList)
+    slotIndex = floor(timestamp / BLOCK_TIME) % len(generatorList)
```
janhack, line 706: We should avoid using `floor` and instead use integer division (`//`).

proposals/lip-0058.md
```diff
@@ -730,10 +728,7 @@ This function returns the slot number corresponding to the input timestamp.
...
 ```python
 def getSlotNumber(timestamp):
-    if timestamp < genesisTimestamp:
-        raise Exception('Input timestamp cannot be smaller than genesis block timestamp.')
-    elapsedTime = timestamp - genesisTimestamp
-    return floor(elapsedTime / BLOCK_TIME)
+    return floor(timestamp / BLOCK_TIME)
 ```
 
 ### Genesis Block Processing
```
janhack, line 734: At the moment `genesisTimestamp` is still set during the genesis block processing. I would remove it from there.

janhack (changes requested):
I just have a few small points.

proposals/lip-0058.md
AndreasKendziorra: I changed it everywhere in LIP 0058 and LIP 0044

proposals/lip-0044.md
AndreasKendziorra: To be honest, I also cannot recall the exact reason why we did not allow it from the beginning. But I thought it is (also) due to security reasons. Maybe there is indeed no issue with this. But I'm not sure we should leave it like this (without the warning) without thinking it completely through. But do you prefer to remove the statement?

proposals/lip-0044.md
AndreasKendziorra: changed everywhere in LIP 0044 and LIP 0058.

proposals/lip-0058.md
AndreasKendziorra: changed everywhere in LIP 0044 and LIP 0058.

proposals/lip-0058.md
AndreasKendziorra: True, shouldn't be there. I removed it.

AndreasKendziorra:
> Minor formatting comments.
> 
> In my understanding, removing the genesis timestamp from the computation implies that now the genesis slot is not necessarily 0. I would still add a clarifying comment on this, since it is not necessarily intuitive. Maybe we could add it to the description of the `getSlotNumber` function, explaining that the slots are counted form the UNIX epoch onward and are independent of the genesis timestamp, hence the genesis slot may be anything.

@ricott1 I added a comment

proposals/lip-0044.md
ricott1: I would not remove the statement without discussing it first.

ricott1 (approved):

janhack (approved):
Looks all good, nice work Andreas!

karmacoma (approved):

PR #191, opened 2022-11-21
Ban every delegate in snapshot block

AndreasKendziorra:
This is a complementing PR for [PR 185](https://github.com/LiskHQ/lips/pull/185). In PR 185, the check that an active delegate has a BLS key different to `INVALID_BLS_KEY` is removed from the DPoS module. This check was introduced for the Mainchain migration from Lisk Core v3 to v4 to avoid that validators without valid BLS key become active validators.
After removing this check, a different mechanism to prevent that validators without valid BLS key become active validators is needed. This is done in this PR by banning every validator in the snapshot block. Validators can get unbanned by submitting a _register key_ transaction. The logic for unbanning was already introduced in PR 185.


proposals/lip-0063.md
```diff
@@ -565,25 +565,25 @@ Once the snapshot block is final, a patched version of Lisk Core v4 will be rele
...
 
-To avoid the need to send an _updated generator key_ right after the hard fork, registered validators that did not forge a block and did not send a transaction in the Lisk Core v3 history could simply send a transaction any time before the hard fork. Then the sender public key of this transaction will be set as the generator key as described above. Nevertheless, every validator is encouraged to perform an _update generator key_ with a generator key different from the public key used for transaction signatures eventually. This is due to [security benefits](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0044.md#generator-key) added by the concept of generator keys.
+The _[register keys](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys)_ transaction does not only allow to set a BLS key but also a new generator key. Hence, every validator has the opportunity to set a new generator key. Notice, however, that this transaction can only be submitted once per valiator, and the [current version of the DPoS module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md) does not provide any update mechanism for the generator key.
```
gkoumout, line 576: "valiator" -> validator

gkoumout (approved):
Looks good to me! Just mentioned one typo. Otherwise, can be approved from my side

proposals/lip-0063.md
```diff
@@ -565,25 +565,25 @@ Once the snapshot block is final, a patched version of Lisk Core v4 will be rele
...
...egistration-and-proof-of-possession) in order to become a validator. This results in an exception for Lisk Core v4 as the existing validators do not have a valid BLS key right after the hard fork. Since the DPoS module expects that each validator account has a BLS key and a proof of possession set in the snapshot block (see [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-state-initialization)), this will be done by assigning a fixed public key for which signature validation will always fail along with a dummy value for the proof of possession. This proof of possession will never be evaluated as the DPoS and the validator module handle an exception for this particular case (see [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-state-finalization) and [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0044.md#registervalidatorwithoutblskey)). Every validator must then register a valid BLS key via a _[register keys](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys)_ transaction in order to become an active validator, ideally within the [bootstrap period](#bootstrap-period) described below.
```
sameersubudhi, line 570: Should `Validator` here begin with a capital letter, since we're referring to the module?
```
This proof of possession will never be evaluated as the DPoS and the validator module...
```

```suggestion
In blockchains created with Lisk SDK v6, one needs to [register](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0044.md#registervalidatorkeys) a [valid BLS key](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0038.md#public-key-registration-and-proof-of-possession) in order to become a validator. This results in an exception for Lisk Core v4 as the existing validators do not have a valid BLS key right after the hard fork. Since the DPoS module expects that each validator account has a BLS key and a proof of possession set in the snapshot block (see [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-state-initialization)), this will be done by assigning a fixed public key for which signature validation will always fail along with a dummy value for the proof of possession. This proof of possession will never be evaluated as the DPoS and the Validator module handle an exception for this particular case (see [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-state-finalization) and [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0044.md#registervalidatorwithoutblskey)). Every validator must then register a valid BLS key via a _[register keys](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys)_ transaction in order to become an active validator, ideally within the [bootstrap period](#bootstrap-period) described below.
```

proposals/lip-0063.md
```diff
@@ -565,25 +565,25 @@ Once the snapshot block is final, a patched version of Lisk Core v4 will be rele
...
+There will be a bootstrap period in which a fixed set of delegates will constitute the set of active delegates. This period will last for one week assuming no missed blocks. This is done by setting the [`initRounds`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-assets-schema) property in the snapshot block to the corresponding value, which is given by `DPOS_INIT_ROUNDS`. The reason to have this period is to achieve that the DPoS module selects only delegates with a valid BLS key. And this is realized by banning every delegate in the snapshot block. The unbanning is triggered by a _[register keys](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-keys)_ transaction. This way, delegates are forced to submit a _register keys_ transaction, and therefore to register a valid BLS key and a (new) generator key, in order to become an active delegate.
```
sameersubudhi, line 584: Should we not be switching terminologies to `validator(s)` instead of `delegate(s)` here, at least for the descriptions for concerned situations occurring post migration?

sameersubudhi (changes requested):
Additionally, should we also not update all references to `DPoS` to `PoS` instead?
Ex: `addDPoSModuleEntry` to `addPoSModuleEntry`.

proposals/lip-0063.md
janhack: > The reason to have this period is to achieve that the DPoS module selects only delegates with a valid BLS key. And this is realized by banning every delegate in the snapshot block.

The new version of the DPoS module actually does not care about BLS keys at all, as it never checks whether a delegate has a valid BLS key in the selection (to have clean and performant code). Instead, the DPoS module only discards banned and punished delegates. This is also the reason that we ban delegates without BLS keys.

Moreover, maybe I would again add "outside the bootstrap period" to make clear that the DPoS selection algorithm only kicks in then.

janhack (changes requested):
I have one small comment on a formulation in the rationale, otherwise all good.

proposals/lip-0063.md
sergeyshemyakov, line 584: Maybe this change is still upcoming, but where do we call the function `unbanDelegate`? Or where are we planning to call it?

sergeyshemyakov (approved):
Just one question from my side, doesn't block this LIP.

proposals/lip-0063.md
AndreasKendziorra: fixed

proposals/lip-0063.md
AndreasKendziorra: @sameersubudhi so far, the terminology switch from DPoS to PoS was not made public. Therefore, I would keep the current way to keep the term "delegate" for DPoS validators for now. And then we switch the whole terminology in one go. But I must admit that we, including me, have been quite sloppy with terminology. We mix "validator" and "delegate" quite a bit in this LIP but also the DPoS module LIP.

> The new version of the DPoS module actually does not care about BLS keys at all,

Yes. I thought my statements are not contradicting this fact. But I changed this paragraph and also the next one (this one really needed a fix) to make it hopefully more clear.

proposals/lip-0063.md
```diff
@@ -565,25 +565,25 @@ Once the snapshot block is final, a patched version of Lisk Core v4 will be rele
...
 
-The bootstrap period allows each delegate to submit a _[register BLS key](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0050.md#register-bls-key)_ transaction in a sufficiently large time window. The expectation is that most delegates have a valid BLS key after the bootstrap period. In the unexpected case that there are less than 101 non-banned delegates with a BLS key, the DPoS module would simply select a smaller set of delegates that get the block slots of a round assigned round robin.
+The bootstrap period allows each delegate to submit a _register keys_ transaction in a sufficiently large time window. The expectation is that most delegates have a valid BLS key after the bootstrap period. In the unexpected case that there are less than 101 non-banned delegates with a BLS key, the DPoS module would simply select a smaller set of delegates that get the block slots of a round assigned round robin.
```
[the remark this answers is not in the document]
AndreasKendziorra: same as above

sameersubudhi (approved):
LGTM 👍

proposals/lip-0063.md
janhack: All good, I resolve this comment.

janhack (approved):
Looks all good.

karmacoma (approved):

PR #200, opened 2022-12-09
Add module dependency and execution order explanation

shuse2:
### Description

- Update LIP 69 to explain execution order and dependency
- Clarify what can be accessed in the hook


proposals/lip-0069.md
```diff
@@ -61,11 +61,11 @@ _Endpoint_ is the interface for a module to an external system through an RPC en
 
 #### Application hooks
 
-Application hooks are module methods which are called during the block execution. These hooks cannot mutate the block, but they can introduce state changes.
+Application hooks are module methods which are called during the block execution. These hooks can access the currently executing block header and the block assets, and make state mutations.
```
janhack, line 64:
```suggestion
Application hooks are module methods which are called during the block execution. These hooks can access the block header being currently executed and its block assets, and make state mutations.
```

proposals/lip-0069.md
```diff
@@ -61,11 +61,11 @@ _Endpoint_ is the interface for a module to an external system through an RPC en
...
 
 #### Block Generation hooks
 
-Block generation hooks are module methods which are called only during the block generation. These hooks cannot introduce state changes, but they can add information to the block assets.
+Block generation hooks are module methods which are called only during the block generation. These hooks cannot introduce state changes, but they have access to the currently generating block and they can add information to the block assets.
```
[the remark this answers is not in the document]
janhack: I think passive voice makes more sense.

proposals/lip-0069.md
janhack: Should we also mention that they can access the state store?

proposals/lip-0069.md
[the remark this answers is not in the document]
janhack: Should we also mention that they can access the state store?

shuse2:
@janhack Accepted the suggestions and add mention about the access to the current state.

As we discussed, I think own chain ID and temporal memory are quite implementations detail and minor things in the context of general module architecture, so I think it would be better to keep it out of this LIP

proposals/lip-0069.md
```diff
@@ -61,11 +61,11 @@ _Endpoint_ is the interface for a module to an external system through an RPC en
 
 #### Application hooks
 
-Application hooks are module methods which are called during the block execution. These hooks cannot mutate the block, but they can introduce state changes.
+Application hooks are module methods which are called during the block execution. These hooks can access the current state, the block header being currently executed and its block assets, and make state mutations.
```
ricott1, line 64: If we say that these hooks access the block assets, shouldn't we also say that they access the block transactions? I think what we want to say is that they can access the current state and the block header to process the state transitions induced by the block assets and the block transactions.

proposals/lip-0069.md
```diff
@@ -81,6 +81,10 @@ _Events_ are on-chain data emitted during the processing of a block which add ex
 
 Lisk SDK modules can define block generation hooks, state machine hooks and commands to add logic to the state machine. Each hook of the module will be called as described in the life cycle below (also described in [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md)).
 
+
+For the hooks which call all registered modules, the hook of a module will be called in the order of the registration.
+In some modules, it requires methods from other modules, and for that, the dependency needs to be resolved before registering the module to the application.
```
ricott1, line 86: What do we mean concretely by "the dependency needs to be resolved"?

proposals/lip-0069.md
janhack: So basically it is the whole block and the state. Depending on the scope of the hook, you can only access a part of the block, e.g., one transaction and the block header, or one asset and the block header.

janhack: Maybe we could write something like the following:

> Application hooks are module methods which are called during the block execution. In general, these hooks can access the current state and the block being currently executed, and make state mutations. In most cases we restrict the access to the state and block to have a clean and modular framework for state mutations.  For instance, in the hook for executing a command only the containing transaction, certain properties in the block header and the state of the respective module is accessible.

janhack (changes requested):
I did not have any new comments and my previous comments were resolved.

The only two open points, also from my side, are the comments by Alessandro.

proposals/lip-0069.md
shuse2: What i wanted to mention was [these logics](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/application.ts#L183-L200), where dependencies are manually resolved

proposals/lip-0069.md
janhack: Can be resolved from my side.

janhack (approved):
Looks all good.

proposals/lip-0069.md
ricott1: Then I would maybe explicitly say that the dependency must be 'stated' or 'added'?

proposals/lip-0069.md
```diff
@@ -61,11 +61,11 @@ _Endpoint_ is the interface for a module to an external system through an RPC en
 
 #### Application hooks
 
-Application hooks are module methods which are called during the block execution. These hooks cannot mutate the block, but they can introduce state changes.
+Application hooks are module methods which are called during the block execution. In general, these hooks can access the current state and the block being currently executed, and make state mutations. In most cases we restrict the access to the state and block to have a clean and modular framework for state mutations. For instance, in the hook for executing a command only the containing transaction, certain properties in the block header and the state of the respective module is accessible.
```
ricott1, line 64:
```suggestion
Application hooks are module methods which are called during the block execution. In general, these hooks can access the current state and the block being currently executed, and make state mutations. In most cases we restrict the access to the state and block to have a clean and modular framework for state mutations. For instance, in the hook for executing a command only the containing transaction, certain properties in the block header and the state of the respective module are accessible.
```

ricott1 (changes requested):
a couple minor comments

ricott1 (approved):

PR #204, opened 2023-01-12
Small fixes to PoA module LIP

sergeyshemyakov:
Addressed the following:

- In the section "Migration from PoA to DPoS" we should add that when having PoA the reward module of LIP0042 is used but when switching to DPoS we recommend using the new Dynamic Block rewards module (although DPoS might also work with the old one, but no reason for that).
- The type Transaction is used in several places, but it is not defined. In particular, it is not clear that for an instance of Transaction the params property is of type object instead of bytes.
- Remove extra command fee from the "registerAuthority" command and call fee.payFee(REGISTRATION_FEE) instead.

Note that the fee was already paid correctly, only extra command fee was present in the schema and verification.


AndreasKendziorra (changes requested):

mjerkov (approved):
Looks good!

AndreasKendziorra (approved):

PR #206, opened 2023-01-13
Small fixes in PoS and reward sharing

gkoumout:
This PR fixes some small issues on the PoS and reward sharing LIPs. In particular:

 - Sets `MAX_COMMISSION_INCREASE_RATE` and `COMMISSION_INCREASE_PERIOD` to configurable constants, instead of PoS constants that were before.
 - Fixes the logic of the `isUnlockable` function. It was not returning the correct outcome for the following case (assuming `PUNISHMENT_WINDOW_STAKING = 1 month` and `PUNISHMENT_WINDOW_SELF_STAKING = 3 months`):  In case a staker unstakes from a validator 2 months after the last misbehavior report, and then attempts for unlock immediately, previous specs would allow to unlock (`isUnlockable` would return `True`) without checking if the locking time has expired. 
 - Small other typos/inconsistencies.


janhack (approved):
The PR looks good, but could you explain what the issue in `isUnlockable` was?

I understand that the logic now and previously, is different for odd parameter choices, such as `LOCKING_PERIOD_SELF_STAKING > PUNISHMENT_WINDOW_SELF_STAKING` or `LOCKING_PERIOD_STAKING > PUNISHMENT_WINDOW_STAKING`.

proposals/lip-0057.md
```diff
@@ -879,21 +881,22 @@ The definition and rationale for the `isCertificateGenerated` function is part o
...
 
-    else: # Validator is punished.
+    if len(validatorStore(validatorAddress).reportMisbehaviorHeights) > 0: # Validator might be punished.
         let lastReportMisbehaviorHeight be the last element of validatorStore(validatorAddress).reportMisbehaviorHeights
         # lastReportMisbehaviorHeight is also the largest element of the reportMisbehaviorHeights array.
-        punishmentWindow = PUNISHMENT_WINDOW_SELF_STAKING if validatorAddress == senderAddress else PUNISHMENT_WINDOW_STAKING
-
-        if height – lastReportMisbehaviorHeight < punishmentWindow and lastReportMisbehaviorHeight < unlockObject.unstakeHeight + lockingPeriod:
+ 
+        if height < lastReportMisbehaviorHeight + punishmentWindow and lastReportMisbehaviorHeight < unlockObject.unstakeHeight + lockingPeriod:
```
ricott1, line 899: Basically this is not `<=` because locking and punishment periods have to be intended as 'after how many blocks is the period over', right?

ricott1: also, not elated to this PR, but I think the description of `PUNISHMENT_WINDOW_STAKING` and `PUNISHMENT_WINDOW_SELF_STAKING` in the table could be improved. Instead of 'The punishment time' I would maybe say 'punishment period' or even 'number of blocks' to make it clear that this is a number of blocks, not a time.

proposals/lip-0057.md
```diff
@@ -879,21 +881,22 @@ The definition and rationale for the `isCertificateGenerated` function is part o
 
 ```python
```
janhack, line 882: Do we now still need the following sentence above:

> It's logic is the concatenation of the functions hasWaited and isPunished that are rationalized in [LIP 0023](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0023.md#explicit-unlock-mechanism) and [LIP 0024](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0024.md#rationale) respectively.

This seems not relevant at all for the specifications here and is rather confusing I find.

ricott1: also there is a typo :)

ricott1 (approved):

proposals/lip-0057.md
janhack: We already refer to those LIPs in the rationale, so I think it is fine to remove it here. Can be resolved from my side.

janhack (approved):
Looks all good.

PR #207, opened 2023-01-15
LIP0043: fix registration schemas and store key encoding

ricott1:
- Make sidechain registration transaction parameters order consistent with mainchain registration command.
- Fix missing required property in mainchain registration params schema.
- Add `minItem` and `maxItem` keywords to json schemas.
- Fix constant name `MESSAGE_TAG_CHAIN_REG_MESSAGE` to make it consistent with LIP37
- encode string before using it as store key

Solve parts of https://github.com/LiskHQ/lips/issues/216


proposals/lip-0043.md
```diff
@@ -351,22 +351,23 @@ Transactions executing this command have:
...
             "minLength": MIN_CHAIN_NAME_LENGTH,
             "maxLength": MAX_CHAIN_NAME_LENGTH,
+            "fieldNumber": 1
+        },
+        "ownChainID": {
+            "dataType": "bytes",
+            "length": CHAIN_ID_LENGTH,
             "fieldNumber": 2
         },
         "mainchainValidators": {
```
mitsuaki-u, line 373: Can we add here:  `minItems: 1,
			maxItems: NUMBER_ACTIVE_VALIDATORS_MAINCHAIN,`

And remove: `   # mainchainValidators must have at least 1 element and at most NUMBER_ACTIVE_VALIDATORS_MAINCHAIN elements.
    if len(trs.params.mainchainValidators) < 1 or len(trs.params.mainchainValidators) > NUMBER_ACTIVE_VALIDATORS_MAINCHAIN:
raise Exception(f"Invalid mainchain validators array. It must have at least 1 element and at most {NUMBER_ACTIVE_VALIDATORS_MAINCHAIN} elements.")` from the verify hook?

proposals/lip-0043.md
```diff
@@ -480,6 +477,8 @@ def execute(trs: Transaction) -> None:
             },
             "mainchainValidators": {
                 "type": "array",
+                "minItems": 1,
+                "maxItems": NUMBER_ACTIVE_VALIDATORS_MAINCHAIN,
```
AndreasKendziorra, line 481: These requirements are redundant because the checks are already performed in the `verify` hook. And if the checks are repeated in the execute (if we should decided that encode should always perform full schema validation) than we do some unnecessary execution steps.
The same holds for the lines 474-475 and 489 (I also missed this before).

proposals/lip-0045.md
```diff
@@ -206,7 +206,7 @@ In this section, we specify the substores that are part of the Interoperability
 | `CCM_STATUS_CODE_FAILED_CCM` | uint32 | 4 | Value of `status` of returned CCM due to error: failed ccm execution. |
 | `CCM_STATUS_CODE_RECOVERED` | uint32 | 5 | Value of `status` of CCM that have been recovered with a message recovery command. |
 | **Message Tags** | | | |
-| `MESSAGE_TAG_CHAIN_REG` | bytes | `“LSK_CHAIN_REGISTRATION_”` ASCII-encoded | Message tag for chain registration message. |
+| `MESSAGE_TAG_CHAIN_REG_MESSAGE` | bytes | `“LSK_CHAIN_REGISTRATION_”` ASCII-encoded | Message tag for chain registration message. |
```
gkoumout, line 209: Minor comment: I understand that here we want to be consistent with LIP 37, but is it a good naming strategy to have twice the same word in a constant name? Perhaps keeping the old one and updating LIP37 would make more sense, but no strong opinion.

gkoumout (approved):

AndreasKendziorra (changes requested):

proposals/lip-0043.md
ricott1: done

proposals/lip-0045.md
ricott1: The point here is that the first message refers to the fact that this is an encoded value ( we use the message prefix for all of them), the second one to the fact that this is called 'chain registration message' in LIP 43

proposals/lip-0043.md
ricott1: discussed in person, we keep it in the schema

AndreasKendziorra: I got convinced that there good reasons to keep it.

AndreasKendziorra (approved):

proposals/lip-0045.md
gkoumout: Yep, I understand this.

ricott1: Can I resolve this then? @gkoumout

PR #208, opened 2023-01-18
Fix issues in LIP 0063

sergeyshemyakov:
The issues identified by the Service team and addressed in the PR:

- Total LSK supply did not account for legacy accounts.
- Many arrays used to generate the snapshot block were not explicitly sorted by the address.
- Version 3 account DPoS-related properties were erroneously renamed during DPoS -> PoS terminology change.

I have also addressed https://github.com/LiskHQ/lips/issues/221 in this PR by adding Interoperability genesis asset to the snapshot block.

### DIFF
--- a/proposals/lip-0063.md
+++ b/proposals/lip-0063.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/define-mainnet-configuration-and-mig
 Status: Draft
 Type: Standards Track
 Created: 2022-04-06
-Updated: 2023-01-18
+Updated: 2023-01-20
 Requires: 0060
 ```
 
--- a/proposals/lip-0063.md
+++ b/proposals/lip-0063.md
@@ -68,9 +68,9 @@ This value will be used for every validator account in the snapshot block.
    </td>
    <td>bytes
    </td>
-   <td><code>0x 00 00 00 00</code>
+   <td><code>0x 00 00 00 00</code> (in Lisk Mainnet)
    </td>
-   <td>The chain ID of the mainchain of Lisk Mainnet, see <a href="https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#chain-identifiers">LIP 0037</a>.
+   <td>The chain ID of the mainchain of Lisk, see <a href="https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#chain-identifiers">LIP 0037</a>.
    </td>
   </tr>
   <tr>
@@ -589,7 +589,7 @@ def createGenesisDataObj():
 
 ###### Assets Entry for Interoperability Module
 
-Let `genesisInteroperabilityStoreSchema` be as defined in [LIP 0045](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#genesis-state-initialization). The `assets` entry for the Interoperability module is added by the logic defined in the function `addInteropModuleEntry` in the following pseudo code. The function `getMainchainID` returns the chain ID of the mainchain in the current network, it is defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#mainchain-chain-id).
+Let `genesisInteroperabilityStoreSchema` be as defined in [LIP 0045](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#genesis-state-initialization). The `assets` entry for the Interoperability module is added by the logic defined in the function `addInteropModuleEntry` in the following pseudo code.
 
 ```python
 def addInteropModuleEntry():
@@ -600,7 +600,7 @@ def addInteropModuleEntry():
     InteropObj.chainValidatorsSubstore = []
     ownChainAccount = {
         "name": CHAIN_NAME_MAINCHAIN,
-        "chainId": getMainchainID(),
+        "chainId": CHAIN_ID_MAINCHAIN,
         "nonce": 0
     }
     InteropObj.ownChainDataSubstore = [{
@@ -608,7 +608,7 @@ def addInteropModuleEntry():
         "storeValue": ownChainAccount}]
     InteropObj.terminatedStateSubstore = []
     InteropObj.terminatedOutboxSubstore = []
-    registeredNamesStore = {"chainID": getMainchainID()}
+    registeredNamesStore = {"chainID": CHAIN_ID_MAINCHAIN}
     InteropObj.registeredNamesSubstore = [{
         "storeKey": bytes(CHAIN_NAME_MAINCHAIN, 'utf-8'),
         "storeValue": registeredNamesStore}]
--- a/proposals/lip-0063.md
+++ b/proposals/lip-0063.md
@@ -423,23 +423,26 @@ def createUserSubstoreArray():
             userObj.lockedBalances = getLockedBalances(account)
             userSubstore.append(userObj)
     # Append the legacy reserve account separately.
-    userSubstore.append(createLegacyReserveAccount())
+    addLegacyReserveAccount(userSubstore)
     sort userSubstore in the lexicographical order of (userObj.address + userObj.tokenID)
     return userSubstore
 
-def createLegacyReserveAccount():
-    legacyReserveAccount = account in accounts with account.address == ADDRESS_LEGACY_RESERVE
-    isEmpty = legacyReserveAmount is empty
-    legacyReserve = {}
-    legacyReserve.address = ADDRESS_LEGACY_RESERVE
-    legacyReserve.tokenID = TOKEN_ID_LSK
-    legacyReserve.availableBalance = isEmpty ? 0 : legacyReserveAccount.token.balance
+def addLegacyReserveAccount(userSubstore):
     legacyReserveAmount = 0
     for every account in legacyAccounts:
         legacyReserveAmount += account.token.balance
-    lockedBalances = isEmpty ? [] : getLockedBalances(legacyReserveAccount)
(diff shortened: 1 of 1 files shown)


proposals/lip-0063.md
```diff
@@ -469,31 +474,32 @@ def createValidatorsArray():
...
         validator.blsKey = INVALID_BLS_KEY
         validator.proofOfPossession = DUMMY_PROOF_OF_POSSESSION
         if validatorKeys[account.address]:
             validator.generatorKey = validatorKeys[account.address]
         else:
             validator.generatorKey = INVALID_ED25519_KEY
-        validator.lastGeneratedHeight = account.pos.validator.lastForgedHeight
+        validator.lastGeneratedHeight = account.dpos.delegate.lastForgedHeight
         validator.isBanned = True
-        validator.reportMisbehaviorHeights = account.pos.validator.reportMisbehaviorHeights
```
janhack, line 485: I think it should be `account.dpos.delegate.pomHeights` as this is the property name in v3, see https://github.com/LiskHQ/lips/blob/main/proposals/lip-0030.md.

janhack:
LIP 63 uses permanent links in several places, which link you to an outdated version of the LIP, e.g., 
```
[LIP 0051](https://github.com/LiskHQ/lips/blob/aebeecc40447e4229f14fc641fb11527270dfbb5/proposals/lip-0051.md#genesis-assets-schema)
```
I would suggest to update those to usual links. Possibly, we could also have all links at the bottom of the document. Note that this does not work for Html tables.

proposals/lip-0063.md
```diff
@@ -412,9 +413,9 @@ def createLegacyReserveAccount():
```
janhack, line 413: Regarding the line above, I think the following property is not satisfied for the legacy account:

> The lockedBalances array must be in lexicographic order of module name.

proposals/lip-0063.md
```diff
@@ -428,7 +429,10 @@ def createSupplySubstoreArray():
         lockedBalances = getLockedBalances(account)
         if lockedBalances is not empty:
             totalLSKSupply += lockedBalances[0].amount
-    LSKSupply = {"tokenID": TOKEN_ID_LSK, "totalSupply": totalLSKSupply}
+    legacyReserveAccount = account with account.address == ADDRESS_LEGACY_RESERVE
```
janhack, line 432: This is a bit confusing as `accounts` and `legacyAccounts` contain the state data from Lisk Core v3 and only `tokenObj.userSubstore` contains the accounts and locked tokens in the new format. However, `tokenObj.userSubstore` is not an input parameter to the function at the moment.

proposals/lip-0063.md
```diff
@@ -469,31 +474,32 @@ def createValidatorsArray():
...
         validator.address = account.address
-        validator.name = account.pos.validator.username
+        validator.name = account.dpos.delegate.username
         validator.blsKey = INVALID_BLS_KEY
         validator.proofOfPossession = DUMMY_PROOF_OF_POSSESSION
         if validatorKeys[account.address]:
             validator.generatorKey = validatorKeys[account.address]
         else:
             validator.generatorKey = INVALID_ED25519_KEY
-        validator.lastGeneratedHeight = account.pos.validator.lastForgedHeight
```
janhack, line 483: Note that the property name in LIP 57 is `lastGeneratedHeight`, see https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#genesis-assets-schema.

janhack (changes requested):
There are a few small issues, otherwise it looks good.

proposals/lip-0063.md
sergeyshemyakov: Right, fixed.

proposals/lip-0063.md
sergeyshemyakov: Good point, explicitly sorted the array after adding the Legacy locked tokens.

proposals/lip-0063.md
sergeyshemyakov: I think this is a correct line:
`validator.lastGeneratedHeight = account.dpos.delegate.lastForgedHeight`
I think it is in the LIP. Let me know if you think something should be changed.

proposals/lip-0063.md
sergeyshemyakov: Indeed, the logic or legacy reserve account was wrong because of mixing up v3 accounts with v4 accounts. Now it is hopefully fixed.

AndreasKendziorra (changes requested):

proposals/lip-0063.md
```diff
@@ -423,23 +423,26 @@ def createUserSubstoreArray():
...
     return userSubstore
 
-def createLegacyReserveAccount():
-    legacyReserveAccount = account in accounts with account.address == ADDRESS_LEGACY_RESERVE
-    isEmpty = legacyReserveAmount is empty
-    legacyReserve = {}
-    legacyReserve.address = ADDRESS_LEGACY_RESERVE
-    legacyReserve.tokenID = TOKEN_ID_LSK
-    legacyReserve.availableBalance = isEmpty ? 0 : legacyReserveAccount.token.balance
+def addLegacyReserveAccount(userSubstore):
```
janhack, line 430: Why do we do a big refactoring here where it would have been sufficient to add one line
```
 sort legacyReserve.lockedBalances in the lexicographical order of lockedBalance.module
```

proposals/lip-0063.md
```diff
@@ -537,6 +577,35 @@ def createGenesisDataObj():
...
+```python
+def addInteropModuleEntry():
+    InteropObj = {}
+    InteropObj.outboxRootSubstore = []
+    InteropObj.chainDataSubstore = []
+    InteropObj.channelDataSubstore = []
+    InteropObj.chainValidatorsSubstore = []
+    ownChainAccount = {
+        "name": CHAIN_NAME_MAINCHAIN,
+        "chainId": getMainchainID(),
```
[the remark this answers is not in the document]
janhack: I would recommend to merge the main branch into your branch. The constants table there already defines `CHAIN_ID_MAINCHAIN`. Then you can use this value here.

proposals/lip-0063.md
```diff
@@ -537,6 +577,35 @@ def createGenesisDataObj():
...
+        "nonce": 0
+    }
+    InteropObj.ownChainDataSubstore = [{
+        "storeKey": EMPTY_BYTES,
+        "storeValue": ownChainAccount}]
```
janhack, line 598: I think some serialization step is missing here, i.e.,
`encode(ownChainAccountSchema, ownChainAccount)`

proposals/lip-0063.md
janhack: Lookds good now.

proposals/lip-0063.md
janhack: All good, I guess I could confused which account is the v3 and which one is the v4.

janhack (changes requested):
Just some small comments left.

proposals/lip-0063.md
sergeyshemyakov: I got confused with v3 vs v4 accounts, but these changes were wrong in the end. I reverted to the previous version and added the locked module sorting. 
The `legacyReserveAccount` represents a v3 account, `legacyReserve` represents a v4 account.

proposals/lip-0063.md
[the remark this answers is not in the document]
sergeyshemyakov: I would prefer to leave the function explicitly, since we also want to migrate in Testnet. There is a reference to LIP 0037 after the constants table, but I think it is nice to be more precise and mention it directly in the pseudocode.
Also I have confirmed with the Platform team that this function is available.

proposals/lip-0063.md
```diff
@@ -537,6 +577,35 @@ def createGenesisDataObj():
...
+    InteropObj.terminatedStateSubstore = []
+    InteropObj.terminatedOutboxSubstore = []
+    registeredNamesStore = {"chainID": getMainchainID()}
+    InteropObj.registeredNamesSubstore = [{
+        "storeKey": bytes(CHAIN_NAME_MAINCHAIN, 'utf-8'),
+        "storeValue": registeredNamesStore}]
```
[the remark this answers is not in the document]
sergeyshemyakov: My opinion is that registered names substore should be a part of genesis block for the same reason as own chain account: it should be set to default value only for the mainchain. Waiting for input from @AndreasKendziorra and @ricott1.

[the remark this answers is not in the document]
AndreasKendziorra: Sorry, I missed that this store is only present on the mainchain. So, it's correct to do it here.

AndreasKendziorra (approved):

proposals/lip-0063.md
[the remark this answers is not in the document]
janhack: I would still strongly favor using `CHAIN_ID_MAINCHAIN`. We define the computation of the snapshot block in the context of Lisk Core 3.0 where `OWN_CHAIN_ID` should not be defined and therefore the function `getMainchainID()` does not really make sense. In some place of the Migrator, we will need to hardcode `CHAIN_ID_MAINCHAIN` as this value cannot be computed. And, as already mentioned below the constants table, this hard-coded value will need to be different in depending on where we do the migration. Therefore, I would find it better to stick with the notation `CHAIN_ID_MAINCHAIN` in the whole LIP (even if internally in the migrator tool they introduce a function `getMainchainID()` for it).

janhack (approved):
Looks all good.

proposals/lip-0063.md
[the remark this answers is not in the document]
janhack: Looks all good.

Issue #215, opened 2023-01-18
Update the `Register Keys` command in LIP 0050


mjerkov:
The [statement](https://github.com/LiskHQ/lips/blob/e57af53de675d64c6c3bcc9201058739460c4e4c/proposals/lip-0050.md?plain=1#L269) "This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module" is outdated because the Validators module does allow it.

PR #222, opened 2023-01-19
Fix minor issues in Legacy module

sergeyshemyakov:
Resolves https://github.com/LiskHQ/lips/issues/215.

### DIFF
--- a/proposals/lip-0050.md
+++ b/proposals/lip-0050.md
@@ -64,6 +64,7 @@ We define the following constants:
 | Name | Type | Validation | Description |
 |------|------|------------|-------------|
 | `PublicKeyEd25519` | bytes | Must be of length 32. | Used for Ed25519 public keys. |
+| `Transaction` | object | Must follow the `transactionSchema` schema defined in [LIP 0068](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0068.md) with the only difference that `params` property is not serialized and contains the values of parameters of `paramsSchema` for the corresponding transaction. | An object representing a non-serialized transaction. |
 
 #### Functions from Other Modules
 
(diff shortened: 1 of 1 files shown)


AndreasKendziorra (changes requested):

ricott1 (changes requested):
I agree with @Andreas Kendziorra comment and I would not specify the transaction type here (if I correctly understood it will be introduced by another PR)

proposals/lip-0050.md
```diff
@@ -266,7 +267,7 @@ def execute(trs: Transaction) -> None:
 
 #### Register Keys
 
-This command allows migrated legacy validators register the required keys. In particular, this command is used by all migrated validators to register a BLS key. This command cannot be used to modify an existing BLS key, as this is forbidden by the Validators module. Transactions executing this command have:
+This command allows migrated legacy validators register the required keys. In particular, this command is used by all migrated validators to register a BLS key. It could also be used to modify an existing BLS key. Transactions executing this command have:
```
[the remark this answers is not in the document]
sergeyshemyakov: Ah, missed it. Now added a separate check for existence of validator account.

proposals/lip-0050.md
```diff
@@ -64,6 +64,7 @@ We define the following constants:
 | Name | Type | Validation | Description |
 |------|------|------------|-------------|
 | `PublicKeyEd25519` | bytes | Must be of length 32. | Used for Ed25519 public keys. |
+| `Transaction` | object | Must follow the `transactionSchema` schema defined in [LIP 0068](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0068.md) with the only difference that `params` property is not serialized and contains the values of parameters of `paramsSchema` for the corresponding transaction. | An object representing a non-serialized transaction. |
```
ricott1, line 67: As mentioned, I would not define this type here, also because I think we want to move to a definition where `params` is still of type bytes

AndreasKendziorra: Not a strong opinion on this on. Maybe better to remove it for now. We will not forget about this issue since it is already covered somewhere else.

AndreasKendziorra (approved):

ricott1 (approved):

janhack (approved):

Issue #209, opened 2023-01-18
Allow transfering fees to account instead of burning them


janhack:
# Description

Introduce the possibility that in the Fee module, fees are transferred to a specified account instead of burning them. This could be achieved by:

- Adding configurable constant `BURN_FEES` (boolean) and `ADDRESS_FEE_POOL` (20 byte address), the names are up to discussion here.
- In LIP 48 replace 

```python
Token.burn(address, tokenID, amount),
```
by
```python
if BURN_FEES:
    Token.burn(address, tokenID, amount)
else:
    Token.transfer(address, ADDRESS_FEE_POOL, tokenID, amount)
```

# Motivation

For sidechains, it can be beneficial to have the options to collect fees in a pool (in particular, when using `LSK` for fees) and use them to incentivize validators or other protocol participants. 

Note that an issue with the same suggestion was also created in the SDK repository: https://github.com/LiskHQ/lisk-sdk/issues/8028.

mjerkov:
Do these configurable constants have mainchain values (I suppose that `BURN_FEES = false` on mainchain)?
I have no objections on the namings of constants.

janhack:
@shuse2 What would you think is the most elegant approach here for the Lisk mainchain, leave `ADDRESS_FEE_POOL` undefined (not sure if this is possible for configurable constants) or set some default value such as the all zero address?

shuse2:
we can keep it undefined, and implementation wise, we can do like if ADDRESS_FEE_POOL is defined, make transfer otherwise burn

In this way, we don't need the flag

mjerkov:
The option to leave out the `BURN_FEES` flag seems good to me, if @janhack agrees I can open the PR.

janhack:
Sounds good.

PR #223, opened 2023-01-19
Update LIP 0048 - small changes (fee pool account, CCC hooks naming)

mjerkov:
Small changes to the Fee module LIP, with regards to the following issues:

- [Allow transfering fees to account instead of burning them](https://github.com/LiskHQ/lips/issues/209)
- [LIP 0048 uses some wrong function names for some CCC hooks](https://github.com/LiskHQ/lips/issues/224).

### DIFF
--- a/proposals/lip-0048.md
+++ b/proposals/lip-0048.md
@@ -31,7 +31,11 @@ Each chain can configure the token used to pay fees. On the Lisk mainchain, the 
 
 ### Minimum Fee per Transaction
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte. The minimum part of the fee should be burned, this is to avoid that validators can send transactions in the blocks they generate without cost.
+As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte.
+
+The implementation allows the option to transfer minimum fees to the specified pool address `ADDRESS_FEE_POOL`. This could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants.
+
+If `ADDRESS_FEE_POOL` is left undefined, as is the case with the Lisk mainchain, the minimum part of the fee is burned, to avoid that validators can send transactions in the blocks they generate without cost.
 
 ## Specification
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0048.md
```diff
@@ -31,36 +31,41 @@ Each chain can configure the token used to pay fees. On the Lisk mainchain, the
...
+| `EVENT_NAME_INSUFFICIENT_FEE`        | string | "insufficientFee"                      | Event name of the InsufficientFee event.                          |
+| `LENGTH_ADDRESS`                     | uint32 | 20                                     | The length of an address in bytes.                                |
+| `LENGTH_TOKEN_ID`                    | uint32 | 8                                      | The length of a token ID in bytes.                                |
+| **Configurable Constants**           |        | **Mainchain Value**                    |                                                                   |
+| `MIN_FEE_PER_BYTE`                   | uint64 | 1000                                   | Minimum amount of fee per byte required for transaction validity. |
+| `TOKEN_ID_FEE`                       | bytes  | `OWN_CHAIN_ID[0:1] + '00000000000000'` | Token ID of the token used to pay the transaction fees.           |
+| `ADDRESS_FEE_POOL`                   | bytes  | undefined                              | Address of the fee pool.                                          |
```
sergeyshemyakov, line 58: It would be more consistent to set the value to `None` instead of undefined.

proposals/lip-0048.md
```diff
@@ -275,7 +286,7 @@ def beforeCrossChainCommandExecute(trs: Transaction, ccm: CCM) -> None:
 #### After Cross-chain Command Execution
 
 ```python
-def afterCrossChainCommandExecute(trs: Transaction, ccm: CCM) -> None:
+def afterCrossChainCommandExecution(trs: Transaction, ccm: CCM) -> None:
```
sergeyshemyakov, line 289: In this function the `Token.burn` should also be replaced with more complicated logic.

proposals/lip-0048.md
```diff
@@ -238,7 +243,13 @@ def afterCommandExecute(trs: Transaction) -> None:
...
+
+    if ADDRESS_FEE_POOL is not None:
+        # Transfer the fees to the pool account address, if it exists.
+        Token.transfer(senderAddress, ADDRESS_FEE_POOL, TOKEN_ID_FEE, trs.fee - ctx.availableTransactionFee)
+    else:
+        Token.burn(senderAddress, TOKEN_ID_FEE, trs.fee - ctx.availableTransactionFee)
+
     Token.transfer(senderAddress, generatorAddress, TOKEN_ID_FEE, ctx.availableTransactionFee)
 
     emitEvent(
```
sergeyshemyakov, line 255: The term `burntAmount` is not correct anymore. I would replace it with `minimalAmount`, but maybe there are better alternatives.
For a more specific case of DEX I we use the term `sharedAmountValidators`.

sergeyshemyakov (changes requested):

proposals/lip-0048.md
mjerkov: Right!

proposals/lip-0048.md
mjerkov: Thanks for this comment, I was not sure whether this should also be changes. Could it be that `messageFeeTokenID` is different from `TOKEN_ID_FEE` (I suppose yes)? If yes, then the remaining CCM fee should be transferred to a different account address, not the `ADDRESS_FEE_POOL`, right?
If answers to the above questions are yes, then my suggestion for the account address for the remaining CCM fee is `ADDRESS_CCM_FEE_POOL`. Is this OK with you?

proposals/lip-0048.md
mjerkov: Agreed, I will change all instances of the `burntAmount` to `minimalAmount`.

mjerkov:
If `messageFeeTokenID == TOKEN_ID_FEE`, I suppose that all minimal fees (both transaction and CCM) go to the same address. So, i presume there is a flag that needs to be added to the `afterCrossChainCommandExecution` which decides to which address the CCM fee goes. Not sure what is the cleanest way to solve this...

proposals/lip-0048.md
sergeyshemyakov: I think, it is possible to have different token IDs for message fee and for fee. However I think it would make more sense to have a single fee pool for command fees and ccm fees, there is no problem with having different types of tokens in one account.

sergeyshemyakov (changes requested):
I think, the two fee addresses should be united into one. Other than this, looks good!

proposals/lip-0048.md
mjerkov: I wasn't sure about this too... but now that you mention, of course it is so. OK, later during the day I will correct the code.

sergeyshemyakov (approved):
Looks good now

proposals/lip-0048.md
gkoumout: The comment is valid, but `minimalAmount` is not intuitive at all IMO, I would suggest finding something better otherwise keep the old naming.

proposals/lip-0048.md
gkoumout: Indeed there is no problem having different tokens in the same account, it's same as user Alice having in her account LSK tokens and sidechain tokens.

Corner case, that came to my mind, do we have also to initialize user account for this address and specified token? This is related to the topic of initializing accounts, but holds also for the account of `ADDRESS_FEE_POOL`. (perhaps @sergeyshemyakov knows better)

gkoumout (commented):
Good job Miroslav! Left some comments from my side.

proposals/lip-0048.md
mjerkov: Thanks again for the clarification, Greg!

proposals/lip-0048.md
mjerkov: To me it seems that `minimalAmount` is still better than `burntAmount` since, by the modification this PR introduces, the minimal amount of the fee is not necessarily being burnt anymore. 
Maybe it can be simply `minimalFee`?

proposals/lip-0048.md
```diff
@@ -31,36 +31,41 @@ Each chain can configure the token used to pay fees. On the Lisk mainchain, the
 
 ### Minimum Fee per Transaction
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte. The minimum part of the fee should be burned, this is to avoid that validators can send transactions in the blocks they generate without cost.
+As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte.
+
+The implementation allows the option to transfer minimum fees to the specified pool address `ADDRESS_FEE_POOL`. This could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants.
```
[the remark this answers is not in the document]
mjerkov: Right! I understand that "implementation" as a word can be ambiguous and/or problematic, so it's better to avoid it. I like your suggestion and will write something on that line of thought.

proposals/lip-0048.md
sergeyshemyakov: I think the accounts should be automatically initialized in the [Token.transfer](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#transfer-2) function.
I guess it could be annoying from the user perspective if they provide not enough fee, but this is our protocol.

sergeyshemyakov (approved):
Looks good from my side.

proposals/lip-0048.md
gkoumout: ok in fact now that I think again of it, I see the point of using `minimalAmount` or `minimumFee`, it is the minimum required for the transaction to succeed. We could also use `baseFee`/ `baseAmount` or `requiredAmount`, meaning this is the basic fee you need for the transaction, meaning whatever is above this (ctx.availableTransactionFee``), should be treated as "optional". 

Not a big deal, but we should end up with something intuitive.

proposals/lip-0048.md
gkoumout: @sergeyshemyakov that means that in the very first transaction, the user should know about the extra fee needed for initializing the account for this address? And same for the first ccm?

proposals/lip-0048.md
mjerkov: Actually, `requiredAmount` sounds good to me, if @sergeyshemyakov agrees we can use this option.

gkoumout (approved):
Good job Miroslav!

janhack (approved):
Editorial approval

PR #243, opened 2023-01-26
Add config value for determining validator registration function

### LINKED ISSUE #232: LIP 0057 assumes that SDK can dected the re-genesis block Lisk mainchain v4
LIP 0057 uses the following condition in the genesis block processing: 
```python
# For the snapshot block for the migration from Lisk Core 3 to Lisk Core 4 on
# Lisk mainnet, apply a special rule. This must be done as validators in
# this snapshot block do not have a BLS key.
if the chain is the mainchain and b is the snapshot block for the migration:
```
The 2nd part of the condition (`b is the snapshot block for the migration`) can however not be detected by the SDK. And adding the migration height as a config parameter is not seen as nice solution to the SDK team. Instead, a boolean config value is used that indicates if (A) all validators do not have a BLS key or (B) all validators have a valid BLS key. Case (A) will be used for the re-genesis block for Lisk mainchain v4. Case (B) for all other (re)-genesis blocks. LIP 0057 must be fixed such that this config value is used in the if statement instead of "is the snapshot block for the migration".
Note that the condition `if the chain is the mainchain` MUST still be checked to prevent that any other chain calls `registerValidatorWithoutBLSKey` instead of `registerValidatorKeys`.  

Possibly, this config value must also be specified in LIP 0063. Further clarification is needed.

AndreasKendziorra:
We introduce a configurable constant in the PoS module that is used to determine if either `registerValidatorKeys` or `registerValidatorWithoutBLSKey` should be used for registering validators contained in a genesis block. The latter case should one happen for the v4 re-genesis block of the Lisk mainchain.
LIP 0063 is updated as well by specifying a value for this configurable constant.

It fixes #232. 

Additionally, we remove the definition of `ROUND_LENGTH` from the constants table, as it is not a configurable constant and is already defined below the table.


proposals/lip-0063.md
```diff
@@ -298,6 +298,7 @@ Some new LIPs introduced configurable constants. The values for these constants
 | [PoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants) | `NUMBER_STANDBY_VALIDATORS` | 2 |
 | [PoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants) | `TOKEN_ID_POS` | `TOKEN_ID_LSK` |
 | [PoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants) | `VALIDATOR_REGISTRATION_FEE` | 10 x (10^8) |
+| [PoS Module](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#notation-and-constants) | `INVALID_BLS_KEYS_IN_GENESIS_BLOCK` | `True` |
```
sameersubudhi, line 301: The specifics on how to use the constant `INVALID_BLS_KEYS_IN_GENESIS_BLOCK` within the migration tool/logic is missing. Could you please update the same?

sameersubudhi (changes requested):

gkoumout (approved):

proposals/lip-0063.md
AndreasKendziorra: The migrator tool does not need to use this constant. It is only needed by Lisk Core v4. More precisely, the PoS module needs this config value for the genesis block processing.

sameersubudhi (approved):

gkoumout (approved):

shuse2 (approved):

janhack (approved):
Editorial approval

Issue #242, opened 2023-01-25
Missing update on token module and difficult validation logic


shuse2:
### Description

For this [PR](https://github.com/LiskHQ/lips/pull/192), there were some points which are difficult to implement

* https://github.com/LiskHQ/lips/pull/192/files#diff-851f39ba2dc407b594e12baee157421d0efd44abd262b8d24cd1b7d20ae0dd8fR2522-R2523
Here, it's defined for all chain, but it's only called in the mainchain
* https://github.com/LiskHQ/lips/pull/192/files#diff-851f39ba2dc407b594e12baee157421d0efd44abd262b8d24cd1b7d20ae0dd8fR1855-R1858
At this point, it is not possible to fetch what modules are registered. We could check if dependency is resolved or not, but if checked here, it should checked everywhere

Also,
on [transferCrossChain method](https://github.com/LiskHQ/lips/blob/38802c26bc268cc5879ca5ef4ebc9fb8ae7d5f86/proposals/lip-0051.md#transfercrosschain-2), check for the mainchain ID should be removed

PR #244, opened 2023-01-26
Update token module -  small fixes

gkoumout:
### Description

This PR resolves https://github.com/LiskHQ/lips/issues/242. 

In particular in the function `transferCrossChain`:

- The check whether interoperability module is registered gets removed.
- Support for direct LSK transfers between sidechains is removed.


ricott1 (approved):

sergeyshemyakov (approved):
Looks good.
I would add a small comment about the first point in the issue to the PR since it is not addressed in the PR. Just say something like: "Note that the function `beforeCrossChainMessageForwarding` is called only on the mainchain", would be more clean although not necessary.

proposals/lip-0051.md
```diff
@@ -1859,10 +1858,6 @@ def transferCrossChain(
         emitFailedCrossChainTransferEvent(senderAddress, tokenID, amount, receivingChainID, recipientAddress, DATA_TOO_LONG)
         raise Exception("Data field too long.")
 
-    if Interoperability module is not registered:
```
AndreasKendziorra, line 1862: So when we remove this check, what actually happens if a custom module calls `transferCrossChain` when in the interoperability module is not registered? More precisely, what happens `Interoperability.send` in `transferCrossChainInternal` is called?

proposals/lip-0051.md
```diff
@@ -1879,9 +1874,9 @@ def transferCrossChain(
             raise Exception("Insufficient sender available balance")
 
     # Transfer is only possible for tokens native to either sending, receiving or mainchain.
-    if tokenChainID not in [OWN_CHAIN_ID, receivingChainID, getMainchainID()]:
+    if tokenChainID not in [OWN_CHAIN_ID, receivingChainID]:
```
AndreasKendziorra, line 1877: The comment in the line above must be updated as well.

AndreasKendziorra (changes requested):

proposals/lip-0051.md
shuse2: it will fail, but it wont produce event

shuse2 (approved):

proposals/lip-0051.md
gkoumout: Actually it will fail in `messageFeeTokenID = Interoperability.getMessageFeeTokenID(receivingChainID)` (line 1865); no event will be emitted and `transferCrossChainInternal` will not be called at all.

proposals/lip-0051.md
gkoumout: good catch, fixed.

proposals/lip-0051.md
AndreasKendziorra: ok. I think an additional comment that this function will fail if interoperability module is not registered would be nice, but it's not a must.

AndreasKendziorra (approved):

proposals/lip-0051.md
gkoumout: I am not adding it, because there might be other places were a similar comment should be added, so prefer to skip it for consistency.

janhack (approved):
Editorial approval

PR #246, opened 2023-01-31
Fix issues in LIP-0044

### LINKED ISSUE #241: LIP 0044 defines unused substore prefix
In LIP 0044, there is a `SUBSTORE_PREFIX_GENESIS_DATA` substore prefix defined. However it is not used in the LIP. 
I think, this constant should be removed from the LIP.

AndreasKendziorra:
This PR fixes the following in LIP 0044:
- Remove the unused constant `SUBSTORE_PREFIX_GENESIS_DATA` from the constants table.
- Define `generatorList` in `getGeneratorsBetweenTimestamps`, as the definition was removed by [PR 177](https://github.com/LiskHQ/lips/pull/177).

It fixes #241 .

### DIFF
--- a/proposals/lip-0044.md
+++ b/proposals/lip-0044.md
@@ -533,7 +533,7 @@ def getValidatorParams() -> ValidatorParams:
 
 #### getGeneratorsBetweenTimestamps
 
-This function returns the addresses of the generators active between the two input timestamps and the number of block slots assigned to them. The slots corresponding to the input timestamps are NOT counted, only the slots in between. Notice that for all counted slots, the current value of `validatorParamsStore.validators` must be the used generator list. Otherwise, the behaviour is undefined. It’s the calling module’s responsibility to ensure that.
+This function returns the addresses of the generators active between the two input timestamps and the number of block slots assigned to them. The slots corresponding to the input timestamps are NOT counted, only the slots in between. Notice that the input timestamps must be recent enough sucht that the active validator set for the counted slots is the one stored `validatorParamsStore.validators`. If the input timestamps are older, the result will most likely be incorrect. It’s the calling module’s responsibility to ensure that.
 
 ```python
 def getGeneratorsBetweenTimestamps(startTimestamp: uint32, endTimestamp: uint32) -> dict[Address, uint32]:
--- a/proposals/lip-0044.md
+++ b/proposals/lip-0044.md
@@ -8,7 +8,7 @@ Discussions-To: https://research.lisk.com/t/introduce-validators-module/317
 Status: Draft
 Type: Standards Track
 Created: 2021-08-06
-Updated: 2023-01-30
+Updated: 2023-02-02
 Requires: 0040
 ```
 
--- a/proposals/lip-0044.md
+++ b/proposals/lip-0044.md
@@ -533,7 +533,7 @@ def getValidatorParams() -> ValidatorParams:
 
 #### getGeneratorsBetweenTimestamps
 
-This function returns the addresses of the generators active between the two input timestamps and the number of block slots assigned to them. The slots corresponding to the input timestamps are NOT counted, only the slots in between. Notice that for all counted slots, the current `generatorList` must be the used generator list. Otherwise, the behaviour is undefined. It’s the calling module’s responsibility to ensure that.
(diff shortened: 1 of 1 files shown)


Incede (approved):

proposals/lip-0044.md
```diff
@@ -534,7 +533,7 @@ def getValidatorParams() -> ValidatorParams:
 
 #### getGeneratorsBetweenTimestamps
 
-This function returns the addresses of the generators active between the two input timestamps and the number of block slots assigned to them. The slots corresponding to the input timestamps are NOT counted, only the slots in between. Notice that for all counted slots, the current `generatorList` must be the used generator list. Otherwise, the behaviour is undefined. It’s the calling module’s responsibility to ensure that.
+This function returns the addresses of the generators active between the two input timestamps and the number of block slots assigned to them. The slots corresponding to the input timestamps are NOT counted, only the slots in between. Notice that for all counted slots, the current value of `validatorParamsStore.validators` must be the used generator list. Otherwise, the behaviour is undefined. It’s the calling module’s responsibility to ensure that.
```
gkoumout, line 536: I find this paragraph a bit confusing. Do we want to say that a module (e.g.,PoA or PoS) calling this function should ensure that the validators in the validator params substore are aligned to its current validators? If yes I would explicitly state it like that; currently it does not seem clear to me what "must be the used generator list" means and what we mean by undefined behavior.

proposals/lip-0044.md
```diff
@@ -54,7 +54,6 @@ We define the following constants:
 | `SUBSTORE_PREFIX_VALIDATORS_KEYS` | bytes | 0x0000 | Substore prefix of the validators keys substore. |
 | `SUBSTORE_PREFIX_VALIDATOR_PARAMS` | bytes | 0x4000 | Substore prefix of the validator params substore. |
 | `SUBSTORE_PREFIX_BLS_KEYS` | bytes | 0x8000 | Substore prefix of the registered BLS keys substore. |
-| `SUBSTORE_PREFIX_GENESIS_DATA` | bytes | 0xc000 | Substore prefix of the genesis data substore. |
```
gkoumout, line 57: Line 41: I don't understand what we mean that we check the existence of a BLS key in constant time. I guess it is logarithmic in the number of validators, instead of linear that it would be by checking all addresses if we did not have the registered BLS keys substore. Am I missing something?

gkoumout (commented):
The content being updated looks good to me. I just marked a couple of places where syntax can be improved IMO.

proposals/lip-0044.md
janhack: The basic assumption here is that state store reads are O(1) as it is basically a hash map (the actual rocks DB implementation uses caching and other advanced techniques so maybe this is a simplification). However, with this assumption checking if a BLS key is registered can be done in O(1) utilizing the "Registered BLS Keys Substore". This store contains redundant date in order to achieve this performance.

proposals/lip-0044.md
janhack: Suggestion:

> Notice that this function assumes that `validatorParamsStore.validators` is the list of generators used in the engine to assign validators to block slots in a round-robin fashion (see function getGeneratorAtTimestamp in [LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#getgeneratorattimestamp)).

janhack (changes requested):
The specification fixes are all good. I would still improve the explanation of the function `getGeneratorsBetweenTimestamps`.

proposals/lip-0044.md
gkoumout: Ok I think I see now. You assume that you use hashing therefore you can look up in expected time O(1). So it's just that you use the PublicKeyBLS as a key (instead of e.g., address) that allows you to achieve this? Otherwise by just scanning over all addresses to check for a specific BLS you would need \Omega(n) anyways. Is this the punchline?

AndreasKendziorra: > So it's just that you use the PublicKeyBLS as a key (instead of e.g., address) that allows you to achieve this?
 
Yes. One does not need to scan through all addresses. Only one look up a given store key (the BLS key).

proposals/lip-0044.md
AndreasKendziorra: I updated the description. The point is that the input timestamps can, for instance, NOT be chosen that they refer to some rounds in the past where older validator set was used. The function only works for time stamps since the latest change of  `validatorParamsStore.validators`.

gkoumout (approved):

janhack (approved):

Issue #226, opened 2023-01-19
Make validation and encoding / decoding for params explicit in LIPs


janhack:
In the SDK, during the `Command verification` hook for a transaction, a schema validation of the `params` property is performed. This is currently not mentioned in the specifications. Moreover, the encoding of the `params` is not explicitly defined in LIPs at the moment. The same holds for the `params` property of CCMs.

janhack:
We concluded on performing the following changes. Note that this is not a protocol change (the implementation handles the topic already as desired), but just for improving the clarity of the specifications. 

- We define a function `validateAgainstSchema(schema, object)` for schema validation (name to be fixed).
- We re-define the `Transaction` type to have `params` as type `bytes`.
- We re-define the `CCM` type to have `params` as type `bytes`.
- In the `verify` hook for every command and cross-chain command we decode and validate the `params` format.
- In the `execute` hook for every command and cross-chain command  we decode the `params`.
- We fix the functions encodeCCM/decodeCCM and functions encodeTransaction/decodeTransaction to not encode/decode `params`.
- We add to the [Validation section of LIP 68](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0068.md#validation) the requirement that in the `Command.verify` hook the `params` property needs to be validated.
- We add the requirement that in the `CCCommand.verify` hook of every CCM the `params` property needs to be validated to LIP 49.
- We need to add the root CCM schema validation to [`validateFormat`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#validateformat).

janhack:
As part of this issue, I would also either define a type `Transaction` in all LIPs where this type is used, or, alternatively, conclude that we accept that this type is globally defined and does not need to be redefined everywhere.

janhack:
I added the `protocol change` label as we add the certificate schema validation to the` verifyCertificate` function in LIP 0053. This is the only protocol change, however.

PR #255, opened 2023-02-10
Make validation and encoding / decoding for params explicit in LIPs

janhack:
The PR makes the following LIP modifications.

## LIP 27

- [x] Define function `validateObjectSchema(schema, object)` for schema validation.

## LIP 45

- [x]  LIP 45 already defines `CCM` type to have `params` as type `bytes`.

## LIP 49

- [x] Fix the functions `encodeCCM` / `decodeCCM`  to not encode/decode `params`.
- [x] Add the requirement that in the `CCCommand.verify` hook of every CCM the `params` property needs to be validated.
- [x] Add the root CCM schema validation to [`validateFormat`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#validateformat).

## LIP 53

- [x] Add schema validation to the `verifyCertificate` certificate function.

## LIP 68

- [x] Re-define the `Transaction` type to have `params` as type `bytes`.
- [x] Fix the functions `encodeTransaction` / `decodeTransaction` to not encode/decode `params`.
- [x] Add to the [Validation section of LIP 68](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0068.md#validation) the requirement that in the `Command.verify` hook the `params` property needs to be validated.

## In all module LIPs

In the `verify` hook for every command and cross-chain command we decode and validate the `params` format.
In the `execute` hook for every command and cross-chain command  we decode the `params`.

- [x] Auth module
- [x] Dynamic Rewards module (no commands, nothing todo)
- [x] Fee module (no commands, nothing todo)
- [x] Interoperability module
- [x] Legacy module
- [x] NFT module
- [x] PoA module
- [x] PoS module
- [x] Random module  (no commands, nothing todo)
- [x] Token module
- [x] Validators module

Resolves https://github.com/LiskHQ/lips/issues/226


janhack:
Note that the PR is not complete as not all modules all updated. The primary goal of this first review is to align on the changes and the way to implement them for the module LIPs. I only implemented the changes fully for the Token module. 

In particular, I wanted to get your feedback on doing:
```
trs.params = decode(crossChainTransferParamsSchema, trs.params)
```
This is maybe not clean as `trs` now does not have a well-defined type, but it allows to keep the specifications mostly unchanged.

proposals/lip-0053.md
```diff
@@ -128,6 +128,7 @@ def verifyLivenessConditionForRegisteredChains(ccu: Transaction) -> None:
 def verifyCertificate(ccu: Transaction) -> None:
     certificateBytes = ccu.params.certificate
```
AndreasKendziorra, line 129: here, we also still assume that `params` is of type object. If this PR intends to change the definition of the type _Transaction_ globally, then it must also be changed here.

There seem to be some more occurrences in this file.

proposals/lip-0068.md
```diff
@@ -156,10 +156,7 @@ Consider a binary message `trsMsg`, corresponding to a serialized transaction. T
 
 ```python
 def decodeTransaction(trsMsg: bytes) -> Transaction:
-    trsData = decode(transactionSchema,trsMsg)
-    paramsSchema = JSON schema corresponding to (trsData.module, trsData.command) pair
-    trsData.params = decode(paramsSchema,trsData.params)
-    return trsData
+    return decode(transactionSchema, trsMsg)
```
AndreasKendziorra, line 159: This function is not used anywhere. We could remove it.

proposals/lip-0049.md
```diff
@@ -206,12 +204,15 @@ Consider a binary message `ccmBytes`, corresponding to a serialized cross-chain
 
 ```python
 def decodeCCM(ccmBytes: bytes) -> CCM:
```
AndreasKendziorra, line 206: This function is not used anywhere. We could remove it.

AndreasKendziorra (changes requested):

AndreasKendziorra:
> Note that the PR is not complete as not all modules all updated. The primary goal of this first review is to align on the changes and the way to implement them for the module LIPs. I only implemented the changes fully for the Token module.
> 
> In particular, I wanted to get your feedback on doing:
> 
> ```
> trs.params = decode(crossChainTransferParamsSchema, trs.params)
> ```
> 
> This is maybe not clean as `trs` now does not have a well-defined type, but it allows to keep the specifications mostly unchanged.

This approach looks good to me. If we want to be strict about the type, then we could create a new object in which the `params` property is of type _object_. But I would say that this is rather too much.

proposals/lip-0051.md
```diff
@@ -407,6 +407,9 @@ transferParamsSchema = {
 
 ```python
 def verify(trs: Transaction):
+    trs.params = decode(transferParamsSchema, trs.params)
```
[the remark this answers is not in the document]
janhack: The new definition is supposed to be used globally. I would actually propose to remove the type definitions for `Transaction` everywhere except for in LIP 68 and the type definitions for `CCM` except for LIP 45. These types are just fundamental building blocks of our blockchain and I think redefining them everywhere just leads to extra maintenance cost without any benefit.

proposals/lip-0053.md
janhack: I only applied all the changes to the Token module so far, not to Interoperability yet. Once we are aligned, I would also add the decoding step here (or we assume in this LIP that the type is actually CCU and not Transaction).

proposals/lip-0068.md
janhack: I would keep it. We actually use `encodeTransaction` so I would just keep the reverse function for consistency.

proposals/lip-0049.md
janhack: I would also keep this function for consistency.

proposals/lip-0049.md
```diff
@@ -636,15 +644,17 @@ def verify(ccu: Transaction, ccm: CCM) -> None:
...
+
     if terminatedStateAccount(ccm.chainID) exists:
         if terminatedStateAccount(ccm.chainID).initialized == True:
             return
-        terminatedStateAccount(trs.params.chainID).stateRoot = ccm.params.stateRoot
-        terminatedStateAccount(trs.params.chainID).mainchainStateRoot = EMPTY_HASH
-        terminatedStateAccount(trs.params.chainID).initialized = True
+        terminatedStateAccount(ccu.params.chainID).stateRoot = ccmParams.stateRoot
+        terminatedStateAccount(ccu.params.chainID).mainchainStateRoot = EMPTY_HASH
+        terminatedStateAccount(ccu.params.chainID).initialized = True
```
janhack, line 655: Please check the lines above as there was a bug before (`trs` was not defined).

proposals/lip-0049.md
```diff
@@ -532,11 +533,14 @@ A registration message is created by the Interoperability module when registerin
...
-def verify(ccu: Transaction, ccm: CCM) -> None:
+def verify(ccu: CCU, ccm: CCM) -> None:
+    ccmParams = decode(registrationCCMParamsSchema, ccm.params)
+    validateObjectSchema(registrationCCMParamsSchema, ccmParams)
+
     if chainAccount(ccm.sendingChainID) does not exist:
         raise Exception("Registration message must be sent from a registered chain.")
 
-    if ccm.sendingChainID != ccu.sendingChainID:
+    if ccm.sendingChainID != ccu.params.sendingChainID:
```
janhack, line 543: There was a bug in the row above before.

proposals/lip-0052.md
```diff
@@ -456,10 +457,10 @@ Transactions executing this command have:
 
 ##### Parameters Schema
 
-The `params` property of a cross-chain NFT transfer transaction follows the `crossChainTransferParamsSchema` schema.
+The `params` property of a cross-chain NFT transfer transaction follows the `crossChainNFTTransferParamsSchema` schema.
 
 ```java
-crossChainTransferParamsSchema = {
+crossChainNFTTransferParamsSchema = {
```
janhack, line 463: I renamed this schema to avoid having the same schema name as in LIP 51.

proposals/lip-0052.md
```diff
@@ -577,10 +581,10 @@ Cross-chain messages executing this cross-chain command have:
 
 ##### CCM Parameters Schema
 
-The `params` property of a cross-chain NFT transfer message follows the `crossChainTransferMessageParamsSchema`.
+The `params` property of a cross-chain NFT transfer message follows the `crossChainNFTTransferMessageParamsSchema`.
 
 ```java
-crossChainTransferMessageParamsSchema = {
+crossChainNFTTransferMessageParamsSchema = {
```
janhack, line 587: Same as above.

proposals/lip-0053.md
```diff
@@ -125,9 +125,10 @@ def verifyLivenessConditionForRegisteredChains(ccu: Transaction) -> None:
 #### verifyCertificate
 
 ```python
-def verifyCertificate(ccu: Transaction) -> None:
+def verifyCertificate(ccu: CCU) -> None:
     certificateBytes = ccu.params.certificate      
     certificate = decode(certificateSchema, certificateBytes)
+    validateObjectSchema(certificateSchema, certificate)
```
janhack, line 131: @ricott1 Is this implemented like this already? In principle, this is a protocol change here.

ricott1: It looks to me that the schema check is not done here: https://github.com/LiskHQ/lisk-sdk/blob/bf04175f212550492b0b360972ddf1a4d5dabf31/framework/src/modules/interoperability/base_interoperability_internal_methods.ts#L378

Notice however that in principle we should already check in `verifyLivenessConditionForRegisteredChains` (which is also not done), or maybe move the certificate verification to the top, although maybe it would not be so nice in the SDK implementation, as it is only done in the mainchain. @shuse2 can maybe pitch in here as well.

proposals/lip-0045.md
```diff
@@ -265,6 +265,7 @@ EMPTY_CCM = {
 | `ChainID` | bytes | Must be of length `CHAIN_ID_LENGTH`. | ID of a chain. |
 | `TokenID` | bytes | Must be of length `TOKEN_ID_LENGTH`. | Used for token identifiers. |
 | `CCM` | object | Must follow the [`crossChainMessageSchema`][lip-0049#ccmschema] schema. | The type of cross-chain messages. |
+| `CCU` | object | Must follow the `transactionSchema` schema, except that the `params` property satisfies `crossChainUpdateTransactionParams` instead of being of type `byte`. | The type of a cross-chain update transaction with deserialized `params` property. |
```
ricott1, line 268:
```suggestion
| `CCU` | object | Must follow the `transactionSchema` schema, except that the `params` property satisfies `crossChainUpdateTransactionParams` instead of being of type `bytes`. | The type of a cross-chain update transaction with deserialized `params` property. |
```

proposals/lip-0048.md
```diff
@@ -282,11 +280,11 @@ def afterCommandExecute(trs: Transaction) -> None:
 #### Before Cross-chain Command Execution
 
 ```python
-def beforeCrossChainCommandExecution(trs: Transaction, ccm: CCM) -> None:
+def beforeCrossChainCommandExecution(ccu: Transaction, ccm: CCM) -> None:
```
ricott1, line 283: This is fine as well, but I find this to be a cosmetic change at this point.

proposals/lip-0053.md
```diff
@@ -115,7 +115,7 @@ The certificate schema `certificateSchema` is defined in [LIP 0061][lip-0061#cer
 #### verifyLivenessConditionForRegisteredChains
 
 ```python
-def verifyLivenessConditionForRegisteredChains(ccu: Transaction) -> None:
+def verifyLivenessConditionForRegisteredChains(ccu: CCU) -> None:
      certificate = decode(certificateSchema, ccu.params.certificate)
```
ricott1, line 119: In principle we should validate the schema here, as this is the first time we decode the certificate bytes

proposals/lip-0068.md
```diff
@@ -137,17 +137,17 @@ For a transaction `trs` to be valid, it must satisfy the following:
...
+The checks above are performed in the [static validation stage][lip-0055#block-processing] of the block processing.
+
+Further, it is the responsibility of every module to check in the "Command Verification" hook of a command that the bytes in `trs.params` satisfy the parameter schema of the respective command. 
 
 ### Serialization
 
 The serialization of an object of type `Transaction` is described in the following pseudocode.
 
 ```python
 def encodeTransaction(trs: Transaction) -> bytes:
```
ricott1, line 149: These functions now are pretty redundant, but still ok to keep them

ricott1 (changes requested):
Some minor comments, you can disregard most of them. One non-trivial comment is about validating certificate bytes when processing a ccu

proposals/lip-0049.md
```diff
@@ -572,7 +576,8 @@ def verify(ccu: Transaction, ccm: CCM) -> None:
 ##### Execution
 
 ```python
-def execute(ccu: Transaction, ccm: CCM) -> None:
+def execute(ccu: CCU, ccm: CCM) -> None:
+    ccmParams = decode(registrationCCMParamsSchema, ccm.params)
```
AndreasKendziorra, line 581: it seems that we don't need to decode the params in this function

proposals/lip-0049.md
AndreasKendziorra: `ccmParams.chainID` in lines 650, 651 and 653-655

proposals/lip-0051.md
```diff
@@ -202,8 +202,6 @@ We further use the utility function `getMainchainID()` defined in [LIP 0037][lip
 | `Module ` | string | Must be of length at least `MIN_MODULE_NAME_LENGTH` and at most `MAX_MODULE_NAME_LENGTH`.| Used for identifying modules.|
 | `TokenID` | bytes | Must be of length `TOKEN_ID_LENGTH`. | Used for token identifiers. |
 | `ChainID` | bytes | Must be of length `CHAIN_ID_LENGTH`. | Used for chain identifiers. |
-| `Transaction` | object | Must follow the `transactionSchema` schema defined in the [LIP 0068][lip-0068] with the only difference that `params` property is not serialized and contains the values of parameters of `paramsSchema` for the corresponding transaction. | An object representing a non-serialized transaction. |
-| `CCM` | object | Must follow the [`crossChainMessageSchema` schema defined in LIP 0049][lip-0049#ccmschema]. | The type of cross-chain messages. |
```
AndreasKendziorra, line 206: Also here we my consider to keep the definition

proposals/lip-0052.md
```diff
@@ -197,8 +197,6 @@ The following constants are used throughout the document:
 | `ChainID` | bytes | Must be of length `LENGTH_CHAIN_ID`. | Used for chain identifiers. |
 | `CollectionID` | bytes | Must be of length `LENGTH_COLLECTION_ID`. | Used for NFT collection identifiers. |
 | `AttributesArray` | (Module &#124; bytes)[] | Two-dimensional array consisting of `Module` names and corresponding `attributes`. | Used for array of NFT attributes. |
-| `Transaction` | object | Must follow the `transactionSchema` schema defined in the [LIP 0068][lip-0068] with the only difference that `params` property is not serialized and contains the values of parameters of `paramsSchema` for the corresponding transaction. | An object representing a non-serialized transaction. |
-| `CCM` | object | Must follow the `crossChainMessageSchema` schema defined in [LIP 0049][lip-0049#ccmschema]. | The type of cross-chain messages. |
```
AndreasKendziorra, line 201: Also here we my consider to keep the definition

AndreasKendziorra (changes requested):

proposals/lip-0049.md
AndreasKendziorra: Now I saw that it was already changed. All good

proposals/lip-0053.md
AndreasKendziorra: I guess it is not important in which order it is done. If the serialized certificate cannot be decoded, then `verifyLivenessConditionForRegisteredChains` will throw. If it can be decoded but the schema is obeyed, then `verifyCertificate` does throw.

proposals/lip-0049.md
janhack: I assume this can now be resolved then.

proposals/lip-0053.md
janhack: I would still find it nicer to have `verifyCertificate(ccu)` before `verifyLivenessConditionForRegisteredChains(ccu)` as it is not good practice to work with an input object that is not fully validated and only do a validation later.

I would, however, not want to make this change as part of this PR as it is out-of-scope.

Therefore, my suggestion would be that you open any issue that you see necessary for Interoperability and we could resolve this discussion here.

proposals/lip-0053.md
janhack: I would agree that we should validate the schema as early as possible, see also discussion above.

proposals/lip-0068.md
janhack: I would favor to keep them.

proposals/lip-0048.md
```diff
@@ -62,8 +62,6 @@ We use the definition of the following types:
...
 | ------------- | ------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- |
-| `Transaction` | object | Must follow the `transactionSchema` schema defined in [LIP 0068](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0068.md) with the only difference that `params` property is not serialized and contains the values of parameters of `paramsSchema` for the corresponding transaction. | An object representing a non-serialized transaction. |
-| `CCM`         | object | Must follow the `crossChainMessageSchema` schema defined in [LIP 0049][lip-0049#ccmschema].                                                                                                                                                                                                   | The type of cross-chain messages.                    |
```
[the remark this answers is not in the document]
janhack: As we just have one type `CCM` that is the same everywhere and I don't see any risk of confusing about it, I would prefer to avoid this redundancy in LIPs.

proposals/lip-0051.md
janhack: As we just have one type `CCM` that is the same everywhere and I don't see any risk of confusing about it, I would prefer to avoid this redundancy in LIPs.

proposals/lip-0052.md
janhack: As we just have one type `CCM` that is the same everywhere and I don't see any risk of confusing about it, I would prefer to avoid this redundancy in LIPs.

proposals/lip-0049.md
janhack: I removed this line.

proposals/lip-0048.md
janhack: I updated the PR and now use universally the variable `ccu` for type `CCU` and `trs` for type `Transaction`.

proposals/lip-0053.md
ricott1: The only problem is that in the implementation the common verification is done together, so it would be a bit awkward to interpose the verification for liveness done only on mainchain. I can create the issue in the meanwhile

ricott1 (approved):

proposals/lip-0049.md
AndreasKendziorra: Just to be sure: will someone from SDK team double check that the implementation is indeed using the correct ID?

proposals/lip-0053.md
AndreasKendziorra: can be resolved

proposals/lip-0048.md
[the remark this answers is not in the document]
AndreasKendziorra: ok

AndreasKendziorra (approved):

proposals/lip-0053.md
janhack: As discussed, @ricott1 will still check this point.

proposals/lip-0049.md
janhack: As discussed, @ricott1 will still check this point.

PR #262, opened 2023-02-16
Do not call validator module from consensus in LIP 0055

### LINKED ISSUE #253: Engine calls validators module in LIP 0055
In [this line](https://github.com/LiskHQ/lips/blob/4fbe9c41360311e71baf8a3dde5edb5b997e8bc8/proposals/lip-0055.md?plain=1#L663), the engine calls the validators module to get some data (`generatorKey`). This is, however, not possible with the current architecture. Moreover, the implementation differs from the LIP in this aspect: it gets the data [from the BFT store](https://github.com/LiskHQ/lisk-sdk/blob/f45f5fa24858b133841953cbe61286531391e7ff/framework/src/engine/consensus/consensus.ts#L737-L740). The LIP should be updated to follow the implementation in this regard.

AndreasKendziorra:
During block signature validation (part of consensus domain), call the consensus function `getBFTParameters` instead of the Validators module function `getValidatorKeys` in order to get the generator key for a given generator address.

This fixes #253.


proposals/lip-0055.md
```diff
@@ -660,7 +660,12 @@ The signature is verified by calling the `verifyBlockSignature` function. This f
 
 ```python
 def verifyBlockSignature():
-  generatorKey = validators.getValidatorKeys(block.header.generatorAddress).generatorKey
+  validators = getBFTParameters(block.header.height).validators
+  keys = [v.generatorKey for v in validators if v.address == block.header.generatorAddress]
+  if len(keys) == 0:
```
ricott1, line 665: Duplications shouldn't happen, but we could check like

```python
if len(keys) != 1:
```

to be completely covered.

AndreasKendziorra: Yes, we could. But I wanted to follow the [implementation](https://github.com/LiskHQ/lisk-sdk/blob/f45f5fa24858b133841953cbe61286531391e7ff/framework/src/engine/consensus/consensus.ts#L737-L747) also with regard to the exception message. If we change it, then it may also be required to change the implementation if we want to be strict.

janhack: I would also ensure here that the length is exactly 1, even if that means the implementation has to change.

AndreasKendziorra: I'm not very happy with it as we are adding some (probably negligible) unnecessary computation steps to detect a bug and not to detect some valid states, but I'll do it. Tagging also @shuse2.

shuse2: if duplication doesn't happen, i think no point to add this?
The implementation assumes there is no duplication, and if it happens it should be checked when it's inserted?

ricott1: In my understanding `getBFTParameters` is done in the engine (LIP58), so that it is not customizable. If we can guarantee that no duplication can happen, I agree it's fine to just check `len(keys) == 0`.

AndreasKendziorra: We decided to keep this PR as it is.

@shuse2 we nevertheless found that there is currently not check that ensures that there is no duplication when data is set in consensus. @janhack will create a separate PR in which the checks are added.

janhack (approved):

ricott1 (approved):

proposals/lip-0055.md
janhack: Here is the issue that I created https://github.com/LiskHQ/lips/issues/279.

Issue #263, opened 2023-02-17
Fix issues related to validator selection in PoS module


gkoumout:
Currently in the [PoS module (LIP 57)](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md) there are the following logic mistakes in the validator selection:

 - The [`unbanValidator`  function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#unbanvalidator) does not include the unbanned validator in the eligible delegates even if they meet the requirements. 
 - [`getActiveValidators` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#getactivevalidators) has some flaws for the case of "hybrid" rounds (i.e., the rounds between bootstrap period and full PoS period): First it implicitly assumes that there are always `nbrElectedValidators` eligible validators which might not be the case. If this is not the case the average bft weight of elected validators needs to be calculated differently. Also it has to account for the case where `nbrElectedValidators = 0`.

PR #265, opened 2023-02-17
Fix pos module issues

gkoumout:
This PR resolves https://github.com/LiskHQ/lips/issues/263. 

The following changes are performed in the PoS Module:

 - `unbanValidator` function is updated to add the un-banned validator to the eligible validators if they have the required weight.
 - `getActiveValidators` function is updated to take care of the cases where the size of `electedValidatorAddresses` array is less than `nbrElectedValidators`.
 - Fixes in the referred roundNumber in certain cases where it was inconsistent.
 - Refactor `afterTransactionsExecute` without performing protocol changes, in order to make it easier to check correspondence with implementation.  
 - Update the notation related to snapshot substore in coordination with the implementation team.


proposals/lip-0057.md
```diff
@@ -150,6 +150,7 @@ We also define `ROUND_LENGTH` to be the length of a round, i.e.,  `ROUND_LENGTH
 
 | Name | Type | Validation | Description |
 |------|------|------------|-------------|
+| `BlockHeader` | object   | Must follow the [`blockHeaderSchema`][lip-0055#block-header-json-schema] schema defined in LIP 0055.| The header of a block. |
```
janhack, line 153: Note that the LIP already uses a type `Header`, e.g., in `verifyBlockSignature(header: Header)`. It would be nice to use one consistent type.

proposals/lip-0057.md
```diff
@@ -1470,6 +1475,7 @@ def unbanValidator(address: Address)-> None:
         raise Exception('The specified validator is not banned.')
     if len(validatorStore(address).reportMisbehaviorHeights) < REPORT_MISBEHAVIOR_LIMIT_BANNED:
```
janhack, line 1476: I find it a bit odd then in the case of `len(validatorStore(address).reportMisbehaviorHeights) >= REPORT_MISBEHAVIOR_LIMIT_BANNED` we just do not do anything. I would find it better if for all cases that do not allow unbanning, an exception is thrown and in the success case we unband and update the weight.

proposals/lip-0057.md
```diff
@@ -1910,6 +1916,46 @@ After the transactions in a block `b` are executed, the properties related to mi
...
+    # If there are no eligible validators, validatorList is empty.
+    # In this case, the BFT parameters and validator list are not updated.
+    if validatorList is empty:
+        return
+
+    shuffleAndSetValidators(validatorList, height)
+
+
+
+def updateProductivity(b.header: BlockHeader) -> None:
```
janhack, line 1956: For clarity, could we put each of the functions in a separate code environment.

proposals/lip-0057.md
```diff
@@ -1910,6 +1916,46 @@ After the transactions in a block `b` are executed, the properties related to mi
...
+
+    # if not last block of round no further updates are needed.
+    if isEndOfRound(height) == False:
+        return
+
+    # Block b is an end-of-round block. Need to update snapshot and select validators for next round.
+    # This must be done after the properties related to missed blocks are updated.
+    
+    # No updates in validators and BFT weights are performed during bootstrap period.
+    if roundNumber(height+1) <= genesisDataStore.initRounds:
```
janhack, line 1930: As we are refactoring, can we also improve the function naming. `roundNumber` is not a good name and previously was both used as function name and for local variables. I would suggest `getRoundNumber` or `getRoundNumberFromHeight`.

janhack: Further, I would suggest to once define a local variable
`currentRound = roundNumber(height)` instead of using several function calls here. Then, I would write `currentRound + 1` instead of `roundNumber(height+1)` (I assume this is the same as `isEndOfRound(height) == True`). I would find this more clear.

janhack (changes requested):
A few minor comments and suggestion. Overall, it looks good.

proposals/lip-0057.md
```diff
@@ -1349,8 +1350,12 @@ def getActiveValidators(validatorsTwoRoundsAgo: list[EligibleValidatorObject], r
 
         # Determine active validators and their BFT weight.
         activeValidators = [{"address": item.address, "bftWeight": ceiling(item.weight, WEIGHT_SCALE_FACTOR) } for item in validatorsTwoRoundsAgo[:nbrElectedValidators]]
-        averageElectedValidatorBftWeight = sum([item.bftWeight for item in activeValidators]) // nbrElectedValidators
-        activeValidators+= [{"address": item, "bftWeight": averageElectedValidatorBftWeight} for item in chosenInitValidators]
+        if len(activeValidators) > 0:
+            averageElectedValidatorBftWeight = sum([item.bftWeight for item in activeValidators]) // len(activeValidators)
```
ricott1, line 1354: Question: could this value still be 0 if `activeValidators` contains elements with `bftWeight = 0`?

proposals/lip-0057.md
```diff
@@ -1945,38 +1989,17 @@ def afterTransactionsExecute(b: Block) -> None:
...
     # when being read from the end to the beginning of the store.
 
-    roundNumber = roundNumber(height)
-
     snapshotState = {
             "validatorWeightSnapshot": eligibleValidators
         }
     create an entry in the snapshot substore with
-        storeKey = roundNumber.to_bytes(4,'big'),
+        storeKey = roundNumber(height).to_bytes(4,'big'),
```
ricott1, line 1996: Did we align with the implementation here? Otherwise we could use this PR also to fix the misalignment.

ricott1 (changes requested):
1 comment to make sure the average BFT weight goes as expected, another to align with the SDK implementation

proposals/lip-0057.md
gkoumout: Good catch, fixed.

proposals/lip-0057.md
gkoumout: Good question. In general it could be possible, but for our PoS, to become eligible you need at least weight 1000, therefore the bftweight would be at least 1. So no elements of this array should have `bftWeight = 0`.

proposals/lip-0057.md
gkoumout: I updated the function according to your proposal.

proposals/lip-0057.md
gkoumout: Done.

proposals/lip-0057.md
gkoumout: Good point. I refactored and used also `nextRound = roundNumber(height+1)` which is also used in the implementation, for clarity. Have a look. Of course this can be skipped by replacing `nextRound` by `currentRound+1`.

proposals/lip-0057.md
gkoumout: Yep, it is supposed to be fixed in the implementation by this [PR](https://github.com/LiskHQ/lisk-sdk/pull/8142/commits/72027e7cdee9d697df49f53cb994dfad844a0f15) which I will review today.

proposals/lip-0057.md
```diff
@@ -1984,6 +1984,11 @@ def updateProductivity(b.header: BlockHeader) -> None:
 
 ```python
 def updateSnapshotSubstore(height: uint32)-> None:
+    currentRound = getRoundNumberFromHeight(height)
```
gkoumout, line 1987: For some reason that I can't figure out, the name of this function is not shown in proper color in the LIP. If one of the reviewers knows how to fix this, let me know.

proposals/lip-0057.md
```diff
@@ -1927,15 +1927,16 @@ def afterTransactionsExecute(b: Block) -> None:
...
     # This must be done after the properties related to missed blocks are updated.
-    
-    # No updates in validators and BFT weights are performed during bootstrap period.
     currentRound = getRoundNumberFromHeight(height)
     nextRound = currentRound + 1
-    if nextRound <= genesisDataStore.initRounds:
-        return
 
-    updateSnapshotSubstore(height)
+    updateSnapshotSubstore(height)
```
ricott1, line 1933:
```suggestion
    updateSnapshotSubstore(height)
```

proposals/lip-0057.md
```diff
@@ -1351,6 +1351,7 @@ def getActiveValidators(validatorsTwoRoundsAgo: list[EligibleValidatorObject], r
         # Determine active validators and their BFT weight.
         activeValidators = [{"address": item.address, "bftWeight": ceiling(item.weight, WEIGHT_SCALE_FACTOR) } for item in validatorsTwoRoundsAgo[:nbrElectedValidators]]
         if len(activeValidators) > 0:
+            # average BFT weight can not be 0, since BFT weight of elected validators is at least 1.
```
ricott1, line 1354: I would explicitly mention that this is because `MIN_WEIGHT>=1`

ricott1 (approved):
Approved, left a minor comment to improve one code comment

shuse2 (approved):

proposals/lip-0057.md
gkoumout: Done

proposals/lip-0057.md
```diff
@@ -1928,22 +1928,21 @@ def afterTransactionsExecute(b: Block) -> None:
...
+    nextRound = getRoundNumberFromHeight(height+1)
+
     # No updates in validators and BFT weights are performed during bootstrap period.
     if nextRound <= genesisDataStore.initRounds:
         return
     
     # Validator selection for next round.
     # Calculate the active validators for next round based on the snapshot taken two rounds ago.
-    validatorList = getActiveValidators(snapshotStore(currentRound-2), nextRound)
+    validatorList = getActiveValidators(snapshotStore(nextRound), nextRound)
```
ricott1, line 1941: minor thing, but we could even clean up the interface now, and just pass `nextRound` as a parameter and get the snapshot there directly

proposals/lip-0057.md
```diff
@@ -2005,9 +2005,10 @@ def updateSnapshotSubstore(height: uint32) -> None:
             "validatorWeightSnapshot": eligibleValidators
         }
     create an entry in the snapshot substore with
-        storeKey = currentRound.to_bytes(4,'big'),
+        storeKey = snapshotRound.to_bytes(4,'big'),
         storeValue = encode(snapshotStoreSchema, snapshotState)
-    delete any entries from the snapshot substore snapshotStore(x) for x <= currentRound-3
+
+    delete any entries from the snapshot substore snapshotStore(x) for x <= roundNumber - 1
```
ricott1, line 2011: minor, but I would find easier to read:

```suggestion
delete any entries from the snapshot substore snapshotStore(x) for x < roundNumber
```

ricott1 (approved):
Some comments to improve the code (in my opinion ;) ,you can disregard them if you don't agree

proposals/lip-0057.md
gkoumout: Good catch. It is one of the improvements that we coundn't see in the previous stage. I think this way is cleaner, so I proceeded making this change.

proposals/lip-0057.md
```diff
@@ -1344,10 +1345,12 @@ def isPunished(address: Address, height: uint32) -> bool:
 This function computes the set of active validators and their BFT weight based on the input set of eligible validators.
 
 ```python
-def getActiveValidators(validatorsTwoRoundsAgo: list[EligibleValidatorObject], roundNumber: uint32) -> list[ValidatorObject]:
+def getActiveValidators(roundNumber: uint32) -> list[ValidatorObject]:
     initRounds = genesisDataStore.initRounds
     initValidators = genesisDataStore.initValidators
 
+    validatorsTwoRoundsAgo = snapshotStore(nextRound)
```
janhack, line 1352: - `nextRound` is not defined,
- with the changed definition of the snapshot store, `validatorsTwoRoundsAgo` is also confusing

proposals/lip-0057.md
```diff
@@ -1940,12 +1994,16 @@ def afterTransactionsExecute(b: Block) -> None:
...
     # Update previousTimestamp substore.
     previousTimestamp = b.header.timestamp
+```
 
-    if isEndOfRound(height) == False:
-        return
+```python
+def updateSnapshotSubstore(height: uint32) -> None:
+    roundNumber = getRoundNumberFromHeight(height+1)
+    snapshotRound = roundNumber + 2
```
janhack, line 2002: Is there are specific reason why you don't do 
```
roundNumber = getRoundNumberFromHeight(height)
snapshotRound = roundNumber + 3
```
This would somehow be more canonical in how you define the snapshot store at the beginning. I would also try to avoid to use the generic name `roundNumber` for any round number, but be more specific, i.e., `currentRound`, `nextRound`.

proposals/lip-0057.md
janhack: Looks good.

proposals/lip-0057.md
janhack: I couldn't find any issue.

janhack (changes requested):
Nice work @gkoumout , definitely looks a lot cleaner.

I just have a comment on one type and some cosmetic suggestion.

proposals/lip-0057.md
gkoumout: Good catch.  `nextRound` replaced by `roundNumber` and modified the name of the snapshot validators to `eligibleValidators`.

proposals/lip-0057.md
gkoumout: Replaced `roundNumber` by `nextRound`. The particular choice was made in alignment with the implementation and after discussion with Shu where we agreed that it's better to make clear that at last block of round we select validators for round +1 and the offset of rounds is 2. I agree there are many equivalent ways to define the same thing and I would just leave it as is now.

janhack (approved):
Looks all good.

PR #266, opened 2023-02-17
Fix seedReveal in Reward and Random module

vjaiman:
The PR makes the following LIP modifications.

Closes issue https://github.com/LiskHQ/lips/pull/266
Closes issue https://github.com/LiskHQ/lips/issues/277

## [Reward module (LIP 42)](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0042.md)

- Function `getBlockReward`  use the argument `blockHeader.seedReveal` , which does not exist according to the new block schema. Instead this property is part of the block assets; therefore the block assets should also be arguments of this function. Fixing this requires updating the `isSeedRevealValid` function of the random module where the second argument should be the block assets and the de-serialization happens inside this function. 
- Move getBlockReward() to beforeTransactionExecute()
- Fix header level

## [Dynamic block rewards module (LIP 71)](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0071.md)

- Function `getDynamicBlockReward`  use the argument `blockHeader.seedReveal`, which does not exist according to the new block schema (similar to Reward module). 
- Fix header level

## [Random module (LIP 46)](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0046.md)

 - Fix `isSeedRevealValid` function.
- Inconsistent type hinting: function `H()` has an input bytes. However in `getRandomBytes()` it is called with `uint32` as input. There is a need to transform `int` to `bytes` in the call of `H()`. 
- `AfterTransactionsExecution` is described in an inconsistent way: It checks if "size of the validatorReveals array is larger than `MAX_LENGTH_VALIDATOR_REVEALS`" and if yes, it removes the last element. It should be "if size of array == `MAX_LENGTH_VALIDATOR_REVEALS`" since it should pop the element when it is equal and add a new element afterwards. 
- Description about `MAX_LENGTH_VALIDATOR_REVEALS` is irrelevant here and could be moved to the table.


proposals/lip-0042.md
```diff
@@ -206,4 +210,5 @@ TBA
```
sergeyshemyakov, line 210: In the `afterTransactionsExecute` you should call `getBlockReward` with the block assets as well. Keep in mind the format that you end up choosing for the assets (decoded / not decoded).

proposals/lip-0071.md
```diff
@@ -280,9 +291,9 @@ The Dynamic Block Rewards module does not execute any logic during the genesis s
 Before the transactions of a block `b` are executed, the following logic is executed:
 
 ```python
-def beforeTransactionsExecute(b: Block) -> None:
+def beforeTransactionsExecute(b: Block, blockAssets: BlockAssets) -> None:
```
sergeyshemyakov, line 294: The hook `beforeTransactionsExecute` can have only one argument, `b: Block`. You should obtain the assets as `b.assets`, keep in mind the format that you end up choosing for the assets (decoded / not decoded).

proposals/lip-0042.md
```diff
@@ -135,11 +139,11 @@ This function is used to retrieve the reward of a block at a given height.
...
-def getBlockReward(blockHeader: BlockHeader) -> tuple[uint64, uint32]:
+def getBlockReward(blockHeader: BlockHeader, blockAssets: BlockAssets) -> tuple[uint64, uint32]:
   if not Token.userSubstoreExists(blockHeader.generatorAddress, TOKEN_ID_REWARD):
     return (0, REWARD_REDUCTION_NO_ACCOUNT)
 
-  if Random.isSeedRevealValid(blockHeader.generatorAddress, blockHeader.seedReveal) == False:
+  if Random.isSeedRevealValid(blockHeader.generatorAddress, blockAssets) == False:
     return (0, REWARD_REDUCTION_SEED_REVEAL)
 
   defaultReward = getDefaultRewardAtHeight(blockHeader.height)
```
gkoumout, line 149: minor suggestion: Should we use // for integer division in line 151 and remove line 156?

proposals/lip-0042.md
```diff
@@ -135,11 +139,11 @@ This function is used to retrieve the reward of a block at a given height.
 ##### Execution
 
 ```python
-def getBlockReward(blockHeader: BlockHeader) -> tuple[uint64, uint32]:
+def getBlockReward(blockHeader: BlockHeader, blockAssets: BlockAssets) -> tuple[uint64, uint32]:
```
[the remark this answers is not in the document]
gkoumout: I would not remove Block, just add BlockAssets.

proposals/lip-0071.md
gkoumout: Agree with Sergey. So just use `b.blockAssets` as argument for `getDynamicBlockReward`

proposals/lip-0046.md
```diff
@@ -54,7 +54,7 @@ We define the following constants:
 | `ADDRESS_LENGTH`               | uint32   | 20              | Length in bytes of type Address.                |
 | `SEED_LENGTH`                  | uint32   | 16              | Length in bytes of a valid seed revealed.       |
 | **Config parameters**          |          | Mainchain value |                                                 |
-| `MAX_LENGTH_VALIDATOR_REVEALS` | uint32   | 206             | Maximum length of the `validatorReveals` array. |
+| `MAX_LENGTH_VALIDATOR_REVEALS` | uint32   | 206             | Maximum length of the `validatorReveals` array. It should be set as twice the maximum length of the chains validator set. |
```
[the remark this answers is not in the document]
gkoumout: That's a good point. I think it should be config parameter since it somehow depends on other module's value. It should also be compatible with PoA or whatever consensus module SDK provides (this is why it is written in the form "the maximum length of the chains validator set" and not in PoS-specific language). 

Also I don't think it is necessary to be 2 * length, it's just a recommendation (this we need to ask to make sure it's true). We need it to be 2 *length for the commit-and-reveal process but in general it could have other values. Let's leave it open for now and discuss it again.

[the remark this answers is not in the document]
gkoumout: Good point, I think it means twice the length of a round, so that if a validator committed in the previous round we can always find their commited value (extreme case that in the previous round generated the first block and in the current round generates the last one).

proposals/lip-0046.md
```diff
@@ -268,9 +272,8 @@ Let `blockAssetBytes` be the bytes included in the block asset for the Random mo
 
 After all the transactions of a block `b` are executed, the following logic is applied:
 
-* While the size of the `validatorReveals` array is larger than `MAX_LENGTH_VALIDATOR_REVEALS`, delete the element of `validatorReveals` array with the smallest value of `height`.
+* While the size of the `validatorReveals` array is equal to `MAX_LENGTH_VALIDATOR_REVEALS`, delete the element of `validatorReveals` array with the smallest value of `height`.
```
gkoumout, line 275: I would say "If the size of ... "

proposals/lip-0046.md
```diff
@@ -180,12 +180,12 @@ The Random module exposes the following logic to other modules.
 
 #### isSeedRevealValid
 
-This function assesses the validity of the `seedReveal` property of a block.
+This function assesses the validity of the `seedReveal` property of a block. `seedReveal` is a `SEED_LENGTH` bytes value with the seed revealed by the generator of a certain block.
```
gkoumout, line 183: I would make it a bit more clear here, perhaps mentioning that similar to the `seedReveal` property of `blockHeaderAssetRandomModule` schema.

proposals/lip-0046.md
```diff
@@ -196,7 +196,10 @@ This function returns `True` if `seedReveal` was correctly revealed, otherwise,
 It is specified as:
 
 ```python
-def isSeedRevealValid(generatorAddress: Address, seedReveal: bytes) -> bool:
+def isSeedRevealValid(generatorAddress: Address, blockAssets: BlockAssets) -> bool:
+    # let `blockAssetBytes` be the bytes included in the `blockAssets` for the Random module
```
gkoumout, line 200: Perhaps we should add some explanation on how do we find this particular entry of the `blockAssets` array (check implementation?).

proposals/lip-0042.md
```diff
@@ -127,6 +127,10 @@ The Reward module has the following internal function.
 
 This function is used to retrieve the reward of a block at a given height.
 
+##### Parameters
+* `blockHeader`: header of a block.
+* `blockAssets`: a block property as described in [LIP 55][lip-0055#block-assets].
```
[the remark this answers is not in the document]
gkoumout: I think it make sense to go for the first choice (which is essentially what Vikas does here), since we are in the reward module and we just have access to the block, so it make sense to me to leave the decoding and the search for the particular asset to the random module. But not really strong opinion here.

proposals/lip-0042.md
```diff
@@ -135,11 +139,11 @@ This function is used to retrieve the reward of a block at a given height.
 ##### Execution
 
 ```python
-def getBlockReward(blockHeader: BlockHeader) -> tuple[uint64, uint32]:
+def getBlockReward(blockHeader: BlockHeader, blockAssets: BlockAssets) -> tuple[uint64, uint32]:
   if not Token.userSubstoreExists(blockHeader.generatorAddress, TOKEN_ID_REWARD):
     return (0, REWARD_REDUCTION_NO_ACCOUNT)
 
-  if Random.isSeedRevealValid(blockHeader.generatorAddress, blockHeader.seedReveal) == False:
+  if Random.isSeedRevealValid(blockHeader.generatorAddress, blockAssets) == False:
```
gkoumout, line 146: Hmm now that I look it again, it seems to be a bit problematic to do this check in `afterTransactionsExecute`, because we don't know the order of execution between random and reward modules. In particular, if the random module's `afterTransactionsExecute` has been already executed and the `validatorReveals` array is updated, then this call of `isSeedRevealValid` might not produce a correct result. 

This could be avoided either by moving this to `beforeTransactionsExecute` or perhaps by adding height as a parameter to this function, so that even if an item with height = b.height is already inserted in the `validatorReveals` array, then it could be somehow ignored.  

I think it makes more sense the first choice and then the two reward modules will also be aligned.

gkoumout (changes requested):
Good job Vikas! Left some comments, perhaps we have to dig a bit more in some points to make sure the protocol works as we wish.

proposals/lip-0042.md
```diff
@@ -127,6 +127,10 @@ The Reward module has the following internal function.
 
 This function is used to retrieve the reward of a block at a given height.
 
+##### Parameters
```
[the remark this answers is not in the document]
vjaiman: Removed it. Added the `BlockAssets` in the Types table.

proposals/lip-0042.md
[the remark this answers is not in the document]
vjaiman: I agree with Sergey, block property is not needed since it is not mentioned in any LIP specifications. I added 'BlockAssets'.

proposals/lip-0071.md
```diff
@@ -220,10 +226,15 @@ Here `//` represents integer division.
 
 This function is used to retrieve the block reward for a validator.
 
-#### Returns
+##### Parameters
+* `blockHeader`: header of a block
```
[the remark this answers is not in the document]
vjaiman: removed it.

proposals/lip-0071.md
```diff
@@ -196,14 +196,20 @@ The Dynamic Block Rewards module has the following internal functions.
...
 
 * `amount`: amount of block reward to be minted.
 * `reduction`: an integer indicating whether the reward is reduced and if yes, for which reason. Possible values are `REWARD_NO_REDUCTION`, and `REWARD_REDUCTION_SEED_REVEAL`.
 
+##### Execution
+
 ```python
-def getDynamicBlockReward(blockHeader: BlockHeader)-> tuple[uint64, uint32]:
-    if Random.isSeedRevealValid(blockHeader.generatorAddress, blockHeader.seedReveal) == False:
+def getDynamicBlockReward(blockHeader: BlockHeader, blockAssets: BlockAssets)-> tuple[uint64, uint32]:
```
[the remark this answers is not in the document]
vjaiman: added.

proposals/lip-0042.md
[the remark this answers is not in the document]
vjaiman: Agree with Greg but we can discuss it.

proposals/lip-0042.md
vjaiman: Added, for the moment I am considering it as encoded.

proposals/lip-0071.md
vjaiman: Updated.

proposals/lip-0042.md
vjaiman: Yes, thanks. Updated.

proposals/lip-0046.md
vjaiman: Updated.

proposals/lip-0046.md
vjaiman: Updated.

proposals/lip-0042.md
vjaiman: I think it would be nicer to move it to `beforeTransactionExecute` (similar to dynamic block rewards) but I will give it a more thought and discuss it again.

proposals/lip-0042.md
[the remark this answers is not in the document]
vjaiman: I checked the sdk code, they are getting it as below
asset = blockAssets.getAsset(this._moduleName);

I assume our LIP description is sufficient.

proposals/lip-0046.md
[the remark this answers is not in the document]
vjaiman: In general, it is double the round length.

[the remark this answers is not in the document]
vjaiman: To make it clear, I rephrase it.

proposals/lip-0046.md
vjaiman: We need to add the asset.
https://github.com/LiskHQ/lisk-sdk/blob/5613c8fd6df83132eb7528a90c20fe837375ef0e/framework/src/state_machine/state_machine.ts#L82

I do not think we need an explanation here.

proposals/lip-0046.md
```diff
@@ -196,7 +196,10 @@ This function returns `True` if `seedReveal` was correctly revealed, otherwise,
 It is specified as:
 
 ```python
-def isSeedRevealValid(generatorAddress: Address, seedReveal: bytes) -> bool:
+def isSeedRevealValid(generatorAddress: Address, blockAssets: BlockAssets) -> bool:
+    let `blockAssetBytes` be the bytes included in the `blockAssets` for the Random module
```
gkoumout, line 200: Minor point: Perhaps we should check with implementation how they recover this and do it the same way for clarity.

proposals/lip-0046.md
```diff
@@ -268,9 +272,8 @@ Let `blockAssetBytes` be the bytes included in the block asset for the Random mo
 
 After all the transactions of a block `b` are executed, the following logic is applied:
 
-* While the size of the `validatorReveals` array is larger than `MAX_LENGTH_VALIDATOR_REVEALS`, delete the element of `validatorReveals` array with the smallest value of `height`.
+* If the size of the `validatorReveals` array is equal to `MAX_LENGTH_VALIDATOR_REVEALS`, delete the element of `validatorReveals` array with the smallest value of `height`.
```
gkoumout, line 275: Comment about line 264: I think it is good to add at some place a link to LIP55 definition of assets, since there is not any reference in this LIP; since block assets are not used in so many LIPs, I think it would be good to make easier for the reader to navigate.

gkoumout (changes requested):
Good job Vikas! I still have some open points, but I think we are almost there!

proposals/lip-0042.md
```diff
@@ -37,8 +37,7 @@ In this section, we specify the protocol logic in the lifecycle of a block injec
 
 | **Name**    | **Type** | **Validation**                   | **Description**        |
 |-------------|----------|----------------------------------|------------------------|
-| BlockHeader | object   | see [LIP 0055][block-schema-lip] | The header of a block. |
-| Block       | object   | see [LIP 0055][block-schema-lip] | The block object.      |
+| BlockAssets | object   | see [LIP 0055][lip-0055#block-assets]| a block property   |
```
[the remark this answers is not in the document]
vjaiman: Yes because the block description wasn't in other LIPs also. But I have now updated it according to your suggestion.

proposals/lip-0042.md
```diff
@@ -135,22 +134,20 @@ This function is used to retrieve the reward of a block at a given height.
 ##### Execution
 
 ```python
-def getBlockReward(blockHeader: BlockHeader) -> tuple[uint64, uint32]:
+def getBlockReward(blockHeader: BlockHeader, blockAssets: BlockAssets) -> tuple[uint64, uint32]:
   if not Token.userSubstoreExists(blockHeader.generatorAddress, TOKEN_ID_REWARD):
```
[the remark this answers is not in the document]
vjaiman: Thanks. Updated it.

proposals/lip-0042.md
```diff
@@ -173,22 +170,33 @@ The default reward of the block as a 64-bit unsigned integer.
...
 #### After Transactions Execution
 
 The following function assigns block rewards after all transactions in the block are executed. 
 
 ```python
 def afterTransactionsExecute(b: Block) -> None:
-    (blockReward, reduction) = getBlockReward(b.header)
-    if blockReward > 0:
-        Token.mint(b.header.generatorAddress, TOKEN_ID_REWARD, blockReward)
+    if ctx.blockReward > 0:
```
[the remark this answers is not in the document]
vjaiman: Yes, done.

proposals/lip-0046.md
vjaiman: I already referenced the link about block assets in the table.

proposals/lip-0046.md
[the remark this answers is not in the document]
sergeyshemyakov: I would say that this constant must be at least twice the length of the round for the Random module logic to work correctly.

sergeyshemyakov (approved):
I still have one suggestion to the formulation of the `MAX_LENGTH_VALIDATOR_REVEALS` constant description.

proposals/lip-0046.md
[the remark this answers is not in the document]
vjaiman: @sergeyshemyakov I have already updated it. See here https://github.com/LiskHQ/lips/pull/266#discussion_r1116683969

proposals/lip-0046.md
gkoumout: I am not sure that I understand.

gkoumout (approved):
Approved from my side with two comments

proposals/lip-0071.md
```diff
@@ -73,6 +73,12 @@ We define the following constants:
 
 Furthermore, for the rest of this LIP we indicate with `ctx` the execution context which is passed as extra input to each method call.
 
+### Types
+
+| **Name**    | **Type** | **Validation**                   | **Description**        |
+|-------------|----------|----------------------------------|------------------------|
+| BlockAssets | object   | see [LIP 0055][lip-0055#block-assets]| a block property   |
```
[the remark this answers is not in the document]
vjaiman: Updated.

proposals/lip-0046.md
shuse2: What do you mean by recover?

shuse2 (approved):
Assuming the protocol change is only `getBlockReward` and before/afterTransactionsExecute, it looks good
Also, created relevant SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8194

proposals/lip-0046.md
sergeyshemyakov: I agree with Vikas on that we do not necessarily need to provide all the technical details here. I hope it should be clear enough for the devs.

proposals/lip-0046.md
sergeyshemyakov: I agree with Vikas on that we do not necessarily need to provide all the technical details here. I hope it should be clear enough for the devs.

proposals/lip-0046.md
```diff
@@ -54,7 +54,7 @@ We define the following constants:
 | `ADDRESS_LENGTH`               | uint32   | 20              | Length in bytes of type Address.                |
 | `SEED_LENGTH`                  | uint32   | 16              | Length in bytes of a valid seed revealed.       |
 | **Config parameters**          |          | Mainchain value |                                                 |
-| `MAX_LENGTH_VALIDATOR_REVEALS` | uint32   | 206             | Maximum length of the `validatorReveals` array. |
+| `MAX_LENGTH_VALIDATOR_REVEALS` | uint32   | 206             | Maximum length of the `validatorReveals` array. It should be set as twice the length of a round. |
```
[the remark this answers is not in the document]
vjaiman: I think it is correct but @janhack can confirm this.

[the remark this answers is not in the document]
janhack: I have not been involved in the Random module LIP, but this is the reviewed protocol and I think it works fine so I think it can stay as is.

janhack (approved):
Editorial approval

PR #289, opened 2023-03-01
Add extra rule regarding FastAggregateVerify and INVALID_BLS_KEY

AndreasKendziorra:
This PR does NOT imply any changes of the production code, but it DOES require to add some additional unit tests to the code base.

This PR adds a rule to LIP 0038 which defines the behavior of `FastAggregateVerify` when one of the input public keys is the public key where all bytes are set to zero, which is used in the starting phase of Lisk Core v4.


proposals/lip-0038.md
```diff
@@ -153,12 +159,16 @@ We choose the variant _[minimal-pubkey-size](https://tools.ietf.org/html/draft-i
...
 
+### Why Additional Rule for FastAggregateVerify?
+
+Lisk Core v4 poses an exception to the rule that only public keys, for which a valid proof of possession exists, can be used. More precisely, during the starting phase of Lisk Core v4, there will be validators with the invalid BLS public key where all 48 bytes are set to zero. See [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0063.md#assigning-bls-key-to-validator-accounts) for more details why this is needed. Due to this, it can happen that Lisk Core v4 nodes or even sidechain nodes must evaluate `FastAggregateVerify` where one of the input public keys is the one where all 48 bytes are set to zero. Although the IETF draft does not specify the behavior, the used BLS-library in the Lisk SDK 6 does actually return `INVALID/False` for this specific case. Therefore, it was considered easier to add this extra rule that matches with the implementation to this LIP instead of adding additional rules in the SDK that prevent that `FastAggregateVerify` is called with the key where all 48 bytes are set to zero.
```
ricott1, line 170:
```suggestion
Lisk Core v4 poses an exception to the rule that only public keys for which a valid proof of possession exists can be used. More precisely, during the starting phase of Lisk Core v4, there will be validators with the invalid BLS public key where all 48 bytes are set to zero. See [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0063.md#assigning-bls-key-to-validator-accounts) for more details why this is needed. Due to this, it can happen that Lisk Core v4 nodes or even sidechain nodes must evaluate `FastAggregateVerify` where one of the input public keys is the one where all 48 bytes are set to zero. Although the IETF draft does not specify the behavior, the used BLS-library in the Lisk SDK 6 does actually return `INVALID/False` for this specific case. Therefore, it was considered easier to add this extra rule that matches with the implementation to this LIP instead of adding additional rules in the SDK that prevent that `FastAggregateVerify` is called with the key where all 48 bytes are set to zero.
```

ricott1: Somehow flows better to me, feel free to disregard.

ricott1 (changes requested):
A minor cosmetic comment. I would also add a sentence at the end of the "Why Additional Rule for FastAggregateVerify?" section to explain that this will not be a problem after the end of the bootstrap (init+hybrid rounds) phase, as validators with invalid BLS keys will be banned.

vjaiman (approved):

ricott1 (approved):

shuse2 (approved):
Added SDK issue: https://github.com/LiskHQ/lisk-sdk/issues/8242

PR #292, opened 2023-03-02
Change type CCU to Transaction in several functions

### LINKED ISSUE #284: LIP0049: some cross chain command functions are called with wrong type
1. The interfaces of the `execute` and `verify` functions of cross chain commands require that the first argument is of type `Transaction`. For example, the interface of the verify function is in all existing cases ```def verify(trs: Transaction, ccm: CCM) -> None:```. But they are called with arguments of type `CCU` [here](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0049.md?plain=1#L282), [here](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0049.md?plain=1#L282) and [here](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0049.md?plain=1#L321).
2. The function `verifyCrossChainMessage` in the [Token module LIP](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0051.md?plain=1#L2545) also expects an object of type `Transaction` as the first argument, but when called from LIP 0049 ([here](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0049.md?plain=1#L260) and [here](https://github.com/LiskHQ/lips/blob/abecb5d635e35374526b07ece8162d4d3f4b0245/proposals/lip-0049.md?plain=1#L365)), an object of type `CCU` is provided.

Both ways, either changing the interface or changing the calls, should be fine.

AndreasKendziorra:
We do the following:

- Fix several function calls in LIP 0049, where an argument of type `Transaction` is expected, but an argument of type `CCU` is provided. This includes the functions 
  - `verifyCrossChainMessage`
  - `crossChainCommand.verify`
  - `crossChainCommand.execute`
  - `beforeCrossChainCommandExecution`
  - `afterCrossChainCommandExecution`.
- Change the type of the first argument of the functions `apply` and `forward` from LIP 0049 from `CCU` to `Transaction`. This implies that we still have to decode the `params` property of the transaction (which is a CCU) in an additional step in the `apply` function. But this is better than encoding the `params` property of an object of type `CCU` in `apply` and `forward` in order to get an object of type `Transaction` which is required for the function listed above.

This fixes #284.

### DIFF
--- a/proposals/lip-0049.md
+++ b/proposals/lip-0049.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/introduce-cross-chain-messages/299
 Status: Draft
 Type: Standards Track
 Created: 2021-05-22
-Updated: 2023-02-24
+Updated: 2023-03-06
 Requires: 0053
 ```
 
--- a/proposals/lip-0049.md
+++ b/proposals/lip-0049.md
@@ -243,21 +243,21 @@ def validateFormat(ccm: CCM) -> None:
 
 #### apply
 
-The `apply` function applies a cross-chain message, in the context of a cross-chain update `ccu`.
+The `apply` function applies a cross-chain message, in the context of a cross-chain update transaction `trs`.
 
 <img src="lip-0049/apply.png" width="80%">
 
 _Figure 1: A schematic of the steps executed in the `apply` function. Blue bordered boxes indicate the emission of an event. Functions called from other modules during the CCM execution (indicated as `mdl.<FUNCTION_NAME>`) are repeated for each module in the chain for which the function exists._
 
 ```python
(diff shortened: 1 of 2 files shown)


janhack (approved):
Looks all good.

ishantiw (approved):
LGTM, related SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8223

Issue #279, opened 2023-02-23
Missing check for address and BLS key uniqueness 


janhack:
In the function [setBFTParameters](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#setbftparameters) we do not check whether the provided addresses, generator keys and BLS keys in the parameter `validatorList` are unique. This is guaranteed by the application layer at the moment (i.e., PoA, PoS and Validators module), but for more robustness it would be better to add this check in the engine. With duplicate addresses and BLS keys the behavior of the engine is not well-defined any more and it would be best to raise an exception explicitly.

PR #295, opened 2023-03-03
Add check for address and BLS key uniquess of validators in LIP 0058

janhack:
This PR add the following checks to the function `setBFTParameters` in LIP 0058:

- Check that addresses of provided validators are unique.
- Check that BLS keys of provided validators are unique.

Resolves https://github.com/LiskHQ/lips/issues/279

### DIFF
--- a/proposals/lip-0058.md
+++ b/proposals/lip-0058.md
@@ -490,8 +490,6 @@ def setBFTParameters(precommitThreshold, certificateThreshold, validatorList):
         raise Exception("Given list of validators is larger than the maximum allowed value.")
     if len(validatorList) != len(set([validator.address for validator in validatorList])):
         raise Exception("Duplicate validator address in validator list.")
-    if len(validatorList) != len(set([validator.generatorKey for validator in validatorList])):
-        raise Exception("Duplicate generator key in validator list.")
     if len(validatorList) != len(set([validator.blsKey for validator in validatorList])):
         raise Exception("Duplicate BLS key in validator list.")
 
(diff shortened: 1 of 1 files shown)


shuse2 (approved):
Added in SDK https://github.com/LiskHQ/lisk-sdk/issues/8222

proposals/lip-0058.md
```diff
@@ -488,6 +488,13 @@ The function is called in the after application processing stage of the genesis
 def setBFTParameters(precommitThreshold, certificateThreshold, validatorList):
     if length(validatorList) > LSK_BFT_BATCH_SIZE:
         raise Exception("Given list of validators is larger than the maximum allowed value.")
+    if len(validatorList) != len(set([validator.address for validator in validatorList])):
+        raise Exception("Duplicate validator address in validator list.")
+    if len(validatorList) != len(set([validator.generatorKey for validator in validatorList])):
```
AndreasKendziorra, line 493: I would remove this check. In the application layer, including validators module and PoA/PoS module, we do allow that two validator accounts use the same generator key. Therefore, it could happen that two validator accounts are created with the same generator key, and eventually both become active validators. But when `setBFTParameters` is called during the processing of last block of the round, then `setBFTParameters` throws and the block is invalid although there was in fact no problem with the block. That means, the chain is stuck from there on.

AndreasKendziorra (changes requested):

proposals/lip-0058.md
janhack: Okay, sounds good to me not require generator key uniqueness.

shuse2: ah i didn't know this is allowed, but looks good to me.

shuse2 (approved):

AndreasKendziorra (approved):

AnanyaSshri (approved):

PR #313, opened 2023-03-23
Update endpoint functions and reorder ownChainName check in LIP0045

### LINKED ISSUE #282: LIP0045: Fix `isChainIDAvailable` enpoint and add one for registered names
- The endpoint `isChainIDAvailable` should be moved to a new endpoint section (currently missing). 
- The logic should return False if the input is not on the same network of the mainchain, or if the input equals the mainchain chainID.
- We should add a similar endpoint for registered names `isChainNameAvailable`. This should check if the name has the required length, uses the correct character set, and if it is available.

Related SDK issues:

- https://github.com/LiskHQ/lisk-sdk/issues/8190
- https://github.com/LiskHQ/lisk-sdk/issues/8191

### LINKED ISSUE #298: LIP0045: Reorder checks for `ownChainName`
https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#sidechain

> If chainInfos is not empty, then check that:
> ownChainName is from the character set a-z0-9!@$&_., has length between MIN_CHAIN_NAME_LENGTH and MAX_CHAIN_NAME_LENGTH, and ownChainName != CHAIN_NAME_MAINCHAIN;

Here it needs to be in this order
> If chainInfos is not empty, then check that:
> ownChainName has length between MIN_CHAIN_NAME_LENGTH and MAX_CHAIN_NAME_LENGTH, is from the character set a-z0-9!@$&_., and ownChainName != CHAIN_NAME_MAINCHAIN;

Because if we provide `ownChainName` with an empty string "", regex check will fail.
Expected behaviour was to check against `MIN_CHAIN_NAME_LENGTH`, but in reality regex check was performed.

### LINKED ISSUE #317: NUMBER_ACTIVE_VALIDATORS_MAINCHAIN needs to be NUM_ACTIVE_VALIDATORS_MAINCHAIN
### Description

```NUMBER_ACTIVE_VALIDATORS_MAINCHAIN	uint32	101	The maximum number of mainchain validators.```

Here `NUMBER_ACTIVE_VALIDATORS_MAINCHAIN` should be replacd with `NUM_ACTIVE_VALIDATORS_MAINCHAIN`
Since the prefix `NUMBER` refers to type integer AND this constant has type `uint32`, there is no point to prefix constant with it's data type. But rather, we are referring to "number of something", so better prefix is "NUM"

### Motivation
Constants shouldn't be prefixed with it's data type. since data type is already given in description, hence it's clear for user. 

### Additional Information
https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#notation-and-constants

### PR DESCRIPTION
**Pull request by AnanyaSshri**
In this PR we have done the following changes:

1. Added separate section for endpoint off-chain services and moved `isChainIDAvailable` to the new section. 
2. Introduced `isChainName` endpoint service function and its corresponding checks.
3. Reordered checks for `ownChainName`
4. Replaced notation `NUMBER_ACTIVE_VALIDATORS_MAINCHAIN` with `NUM_ACTIVE_VALIDATORS_MAINCHAIN`

Resolves #282
Resolves #298 
Resolves #317

### CODE REVIEW

File: proposals/lip-0045.md [hunk 1]
```diff
@@ -1411,15 +1411,41 @@ def getTerminatedOutboxAccount(chainID: ChainID) -> TerminatedOutboxAccount:
...
+### Endpoints for Off-Chain Services
+
 #### isChainIDAvailable
 
 ```python
 def isChainIDAvailable(chainID: ChainID) -> bool:
-    if chainAccount(chainID) does not exist:
-        return True
-    return False
+    if chainAccount(chainID) exist:
```
mitsuaki-u, line 1420: should we also have some validation here like the chainID has correct length and format

ricott1: I think in the pseudocode we use the typing to assume implicitly that these checks are valid

ricott1, line 1420: A minor point, but I would move this check to the end, as it requires accessing the state (expensive) while other checks don't (cheap)

proposals/lip-0045.md
```diff
@@ -1411,15 +1411,41 @@ def getTerminatedOutboxAccount(chainID: ChainID) -> TerminatedOutboxAccount:
...
+```
+
+#### isChainNameAvailable
+
+```python
+def isChainIDAvailable(chainName: str) -> bool:
+    if if not re.match(r"^[a-z0-9!@$&_.]+$", chainName):
+        return False
+    
+    if len("chainName") < minLength or len("chainName") > maxLength:
```
ricott1, line 1439:
```suggestion
    if len(chainName) < MIN_CHAIN_NAME_LENGTH or len(chainName) > MAX_CHAIN_NAME_LENGTH:
```

ricott1: `chainName` is a variable, not a string. Also, we should use the already defined constant names.

proposals/lip-0045.md
mitsuaki-u: not only typing but length checks are missing here

ricott1: What I meant is that in the constant table we already say that type `ChainID` has length 4.

ricott1 (changes requested):
Left a couple comments

proposals/lip-0045.md
AnanyaSshri: Done

proposals/lip-0045.md
```diff
@@ -1417,29 +1417,30 @@ def getTerminatedOutboxAccount(chainID: ChainID) -> TerminatedOutboxAccount:
...
-    if chainAccount(chainID) exist:
-        return False
     
     if chainAccount.chainID[0] != getMainchainID()[0]:
        return False
        
     if chainAccount.chainID == getMainchainID():
        return False
+
+    if chainAccount(chainID) exist:
```
ricott1, line 1427:
```suggestion
    if chainAccount(chainID) exists:
```

ricott1 (changes requested):
Left one last small comment

proposals/lip-0045.md
AnanyaSshri: @mitsuaki-u Please let me know if you agree with the reply from alessandro.

proposals/lip-0045.md
```diff
@@ -1580,7 +1607,7 @@ If `chainInfos` is empty, then check that:
 
 If `chainInfos` is not empty, then check that:
 
-* `ownChainName` is from the character set `a-z0-9!@$&_.`, has length between `MIN_CHAIN_NAME_LENGTH` and `MAX_CHAIN_NAME_LENGTH`, and `ownChainName != CHAIN_NAME_MAINCHAIN`;
+* `ownChainName` has length between MIN_CHAIN_NAME_LENGTH and MAX_CHAIN_NAME_LENGTH, is from the character set a-z0-9!@$&_., and ownChainName != CHAIN_NAME_MAINCHAIN;
```
ricott1, line 1610:
```suggestion
* `ownChainName` has length between MIN_CHAIN_NAME_LENGTH and MAX_CHAIN_NAME_LENGTH, is from the character set `a-z0-9!@$&_.`, and ownChainName != CHAIN_NAME_MAINCHAIN;
```

proposals/lip-0045.md
mitsuaki-u: I think it would be nice to also have length check in pseudo code as I recall seeing it in other pseudo codes but I defer to @ricott1 's decision

ricott1: I don't have a strong opinion, for consistency I wouldn't add it. In the implementation do you use a type for `ChainID` or do you check the length? Maybe we can go with what you do there

mitsuaki-u (approved):

ricott1 (approved):

mitsuaki-u (approved):

janhack (approved):
Editorial approval

Issue #311, opened 2023-03-23
Limit `aggregationBits` size and add additional checks to `verifyWeightedAggSig` in LIP 0062


mjerkov:
In `verifyAggSig` and `verifyWeightedAggSig` of [LIP 0062](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0062.md#signing-and-verifying-with-bls) it was noted that `agreggationBits` can in principle be of any size. 

Furthermore, in `certificateSchema` of [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#certificate)
the `agreggationBits` property is lacking the `maxLength` property, which can lead to potentially arbitrarily large certificates.

PR #314, opened 2023-03-23
Limit `aggregationBits` size in `certificateSchema` of LIP 0061 and add additional checks to LIP 0038 and LIP 0062

mjerkov:
Resolves Issue [#311](https://github.com/LiskHQ/lips/issues/311)


AnanyaSshri (approved):

proposals/lip-0062.md
AndreasKendziorra, line 1: I would propose to add the following test for `verifyAggSig` to this file:

```
aggregationBits = 0x4001
signature = 0xb379644423397a99dedea08df6698ef15cb170a93d16ba3d96dbf65ae54b397362333561487b22a105e7e0d471802d5600391d8097154bd86656d323cb62975d0b768c8bec9b1193b482e0210d55dd81a5c36ae1595f3b98f72e66f0d71ffef4
tag = b"LSK_CE_"
chainID = 0x00000000
message = 0xbeaf
verifyAggSig(KEYS_LIST, aggregationBits, signature, tag, chainID, message) == True
```

proposals/lip-0061.md
```diff
@@ -211,7 +212,16 @@ certificateSchema = {
 }
 ```
 
-Here, the `...` notation, borrowed from [JavaScript ES6 data destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring), indicates that the corresponding part of the schema should be inserted in place, and it is just used for notational convenience.
+Here, `MAX_LENGTH_AGGREGATION_BITS` is calculated as `ceiling(MAX_NUM_VALIDATORS, 8)`, where `MAX_NUM_VALIDATORS` is a constant defined in [LIP 0045][lip-0045#notation-and-constants], and the auxiliary function `ceiling` is defined by
+
+```python
+def ceiling(x: int, y: int) -> int:
```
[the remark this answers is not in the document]
mjerkov: Agreed. I wanted to be precise but yes, this is definitely an overkill and I will remove `ceiling`.

proposals/lip-0062.md
mjerkov: Noted.

proposals/lip-0062.md
```diff
@@ -92,6 +96,10 @@ def verifyWeightedAggSig(keysList: list[bytes], aggregationBits: bytes, signatur
     hashedMessage = sha256(taggedMessage)
     keys = []
     weightSum = 0
+
+    if len(aggregationBits) != ceiling(len(keysList, 8)):
```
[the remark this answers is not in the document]
mjerkov: Thanks, I will correct this.

proposals/lip-0062.md
```diff
@@ -79,6 +79,10 @@ def verifyAggSig(keysList: list[bytes], aggregationBits: bytes, signature: bytes
     taggedMessage = tagMessage(tag, chainID, message)
     hashedMessage = sha256(taggedMessage)
     keys = []
+
+    if len(aggregationBits) != ceiling(len(keysList, 8)):
```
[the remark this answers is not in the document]
mjerkov: Thanks, I will correct this.

proposals/lip-0062.md
AndreasKendziorra:
```python
KEYS_LIST = [
  0x9998f02d85e3851a430333350ed6cc1c0afbd72ee52cf8ad2f23d394f3937bfdc92e056dce713b9d45dac7b106d82883,
  0xa491d1b0ecd9bb917989f0e74f0dea0422eac4a873e5e2644f368dffb9a6e20fd6e10c1b77654d067c0618f6e5a7f79a,
  0x8f116ba0b305fb734405dd0968e255ad06a34d0cacfeece4c320502824da4a2ff90a978bfcffa1206ecae27f62bac645,
  0xb301803f8b5ac4a1133581fc676dfedc60d891dd5fa99028805e5ea5b08d3491af75d0707adab3b70c6a6a580217bf81,
  0xb53d21a4cfd562c469cc81514d4ce5a6b577d8403d32a394dc265dd190b47fa9f829fdd7963afdf972e5e77854051f6f,
  0xa6b6a639f7fa0b64ad3a93be965e9cc34e1d9d0f0427c14c38fc80934a937c5fa745a3cb285f64d4d1c06d0825504488,
  0xa4aa20eedb651b7855ee38ce16f59a263346fc383dd9603ac219aaed166ebfe09d460ebbbb7ea89e71c70d48e06efd1a,
  0x95324a8c4a890e8c1e83c96c6c639254937c9c9cee789556606744b07e98292e292c8c150efd9506b0b5547fea3fdf9f,
  0xa424801164381bbfc0b20c1807ce43a12bb012e47deb11b2a3a273dd82ca9fa6364e2f2b8d6c89bc576da89a04d5118f
]
```

mjerkov: I added the `KEYS_LIST` to `verifyAggSig` test.

proposals/lip-0038.md
```diff
@@ -102,6 +102,8 @@ Here, ceiling is an auxiliary function defined below.
 def verifyAggSig(keysList: list[bytes], aggregationBits: bytes, signature: bytes, tag: bytes, chainID: bytes, message: bytes) -> bool:
     taggedMessage = tagMessage(tag, chainID, message)
     keys = []
+    if len(aggregationBits) != ceiling(len(keysList), 8):
+      raise Exception('Length of aggregationBits bitmap is not equal to the ceiling of the length of the keys list divided by 8.')
```
AndreasKendziorra, line 106: I dIdn't notice it in the first round: So far, we don't use exceptions in our BLS functions. Neither in LIP 0038 nor in 0062. So, I would rather return `false` instead. Or what do you think @shuse2 ?

proposals/lip-0061.md
```diff
@@ -211,7 +212,16 @@ certificateSchema = {
 }
 ```
 
-Here, the `...` notation, borrowed from [JavaScript ES6 data destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring), indicates that the corresponding part of the schema should be inserted in place, and it is just used for notational convenience.
+Here, `MAX_LENGTH_AGGREGATION_BITS` is calculated as `ceiling(MAX_NUM_VALIDATORS, 8)`, where `MAX_NUM_VALIDATORS` is a constant defined in [LIP 0045][lip-0045#notation-and-constants], and the auxiliary function `ceiling` is defined by
```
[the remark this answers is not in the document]
AndreasKendziorra: I think it's fine if a constant depends on another constant. So, the value can be written as `(MAX_NUM_VALIDATORS + 8 - 1) // 8`. You could then either mention in the description where `MAX_NUM_VALIDATORS` comes from, or you add `MAX_NUM_VALIDATORS` to the constants table as well.

AndreasKendziorra (changes requested):

AndreasKendziorra (approved):

proposals/lip-0038.md
```diff
@@ -102,6 +102,8 @@ Here, ceiling is an auxiliary function defined below.
 def verifyAggSig(keysList: list[bytes], aggregationBits: bytes, signature: bytes, tag: bytes, chainID: bytes, message: bytes) -> bool:
     taggedMessage = tagMessage(tag, chainID, message)
     keys = []
+    if len(aggregationBits) != ceiling(len(keysList), 8):
+        return False
```
shuse2, line 106: shouldn't we check if there are more bit on aggregationBits, they must be all zero?

mjerkov: @shuse2 I was following Jan's idea "Ensure that aggregation bits is exactly ceil( # of BLS keys / 8) and for all bits without corresponding key the bits must be 0." I suppose you're relating to the second part of his sentence but I wasn't sure why it is necessary: if we check that `len(aggregationBits) == ceiling(len(keysList), 8)`,  how it can be more bits? I'm missing some part of understading here, please explain... thanks

AndreasKendziorra: @shuse2 is right. It could be, for example, that `len(keysList) = 5`. In this case, the "last" 3 bits of the single byte in `aggregationBits` must be zero.

mjerkov: Ah, OK. Thanks. I'm not used to bytes and bits, so it took me some time to understand. I'll push the solution I think works.

mjerkov: @shuse2 @AndreasKendziorra Before I change the code: since there are four occurences of the code I'm about to enter (twice in LIP 0038 and twice in LIP 0062), can you please check if this is correct beforehand?
```
extra_bits = len(aggregationBits) * 8 - len(keysList)
    if extra_bits > 0:
        last_byte = aggregationBits[-1]
    for i in range(extra_bits):
        bit_position = 8 - extra_bits + i
        bit = (last_byte // (2 ** bit_position)) % 2
        if bit != 0:
            return False
```

mjerkov: Wow, thanks. Looks easy :)

AndreasKendziorra (approved):

janhack (approved):
Editorial approval

PR #315, opened 2023-03-24
fix notation in genesis terminated account verification

gkoumout:
This PR fixes the following inconsistency/typo in genesis assets verification:

While verifying the `terminatedStateAccounts` (see schema [here](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#genesis-assets-schema)), for an entry `stateAccount`  in `terminatedStateAccounts` , we use `stateAccount.stateRoot` , `stateAccount.mainchainStateRoot` and `stateAccount.initialized` which is inconsistent. Instead this should be `stateAccount.terminatedStateAccount.stateRoot` and so on. 

Similarly the usage of `chainData` does not correspond to an item in `chainInfos`; it is corrected such that `chainInfo` is an item in `chainInfos` and use `chainInfo.chainData` instead.


ricott1 (approved):
lgtm. Why don't we allow for uninitialized terminated state accounts tho?

PR #316, opened 2023-03-27
add available balance check in stake verify

gkoumout:
last edit April 3. 

This PR was supposed to make the following modification in PoS Module:

> Currently, when a user stakes using the stake command, there is no check of available balance in the transaction verification stage. Instead, an exception and corresponding failure event is emitted once there is attempt to lock the staked amount, if balance is not enough. In this PR, we add this extra check in the verification, so that a transaction does not proceed to execution unless there is sufficient available balance for staking. 

It was finally decided to don't add this check for performance reasons and keep `verify` as is. We just kept some typo corrections in this PR.

### DIFF
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -732,7 +732,7 @@ Let `trsParams = decode(stakeTransactionParams, trs.params)` for the given trans
 * A given `validatorAddress` is included in at most one stake from the list of stakes (regardless of the associated amounts).
 * For all stakes included in `trsParams.stakes`, we have:
   * `amount` value is a multiple of `BASE_STAKING_AMOUNT`, i.e., `amount % BASE_STAKING_AMOUNT == 0`. For the Lisk mainchain, where `BASE_STAKING_AMOUNT = 10^9` and `TOKEN_ID_POS` is the token ID of the LSK token, this corresponds to multiples of 10 LSK.
-  * `amount != 0`.   
+  * `amount != 0`.
 
 ##### Execution
 
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -9,7 +9,7 @@ Discussions-To: https://research.lisk.com/t/define-state-and-state-transitions-o
 Status: Draft
 Type: Standards Track
 Created: 2021-09-03
-Updated: 2023-02-24
+Updated: 2023-04-05
 Required: 0022, 0023, 0024, 0040, 0044, 0046, 0058, 0059
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0057.md
```diff
@@ -727,12 +727,13 @@ stakeTransactionParams = {
...
 * `trsParams.stakes` has at most `2 * MAX_NUMBER_STAKING_SLOTS` elements. The reason of choosing this bound on the size is to allow a staker to unstake all current validators (which are at most `MAX_NUMBER_STAKING_SLOTS`) and stake with `MAX_NUMBER_STAKING_SLOTS` new ones in the same transaction.
 * A given `validatorAddress` is included in at most one stake from the list of stakes (regardless of the associated amounts).
 * For all stakes included in `trsParams.stakes`, we have:
   * `amount` value is a multiple of `BASE_STAKING_AMOUNT`, i.e., `amount % BASE_STAKING_AMOUNT == 0`. For the Lisk mainchain, where `BASE_STAKING_AMOUNT = 10^9` and `TOKEN_ID_POS` is the token ID of the LSK token, this corresponds to multiples of 10 LSK.
   * `amount != 0`.
+* Let `senderAddress` be the address of the user sending the transaction (i.e., the staker). Let `totalUpstake` be the sum of all positive amounts of the `trsParams.stakes` array, i.e., the total amount that the sender is staking in this transaction. Then the following should hold: `Token.getAvailableBalance(senderAddress, TOKEN_ID_POS) >= totalUpstake`.
```
ricott1, line 736: I think you should still subtract the negative balance, as those are then again available to stake. In that case tho you need extra check to be sure that the unstake amount is correct. Also, why don't we use pseudocode here?

bobanm: I agree with @ricott1 here.

@gkoumout what you wrote in our discussion 4 days ago seems to be more correct:

```typescript
availableBalance + totalUnstakedAmount >= totalStakedAmount
```

Or, using the semantics from this document:

```typescript
Token.getAvailableBalance(senderAddress, TOKEN_ID_POS) >= totalUpstake - totalDownstake
```

The system should allow a staker to move his complete stake to another validator in 1 transaction, even when staker's balance is lower than the amount already staked, right?

bobanm (changes requested):

proposals/lip-0057.md
gkoumout: I also thought like that in the beginning, but this is actually not true for the current protocol: Once unstake, the amount remains locked for the locking period and will be unlocked later using the unlock command. Therefore, just after un-staking, the amount is not available; so I guess the check is correct.

gkoumout: @ricott1 I did not write pseudocode to have it faster, I can refactor this later.

bobanm (approved):

bobanm (approved):

janhack (approved):
Editorial approval

PR #318, opened 2023-04-03
Update terminology of sapphire phase lips

AndreasKendziorra:
Updates and aligns the terminology in the Sapphire phase LIPs. More precisely, the following terminology is used:

- "validator node" (instead of "generator node" or anything similar)
- "blockchain app" (instead of "blockchain application")
- "cross-chain certification" (instead of just "certification")
- "secrete recovery phrase" (instead of "passphrase")

Moreover, the term "sidechain validator" is avoided.


vjaiman (approved):

proposals/lip-0053.md
```diff
@@ -591,7 +591,7 @@ def execute(trs: Transaction) -> None:
     # Update the context to indicate that now we stop the CCM processing.
     ctx.ccmProcessing = False
 
-    # Update sidechain validators.
```
janhack, line 594: I would keep this comment as is because it is more precise. Even if we don't have `sidechain validator` as a term in the sheet, we do have the terms `sidechain` and `validator` separately so I think it is fine to create compound nouns to be more specific (it is just a synonym for validators of a sidechain).

proposals/lip-0053.md
```diff
@@ -742,7 +742,7 @@ def execute(trs: Transaction) -> None:
...
```
janhack, line 745: Same as above.

proposals/lip-0061.md
```diff
@@ -85,10 +85,10 @@ In general, for a height `h`, the certificate threshold could be chosen within t
 floor(1/3*aggregateBFTWeight(h))+1 <= certificateThreshold(h) <= aggregateBFTWeight(h),
 ```
 
-where `aggregateBFTWeight(h)` is the sum of BFT weights at height `h` and `certificateThreshold(h)` is the certificate threshold at height `h`. This allows for a sidechain to choose different trade-offs between satisfying certification liveness (generating certificates with enough signatures) and state transition validity (certificates attest a valid sidechain history):
+where `aggregateBFTWeight(h)` is the sum of BFT weights at height `h` and `certificateThreshold(h)` is the certificate threshold at height `h`. This allows for a sidechain to choose different trade-offs between satisfying cross-chain certification liveness (generating certificates with enough signatures) and state transition validity (certificates attest a valid sidechain history):
```
janhack, line 88: I would say that `cross-chain` does not always need to be added if the context is clear. For instance, I think it is fine to just say `message` if it is clear that it is a cross-chain message. This helps with readability without losing any precision as the context is clear.

Similarly, I would keep sidechain validators as "validators of the sidechain" is just a less readable synonym.

janhack (changes requested):
As suggested in the comments, I would be a bit less strict in a few cases in favor of readability.

proposals/lip-0053.md
AndreasKendziorra: reverted everywhere

proposals/lip-0053.md
AndreasKendziorra: changed. Note that "sidechain" was wrong. It must be "mainchain" in this line.

proposals/lip-0061.md
AndreasKendziorra: I reverted the changes related to "cross-chain".

janhack (approved):

PR #319, opened 2023-04-03
update LIP 70 - add expected rewards endpoint

gkoumout:
The purpose of this PR is to add an endpoint in the PoS module that calculates the expected rewards for users if they stake with a particular validator. This PR was created after request of the Platform team, to provide better user experience in user-facing products such as Lisk Wallet. 

In particular, the following endpoints are introduced:

 -  In the PoS module (reward sharing - LIP 70):  `getExpectedReward(validatorAddress, validatorReward, stake)`, which calculates the expected reward in case a user stakes `stake` amount with validator with address `validatorAddress`, assuming the reward of this validator is `validatorReward`.
 - In the Dynamic Rewards Module (LIP 71): `getExpectedValidatorBlockReward(validatorAddress)`, which estimates the expected reward for the specified validator, assuming the current stake/BFT distribution for different periods of time: for a block, a day, a month, a year.

Those endpoints are expected be used in front-end services as follows: 

0. Assume that the user gives as input the amount they are willing to stake and the address of the validator.
1. The endpoint `getExpectedValidatorBlockReward(validatorAddress)` will be called, to get the various expected rewards (block, daily, monthly, yearly)
2. Then, the endpoint `getExpectedReward(validatorAddress, validatorReward, stake)` is called, with the values of `validatorReward` being one of the outputs of step 1,  depending on which period is of interest (this is a front-end choice).

### DIFF
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -146,7 +146,7 @@ We also define `ROUND_LENGTH` to be the length of a round, i.e.,  `ROUND_LENGTH 
 | `STAKE_FAILED_TOO_MANY_PENDING_UNLOCKS` | uint32 | 3 | Used when a stake transaction fails because it the total number of pending unlocks of staker exceeds `MAX_NUMBER_PENDING_UNLOCKS`. |
 | `STAKE_FAILED_TOO_MANY_VALIDATORS` | uint32 | 4 | Used when a stake transaction fails because the total number of validators that the user has staked with exceeds `MAX_NUMBER_STAKING_SLOTS`. |
 
-### Type Definition
+### Type Definitions
 
 | Name | Type | Validation | Description |
 |------|------|------------|-------------|
@@ -1469,10 +1469,10 @@ Here `x` is the BFT weight at which the elements `validatorList[0].bftWeight, ..
 
 More functions might be made available during implementation.
 
-#### getActiveValidatorsNumber
+#### getNumberActiveValidators
 
 ```python
-def getActiveValidatorsNumber()-> uint32:
+def getNumberActiveValidators()-> uint32:
     return NUMBER_ACTIVE_VALIDATORS
 ```
 
@@ -1533,10 +1533,10 @@ def isNameAvailable(name: str) -> bool:
         return True
 ```
 
-#### getPoSTokenID
+#### getPosTokenID
 
 ```python
-def getPoSTokenID() -> TokenID:
+def getPosTokenID() -> TokenID:
     return TOKEN_ID_POS
 ```
 
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -1469,6 +1469,20 @@ Here `x` is the BFT weight at which the elements `validatorList[0].bftWeight, ..
 
 More functions might be made available during implementation.
 
+#### getActiveValidatorsNumber
+
+```python
+def getActiveValidatorsNumber()-> uint32:
+    return NUMBER_ACTIVE_VALIDATORS
+```
+
+#### getRoundLength
+
+```python
+def getRoundLength()-> uint32:
+    return ROUND_LENGTH
+```
+
 #### getStaker
 
 Returns the stored information relative to the given address.
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -9,7 +9,7 @@ Discussions-To: https://research.lisk.com/t/define-state-and-state-transitions-o
 Status: Draft
 Type: Standards Track
 Created: 2021-09-03
-Updated: 2023-02-24
+Updated: 2023-04-05
 Required: 0022, 0023, 0024, 0040, 0044, 0046, 0058, 0059
 ```
 
@@ -727,7 +727,7 @@ stakeTransactionParams = {
 
 Let `trsParams = decode(stakeTransactionParams, trs.params)` for the given transaction `trs`. Then the following verification steps are performed:
 
-* The function `validateObjectSchema(stakeTransactionParams, trsParams)` muss pass without exception.
+* The function `validateObjectSchema(stakeTransactionParams, trsParams)` must pass without exception.
 * `trsParams.stakes` has at most `2 * MAX_NUMBER_STAKING_SLOTS` elements. The reason of choosing this bound on the size is to allow a staker to unstake all current validators (which are at most `MAX_NUMBER_STAKING_SLOTS`) and stake with `MAX_NUMBER_STAKING_SLOTS` new ones in the same transaction.
 * A given `validatorAddress` is included in at most one stake from the list of stakes (regardless of the associated amounts).
 * For all stakes included in `trsParams.stakes`, we have:
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -9,7 +9,7 @@ Discussions-To: https://research.lisk.com/t/define-state-and-state-transitions-o
 Status: Draft
 Type: Standards Track
 Created: 2021-09-03
-Updated: 2023-04-05
+Updated: 2023-04-13
 Required: 0022, 0023, 0024, 0040, 0044, 0046, 0058, 0059
 ```
 
--- a/proposals/lip-0057.md
+++ b/proposals/lip-0057.md
@@ -1533,10 +1533,10 @@ def isNameAvailable(name: str) -> bool:
         return True
 ```
 
-#### getPosTokenID
+#### getPoSTokenID
 
 ```python
-def getPosTokenID() -> TokenID:
+def getPoSTokenID() -> TokenID:
     return TOKEN_ID_POS
 ```
 
--- a/proposals/lip-0070.md
+++ b/proposals/lip-0070.md
@@ -434,6 +434,31 @@ def getClaimableRewards(stakerAddress: Address) -> dict[TokenID, uint64]:
     return rewards
 ```
 
+#### getExpectedReward
+
+This function calculates the expected reward for staking with a particular validator. 
+
+##### Parameters
+
+* `validatorAddress`: The address of the validator considering to stake with.
+* `validatorReward`: The hypothetical reward for the validator. 
+* `stakeAmount`: The amount the user is considering to stake with the specified validator.
+
+
+```python
+def getClaimableRewards(validatorAddress: Address, validatorReward: uint64, stakeAmount: uint64) -> uint64:
+
+     if validatorStore(validatorAddress).totalStake == 0: # If sharing coefficient can not be defined, we return.
+        return
+    # Use Q96 numbers.
+    validatorReward = Q96(validatorReward)
+    rewardFraction = sub_96(Q96(1), div_96(Q96(validatorStore(validatorAddress).commission), Q96(10000)))
+    totalStake = Q96(validatorStore(validatorAddress).totalStake)
+
+    rewardPerUnitStaked = muldiv_96(validatorReward, rewardFraction, totalStake) 
+    return mul_96(rewardPerUnitStaked, stakeAmount) 
+```
+
 ## Rationale
 
 ### Commission
--- a/proposals/lip-0070.md
+++ b/proposals/lip-0070.md
@@ -444,16 +444,13 @@ This function calculates the expected reward for staking with a particular valid
 * `validatorReward`: The hypothetical reward for the validator. 
 * `stakeAmount`: The amount the user is considering to stake with the specified validator.
 
-
 ```python
-def getClaimableRewards(validatorAddress: Address, validatorReward: uint64, stakeAmount: uint64) -> uint64:
+def getExpectedReward(validatorAddress: Address, validatorReward: uint64, stakeAmount: uint64) -> uint64:
 
-     if validatorStore(validatorAddress).totalStake == 0: # If sharing coefficient can not be defined, we return.
-        return
     # Use Q96 numbers.
     validatorReward = Q96(validatorReward)
     rewardFraction = sub_96(Q96(1), div_96(Q96(validatorStore(validatorAddress).commission), Q96(10000)))
-    totalStake = Q96(validatorStore(validatorAddress).totalStake)
+    totalStake = Q96(validatorStore(validatorAddress).totalStake + stakeAmount)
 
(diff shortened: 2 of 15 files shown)


proposals/lip-0070.md
```diff
@@ -434,6 +434,28 @@ def getClaimableRewards(stakerAddress: Address) -> dict[TokenID, uint64]:
...
 ```
 
+#### getExpectedReward
+
+This function calculates the expected reward for staking with a particular validator. 
+
+##### Parameters
+
+* `validatorAddress`: The address of the validator considering to stake with.
+* `validatorReward`: The hypothetical reward for the validator.
```
janhack, line 444: How is this reward supposed to be obtained? LIP 71 does not provide an endpoint at the moment.
Ideally, something like the function [getValidatorBlockReward](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0071.md#getvalidatorblockreward) would also be accessible. Then the front-end could easily compute the expected ROI or expected shared rewards per month/year.

janhack: Maybe you could add an appendix then how to compute the ROI or rewards per month/year with these endpoints (in particular, based on the round length and block time).

janhack (commented):
The endpoint is correct as far as I can see, but I am wondering who exactly it would be used and, in particular, how `validatorReward` would be obtained.

proposals/lip-0070.md
```diff
@@ -434,6 +434,28 @@ def getClaimableRewards(stakerAddress: Address) -> dict[TokenID, uint64]:
...
 
+#### getExpectedReward
+
+This function calculates the expected reward for staking with a particular validator. 
+
+##### Parameters
+
+* `validatorAddress`: The address of the validator considering to stake with.
+* `validatorReward`: The hypothetical reward for the validator. 
+* `stakeAmount`: The amount the user is considering to stake with the specified validator.
```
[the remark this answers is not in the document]
gkoumout: switched to `stake`

proposals/lip-0070.md
```diff
@@ -434,6 +434,28 @@ def getClaimableRewards(stakerAddress: Address) -> dict[TokenID, uint64]:
...
+
+```python
+def getExpectedReward(validatorAddress: Address, validatorReward: uint64, stakeAmount: uint64) -> uint64:
+
+    # Use Q96 numbers.
+    validatorReward = Q96(validatorReward)
+    rewardFraction = sub_96(Q96(1), div_96(Q96(validatorStore(validatorAddress).commission), Q96(10000)))
+    totalStake = Q96(validatorStore(validatorAddress).totalStake + stakeAmount)
+
+    rewardPerUnitStaked = muldiv_96(validatorReward, rewardFraction, totalStake)
```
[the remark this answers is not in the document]
gkoumout: No problem, but I chose this way to be consistent with the way of calculating actual rewards, i.e., multiplying sharing coefficient by amount. The `rewardPerUnitStaked` is kind of a fake sharing coefficient here.

proposals/lip-0070.md
gkoumout: I assume that LIP 71 provides such an endpoint and the UI would have access to both endpoints (PoS and Dynamic Rewards).

gkoumout: See also changes in LIP 71

proposals/lip-0057.md
```diff
@@ -1469,6 +1469,20 @@ Here `x` is the BFT weight at which the elements `validatorList[0].bftWeight, ..
 
 More functions might be made available during implementation.
 
+#### getActiveValidatorsNumber
```
janhack, line 1472: The word order is a bit odd, I would call it `getNumberOfActiveValidators`.

proposals/lip-0070.md
```diff
@@ -434,6 +434,28 @@ def getClaimableRewards(stakerAddress: Address) -> dict[TokenID, uint64]:
     return rewards
 ```
 
+#### getExpectedReward
```
janhack, line 437: I think the name of this function `getExpectedReward` and the function `getValidatorBlockReward` is not sufficiently clear to distinguish them. My suggestion would be:

- `getExpectedSharedBlockReward` (function here)
- `getExpectedBlockReward` (function in LIP 71)

janhack: I am also not sure whether reward or rewards is better, maybe we could ask a native speaker.

proposals/lip-0070.md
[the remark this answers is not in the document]
janhack: Okay, makes sense. I resolve it.

proposals/lip-0070.md
janhack: Or, if we also expect the endpoint to be used for shared fees in the future, we could call it `getExpectedSharedRewards`.

proposals/lip-0071.md
```diff
@@ -86,8 +89,8 @@ We calculate the rewards for validators as follows:
 | Name | Mainchain Value | Computation | Description |
 |------|-----------------|-------------|-------------|
 | `minimumRewardActiveValidators` | 10<sup>7</sup> | (`FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS * getDefaultRewardAtHeight(blockHeader.height))//DECIMAL_PERCENT_FACTOR` | The minimum reward per round for an active validator. Here `DECIMAL_PERCENT_FACTOR == 10000` and `blockHeader.height` is the current block height. |
-| `totalRewardActiveValidators` | 101*10<sup>8</sup> | `NUMBER_ACTIVE_VALIDATORS * getDefaultRewardAtHeight(blockHeader.height)` | The total reward per round for active validators. Here `blockHeader.height` is the current block height. |
-| `stakeRewardActiveValidators`| 90.9*10<sup>8</sup> | `totalRewardActiveValidators - NUMBER_ACTIVE_VALIDATORS * minimumRewardActiveValidators` | The remaining rewards for active validators (per round) after giving the minimum reward. |
+| `totalRewardActiveValidators` | 101*10<sup>8</sup> | `Pos.getActiveValidatorsNumber * getDefaultRewardAtHeight(blockHeader.height)` | The total reward per round for active validators. Here `blockHeader.height` is the current block height. |
```
janhack, line 92: We don't use a consistent spelling for the method calls to the PoS module. I would either use `PoS` or `pos` consistently in this LIP.

janhack (changes requested):
Looks good. I just have some cosmetic naming or consistency remarks.

proposals/lip-0071.md
```diff
@@ -269,7 +272,36 @@ The default reward of the block as a 64-bit unsigned integer.
...
+This function is used to estimate the expected rewards for a validator assuming the current BFT weight distribution. It provides an estimate of rewards for one block, one day, one month and one year.
+
+##### Returns
+
+* `reward:` amount of block reward for a validator.
+
+##### Execution
+
+```python
+def getExpectedValidatorBlockReward(validatorAddress: Address) -> uint64:
```
vjaiman, line 286:
```suggestion
def getExpectedValidatorBlockReward(validatorAddress: Address) -> tuple[uint64, uint64, uint64, uint64]:
```

proposals/lip-0057.md
```diff
@@ -1469,6 +1469,20 @@ Here `x` is the BFT weight at which the elements `validatorList[0].bftWeight, ..
...
+    return NUMBER_ACTIVE_VALIDATORS
+```
+
+#### getRoundLength
+
+```python
+def getRoundLength()-> uint32:
+    return ROUND_LENGTH
+```
+
```
vjaiman, line 1485: Since we are adding these two, do you think we should add `getStandByValidatorsNumber()` also.

vjaiman (changes requested):

proposals/lip-0057.md
gkoumout: done.

proposals/lip-0070.md
gkoumout: switched to:

- `getExpectedStakingBlockReward` in LIP 70: this way it is clear that  i) it is expected, not guaranteed ii) it has to do with staking and iii) it is only for one block.
- `getExpectedValidatorRewards` (as it was) in LIP 71: This way it is clear that i) it is expected, ii) it is for validators; and I use "rewards" instead of "reward" since it gives potential rewards for several different time intervals.

gkoumout: As for fees: I don't see how we could estimate expected transition fees and did not bother much about it now. Open to discuss this if you want. Then, a suitable name would be `getExpectedStakingRewardsPerBlock` I would say.

proposals/lip-0071.md
gkoumout: Ok I guess this is because I edited a LIP which I did not write. I think the convention in LIPs is to capitalize the first letter of module name in pseudocode, so I would use `Pos.xxx`. 

Plz check it is now consistent.

proposals/lip-0057.md
gkoumout: Since we don't seem to need this for now, I would skip it.

proposals/lip-0071.md
```diff
@@ -269,7 +272,40 @@ The default reward of the block as a 64-bit unsigned integer.
...
+This function is used to estimate the expected rewards for a validator assuming the current BFT weight distribution. It provides an estimate of rewards for one block, one day, one month and one year.
+
+##### Returns
+
+* `reward:` amount of block reward for a validator.
+
+##### Execution
+
+```python
+<<<<<<< HEAD
```
gkoumout, line 286: Here I had some conflict due to applying a suggestion from github and forgetting to pull it, should be fine now.

proposals/lip-0071.md
```diff
@@ -236,7 +239,7 @@ This function is used to retrieve the block reward for a validator.
 
 ```python
 def getValidatorBlockReward(blockHeader: BlockHeader) -> uint64:
-    if len(Validators.getGeneratorsBetweenTimestamps(endOfRoundTimestamp, blockHeader.timestamp)) >= ROUND_LENGTH:
+    if len(Validators.getGeneratorsBetweenTimestamps(endOfRoundTimestamp, blockHeader.timestamp)) >= pos.getRoundLength():
```
vjaiman, line 242:
```suggestion
    if len(Validators.getGeneratorsBetweenTimestamps(endOfRoundTimestamp, blockHeader.timestamp)) >= Pos.getRoundLength():
```

vjaiman (approved):

proposals/lip-0071.md
janhack: Looks good.

proposals/lip-0071.md
```diff
@@ -269,7 +272,36 @@ The default reward of the block as a 64-bit unsigned integer.
...
 ### Endpoints for Off-Chain Services
 
-TBD.
+#### getExpectedValidatorRewards
+
+This function is used to estimate the expected rewards for a validator assuming the current BFT weight distribution. It provides an estimate of rewards for one block, one day, one month and one year.
+
+##### Returns
+
+* `reward:` amount of block reward for a validator.
```
janhack, line 281: This section is outdated.

proposals/lip-0071.md
```diff
@@ -269,7 +272,36 @@ The default reward of the block as a 64-bit unsigned integer.
...
+        return (0, 0, 0, 0) 
+
+    if validator.bftWeight > 0:
+        blockReward = minimumRewardActiveValidators + (validator.bftWeight * stakeRewardActiveValidators)//totalBftWeight 
+    else:
+        blockReward = defaultRewardStandByValidators
+
+    dailyReward =   86400 // (Pos.getRoundLength * BLOCK_TIME)
+    monthlyReward = 2592000 // (Pos.getRoundLength * BLOCK_TIME) 
+    yearlyReward = 31536000 // (Pos.getRoundLength * BLOCK_TIME)
```
janhack, line 301: These numbers are just how often a block is generated. Don't we need to still multiple with the block reward?

proposals/lip-0070.md
janhack: The name in LIP 71 sounds good.

Regarding the name in LIP 71, it was not changed in the files. Further, I would find "shared" as part of the name nicer than staking, as it makes it more clear that this is what is shared after commission. Further, I would remove "block" from the name, as it is not necessarily just for one block, it just totally depends on the input parameter `validatorRewards` of the function. For instance, you could use as input parameter the yearly rewards for a validator or, in case in a sidechain fees are also shared with stakers, the yearly collected fees by a validator. Basically, this function allows to compute for any shared amount coming from any source, how much a single staker receives.

gkoumout: For LIP70 function name:

- The file is now updated.
 - I got confused, or course Block does not make sense.
 - I think "staking rewards" is a standard term used. But ok switched to shared since I have not a strong opinion here.

proposals/lip-0071.md
gkoumout: right, probably got removed during pasting constants from calculator.. Nice that you caught that!

proposals/lip-0070.md
janhack: Looks good.

proposals/lip-0057.md
```diff
@@ -1519,10 +1533,10 @@ def isNameAvailable(name: str) -> bool:
         return True
 ```
 
-#### getPoSTokenID
+#### getPosTokenID
 
 ```python
-def getPoSTokenID() -> TokenID:
+def getPosTokenID() -> TokenID:
```
ManuGowda, line 1539: :( this will be recurring change everywhere? is this important?

ManuGowda (approved):
Endpoint LGTM

proposals/lip-0057.md
shuse2: also, `PoS` seems correct?

shuse2 (approved):
Created issue in SDK https://github.com/LiskHQ/lisk-sdk/issues/8355

proposals/lip-0057.md
vjaiman: @ManuGowda No, this is not that important.

vjaiman: @janhack Do we need to revert the it? or we can keep it like this in the LIP?

janhack: As discussed, I reverted the name of the endpoint to the previous version.

Issue #324, opened 2023-04-15
LIP0053: Unused variable [defaultEventsTopic] can be removed or be part of comment


sitetester:
https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#execution
> Set ccmID (instead of the transaction ID) as default topic of this event.
defaultEventsTopic = ccmID

I see 2 issues here:
- As per comment, "default topic" refers to default topic (of event), while variable (defaultEventsTopic) refers to default event (& not default topic). Hence correct naming is "eventsDefaultTopic".
- There is no further usage of this variable. If we search for "defaultEventsTopic", we only find "defaultEventsTopic = ccmID" (4 times in LIP).

If these variables are declared based on assumption that they will be used at some point (without actually being used in code/lip), then we should mention them only as comment, otherwise, it creates confusion that perhaps some code segment is missing which actually use this variable. Hence I suggest;
- Either remove all 4 assignments of `defaultEventsTopic = ccmID` from page
- Mention them as part of the given comment on same line. In given case, it will look like `Set ccmID (instead of the transaction ID) as default topic of this event, i.e., eventsDefaultTopic = ccmID` (preferable)

Also here - https://github.com/LiskHQ/lips/blob/main/proposals/lip-0054.md#execution-1

ricott1:
The default topic is used in `emitEvent` (in the PR I add a comment explaining this). Also notice that the LIPs present specs in pseudocode format, but it should not be expected that the code is runnable as is.

Issue #309, opened 2023-03-22
LIP0053: fix typo in pseudocode


ricott1:
Fix typo in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#computing-the-validators-update

```python
bftWeightsUpdate = [v.bftWeight for validator in validatorsUpdate]
```

should be

```python
bftWeightsUpdate = [validator.bftWeight for validator in validatorsUpdate]
```

PR #328, opened 2023-04-18
LIP0053: fix typo and improve default events topic comments

ricott1:
Closes https://github.com/LiskHQ/lips/issues/324. Closes https://github.com/LiskHQ/lips/issues/309.

### DIFF
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -522,7 +522,8 @@ def execute(trs: Transaction) -> None:
     for ccmBytes in crossChainMessages:
         ccmID = sha256(ccmBytes)
         # Set ccmID (instead of the transaction ID) as default topic of this event.
-        defaultEventsTopic = ccmID
+        # The default topic is added automatically by emitEvent.
+        eventsDefaultTopic = ccmID
         # Verify general format. Past this point, we can access ccm root properties.
         try:
             ccm = decode(crossChainMessageSchema, ccmBytes)
@@ -574,7 +575,8 @@ def execute(trs: Transaction) -> None:
     for ccmBytes in crossChainMessages:
         ccmID = sha256(ccmBytes)
         # Set ccmID (instead of the transaction ID) as default topic to all events emitted in apply and forward.
-        defaultEventsTopic = ccmID
+        # The default topic is added automatically by emitEvent.
+        eventsDefaultTopic = ccmID
         ccm = decode(crossChainMessageSchema, ccmBytes)
         # If the receiving chain is the mainchain, apply the CCM.
         # This function never raises an error.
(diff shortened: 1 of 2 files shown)


proposals/lip-0053.md
```diff
@@ -522,7 +522,8 @@ def execute(trs: Transaction) -> None:
     for ccmBytes in crossChainMessages:
         ccmID = sha256(ccmBytes)
         # Set ccmID (instead of the transaction ID) as default topic of this event.
-        defaultEventsTopic = ccmID
+        # The default topic is added automatically by emitEvent.
+        eventsDefaultTopic = ccmID
```
AndreasKendziorra, line 526: Now I'm just wondering what's the scope of this change? I mean, when will the transaction ID be the default topic again? After the next event? After the for loop? After the function? Never? Maybe this should be clarified in the comment as well. Or we add an explicit `eventsDefaultTopic = transactionID` in the right place.
Depending on the answer, we may also be able to drop the 2nd `eventsDefaultTopic = ccmID` in this function.

proposals/lip-0053.md
```diff
@@ -732,7 +735,8 @@ def execute(trs: Transaction) -> None:
     for ccmBytes in crossChainMessages:
         ccmID = sha256(ccmBytes)
         # Set ccmID (instead of the transaction ID) as default topic to all events emitted in apply.
-        defaultEventsTopic = ccmID
+        # The default topic is added automatically by emitEvent.
+        eventsDefaultTopic = ccmID
```
AndreasKendziorra, line 739: Same as above: could potentially removed.

AndreasKendziorra (changes requested):

proposals/lip-0053.md
ricott1: Notice that ccmID changes during the loop, so we can't remove this

proposals/lip-0053.md
ricott1: I explicitly added the transaction ID back as default event

proposals/lip-0053.md
```diff
@@ -593,6 +593,10 @@ def execute(trs: Transaction) -> None:
     # Update the context to indicate that now we stop the CCM processing.
     ctx.ccmProcessing = False
 
+    # Set the transaction ID back as default topic.
+    # The default topic is added automatically by emitEvent.
+    eventsDefaultTopic = transaction ID
```
AndreasKendziorra, line 598: 1. I would do this for LIP 0054 as well.
2. Maybe it's a bit cleaner to have than only a single `eventsDefaultTopic = ccmID` statement before the first for loop. But it's also ok to have like this if you prefer.

AndreasKendziorra (changes requested):

proposals/lip-0053.md
AndreasKendziorra: ah, forget the 2nd point. I just read your other comment.

proposals/lip-0053.md
AndreasKendziorra: right, I missed this

proposals/lip-0053.md
ricott1: In LIP54 there are no other events emitted in the transaction execution, that's why I didn't add it

AndreasKendziorra: ok, I see.

AndreasKendziorra (approved):

janhack (approved):
Editorial approval

PR #334, opened 2023-04-28
Fix certificate generation mainchain

gkoumout:
## Short Description

This PR addresses an issue in the certificate generation for mainchain for Lisk Core 4 migration. In short, due to the fact that current validators do not have BLS keys, and they should register after migration, initially the mainchain will not be able to produce certificates. This happens since the specs for certificate generation implicitly assume that all validators have registered BLS keys from genesis height (which is the case for all sidechains).

## Technical explanation of the issue

Mainchain is not able to produce certificates, as a combination of 2 facts:

1. All single commits are verified using the public BLS key `pk` stored in the BFT parameters (see [item 6 in the single commit validation algorithm](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#single-commit-p2p-gossip)). Initially this is empty for all initValidators after migration. Even if the validators register a BLS key, this change will be available in the engine only after BFT parameters are updated. Therefore, for the initial rounds no single commit will be accepted.
2. At each block, the generator attempts to include an aggregateCommit, using [`selectAggregateCommit` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#selectaggregatecommit). This function picks the height interval in which the "certified block to be" is, which is `(maxCertifiedHeight, nextHeight]`. Clearly, in case `nextHeight` is smaller than the min height for which single commits are succesfully created( based on constraint 1), then no certificate can be added in this block. By definition of `nextHeight`, the value will never become larger than the last block of init rounds. 

## Suggested solution

We introduce a configurable constant `MIN_CERTIFICATE_HEIGHT`. For blockchains created using Lisk SDK its value would be equal to the genesis height. However, for the mainchain for next Mainet migration, the value would be set to the first block of the first round after initial and hybrid phase, i.e., after the effect of initial validators finishes. From this height ownwards, we can guarantee that whenever there is a change in BFT parameters, then all validators would have valid BLS keys registered in the engine (otherwise they could not become active according to our PoS). 

Then in [`selectAggregateCommit` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#selectaggregatecommit) we set the expected height to check for new certificates as

`nextHeight = max(nextHeight, MIN_CERTIFICATE_HEIGHT)`

This way, `nextHeight` will never be during init rounds and the first certificate to be created will be at height `MIN_CERTIFICATE_HEIGHT`.


proposals/lip-0061.md
```diff
@@ -520,10 +523,10 @@ def selectAggregateCommit() -> AggregateCommit:
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated by the previous certificate.
     heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified + 1)
     if previous function call returns valid height:
-        nextHeight = min(heightNextBFTParameters-1, getBFTHeights().maxHeightPrecommitted)
+        nextHeight = min(max(heightNextBFTParameters -1, MIN_CERTIFICATE_HEIGHT), getBFTHeights().maxHeightPrecommitted)
```
shuse2, line 526: `MIN_CERTIFICATE_HEIGHT` seems to be the height of the first block of the round, but `heightNextBFTParameters -1` is the last block of the round. isn't it bit inconsistent?

proposals/lip-0061.md
```diff
@@ -520,10 +523,10 @@ def selectAggregateCommit() -> AggregateCommit:
...
     elif previous function call returns Not Found error:
         nextHeight = getBFTHeights().maxHeightPrecommitted
-
+
```
ricott1, line 529: remove extra space
```suggestion

```

proposals/lip-0061.md
janhack: I think what we want here is that in the while loop below, `nextHeight` always has a value of at least `MIN_CERTIFICATE_HEIGHT`, i.e., this is the minimum height for which we generate a certificate. I don't think the current logic guarantees that.

proposals/lip-0061.md
```diff
@@ -570,7 +573,9 @@ def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated
     # by the previous certificate.
     heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
-    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
+    if previous function call returns valid height
+     and heightNextBFTParameters > MIN_CERTIFICATE_HEIGHT
```
janhack, line 577: So with the additional condition we require that a certificate is at the earliest required for the last block of the round that starts with height `MIN_CERTIFICATE_HEIGHT`.

I think what we additionally should check is that we do not have any certificate below height `MIN_CERTIFICATE_HEIGHT` so we are guaranteed to have a chain of trust.

janhack (changes requested):
I am not sure that the current proposed changes are correct and left some comments in that regard.

proposals/lip-0061.md
ricott1, line 577: So basically here we amend to the rule state in the comment as:

`If there are new BFT parameters for a height h > MIN_CERTIFICATE_HEIGHT, then the chain needs to include an aggregate commit for the block at height h-1.`

proposals/lip-0061.md
```diff
@@ -570,7 +573,9 @@ def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated
     # by the previous certificate.
     heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
-    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
+    if previous function call returns valid height
```
ricott1, line 576: Now that we explicitly use the return value, wouldn't it be better to write 

```suggestion
    if previous function call returns valid heightNextBFTParameters
```

or maybe even better use a try statement in the previous function call (catching the exception to return False)

proposals/lip-0061.md
```diff
@@ -520,10 +523,10 @@ def selectAggregateCommit() -> AggregateCommit:
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated by the previous certificate.
     heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified + 1)
     if previous function call returns valid height:
```
ricott1, line 525: same comment as below

proposals/lip-0061.md
ricott1, line 526: question: below we basically use the `max(heightNextBFTParameters, MIN_CERTIFICATE_HEIGHT)`, here instead we have an extra -1, what is the reason?

gkoumout: @shuse2 It is not matter of consistency, but feasibility. The `MIN_CERTIFICATE_HEIGHT` has to be first block of a round, since this is the first block where bft params are guaranteed to exist. For `heightNextBFTParameters -1` the point is that we want to certify the block that decides the new validators and this is the last block of round by construction of PoS, so I think this can not be changed.

proposals/lip-0061.md
gkoumout: Good point, although it is just a cosmetic change, I think it makes it somewhat cleaner.

gkoumout: Done, using the try except.

proposals/lip-0061.md
gkoumout: Indeed, I had this check before and it was erroneously removed. Check now.

proposals/lip-0061.md
gkoumout: I don't think that by catching the exception you return False here. I added try - except for the call of this function in `selectAggregateCommit` since it indeed looks cleaner.

proposals/lip-0061.md
gkoumout: Exactly.

proposals/lip-0061.md
gkoumout: `nextHeight` should be the highest height for which we attempt to create an aggregate commit. What should be the value of `nextHeight`? `heightNextBFTParameters -1`, because if we have BFT params at a height `h`, we want the certificate at height `h-1`. But in case this value is smaller than `MIN_CERTIFY_HEIGHT`, we can not create such a certificate, so set it to `MIN_CERTIFY_HEIGHT`

gkoumout: @janhack in case `heightNextBFTParameters` is undefined, the value of `nextHeight` will not be greater than `MIN_CERTIFICATE_HEIGHT`. But somehow we don't care about this case; even if an aggregate commit could be created, it would be rejected.

On the other hand, we really need the max() in case `heightNextBFTParameters` defined, in order to be able to create valid aggregate commits.

gkoumout: I updated the usage of max operators (reverting to a previous version) since it seemed confusing for everyone.

proposals/lip-0061.md
```diff
@@ -517,14 +520,18 @@ An object following the schema `aggregateCommitSchema`, which can be added the n
...
+        nextHeight = min(heightNextBFTParameters -1, getBFTHeights().maxHeightPrecommitted)
+    except:  
+        # getNextHeightBFTParameters function call returns error, i.e., there are no new bft params since last certificate.
+        # attempt to create certificates for finalized blocks
         nextHeight = getBFTHeights().maxHeightPrecommitted
 
-    while nextHeight > getBFTHeights().maxHeightCertified:
+    nextHeight = max(nextHeight, MIN_CERTIFICATE_HEIGHT)    
+
+    while nextHeight > max(getBFTHeights().maxHeightCertified, MIN_CERTIFICATE_HEIGHT - 1):
```
shuse2, line 534: `MIN_CERTIFICATE_HEIGHT` by default should be `1` in this case?

proposals/lip-0061.md
```diff
@@ -564,14 +571,21 @@ def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
...
     # an aggregate commit for the block at height h-1. This block and the corresponding
     # certificate authenticate the BFT parameters via the validators hash property.
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated
     # by the previous certificate.
-    heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
-    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
-        return False
+    try:
+        heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
+        if heightNextBFTParameters > MIN_CERTIFICATE_HEIGHT
```
ricott1, line 584: I am a bit unsure about this extra condition. I would have said that we should check `aggregateCommit.height > MIN_CERTIFICATE_HEIGHT` (which you did above). Here we are saying that the aggregate commit can be valid if, for instance, `aggregateCommit.height == MIN_CERTIFICATE_HEIGHT >= heightNextBFTParameters`.

gkoumout: This try- except here has to do with chain of trust verification. So, if we ignore `MIN_CERTIFY_HEIGHT` (say it has value -1), what we do is to enforce chain of trust by saying "for every change of bft parameters at height `h`, need to include a certificate for height `h-1`".

When `MIN_CERTIFY_HEIGHT` comes into the game, we want to say "for every change of bft parameters at height `h > MIN_CERTIFICATE_HEIGHT`, need to include a certificate for height `h-1`" . The reason is that for changes of bft params at height `h <= MIN_CERTIFICATE_HEIGHT`, we can not have a certificate at height `h-1`..

Does it make sense?

gkoumout: In particular for the example that you give: Assume the first certificate is produced for height `h = MIN_CERTIFICATE_HEIGHT`, i.e., `aggregateCommit.height = MIN_CERTIFICATE_HEIGHT`. 

Then, I guess this certificate should be accepted and it would be the first one produced. After this one, chain of trust will be enforced.

proposals/lip-0061.md
gkoumout: Can't it be 0?

ricott1: I think it makes more sense if the default is `1`, but probably doesnt change much

shuse2 (approved):

proposals/lip-0061.md
janhack: It does not make sense to generate certificates for the genesis block so I would also set it to 1 by default.

proposals/lip-0061.md
```diff
@@ -564,14 +571,21 @@ def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
...
     # If there are new BFT parameters for a height h, then the chain needs to include
     # an aggregate commit for the block at height h-1. This block and the corresponding
     # certificate authenticate the BFT parameters via the validators hash property.
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated
     # by the previous certificate.
-    heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
-    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
-        return False
+    try:
+        heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
```
janhack, line 583: I think it would be better to do
```
heightNextBFTParameters = max(MIN_CERTIFICATE_HEIGHT + 1, getNextHeightBFTParameters(getBFTHeights().maxHeightCertified + 1)
```
here and remove the additional condition `heightNextBFTParameters > MIN_CERTIFICATE_HEIGHT` below. This implies that we enforce that a certificate for height `MIN_CERTIFICATE_HEIGHT` is generated. Note that with the current specs the height of the first certificate can be arbitrarily large for the Lisk Mainchain (because `heightNextBFTParameters > MIN_CERTIFICATE_HEIGHT` will always be `False` as long as no certificate is generated).

proposals/lip-0063.md
```diff
@@ -279,6 +279,7 @@ Some new LIPs introduced configurable constants. The values for these constants
 | [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants) | `MAX_ASSET_DATA_SIZE_BYTES` | 18 (The Random Module is the only module adding an entry to the assets property whereby the size of the data property will not exceed 18 bytes) |
 | [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#notation-and-constants), [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants) | `OWN_CHAIN_ID` | `CHAIN_ID_MAINCHAIN` |
 | [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#constants) | `BLOCK_TIME` | 10 seconds |
+| [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#constants) | `MIN_CERTIFICATE_HEIGHT` | `(POS_INIT_ROUNDS + NUMBER_ACTIVE_VALIDATORS - 1) * ROUND_LENGTH +1 =  6239741` |
```
janhack, line 282: I think we still need to add `HEIGHT_SNAPSHOT + 1`.

janhack (changes requested):

proposals/lip-0061.md
gkoumout: Ok switched the definition to genesis height + 1.

proposals/lip-0063.md
gkoumout: Good catch. Fixed.

proposals/lip-0061.md
```diff
@@ -517,14 +520,18 @@ An object following the schema `aggregateCommitSchema`, which can be added the n
...
+        # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated by the previous certificate.
+        heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified + 1)
+        nextHeight = min(heightNextBFTParameters -1, getBFTHeights().maxHeightPrecommitted)
+    except:  
+        # getNextHeightBFTParameters function call returns error, i.e., there are no new bft params since last certificate.
+        # attempt to create certificates for finalized blocks
         nextHeight = getBFTHeights().maxHeightPrecommitted
 
-    while nextHeight > getBFTHeights().maxHeightCertified:
+    nextHeight = max(nextHeight, MIN_CERTIFICATE_HEIGHT)
```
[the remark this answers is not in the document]
gkoumout: If I understand well, what you propose is equivalent to

```
heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
nextHeight = min(max(heightNextBFTParameters -1, MIN_CERTIFICATE_HEIGHT), getBFTHeights().maxHeightPrecommitted)

```

which was in a previous version (see [this commit](https://github.com/LiskHQ/lips/pull/334/commits/0b50ca6141d3fe9791a5f270cac79f790ef213be)). Given the comments I received I got confused and reverted it. So I guess you suggest that this assignment was correct because it does not allow to create aggregate commits for non-finalized blocks. Is that right?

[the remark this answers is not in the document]
gkoumout: Switched to the previous setting.

proposals/lip-0061.md
gkoumout: I switched to your suggestion, meaning we force a certificate for height `MIN_CERTIFICATE_HEIGHT`.

proposals/lip-0061.md
janhack: Can be resolved from my side.

proposals/lip-0061.md
```diff
@@ -517,14 +520,16 @@ An object following the schema `aggregateCommitSchema`, which can be added the n
...
 def selectAggregateCommit() -> AggregateCommit:
-    # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated by the previous certificate.
-    heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified + 1)
-    if previous function call returns valid height:
-        nextHeight = min(heightNextBFTParameters-1, getBFTHeights().maxHeightPrecommitted)
-    elif previous function call returns Not Found error:
+    try:
+        # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated by the previous certificate.
+        heightNextBFTParameters = max(getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1), MIN_CERTIFICATE_HEIGHT + 1)
+        nextHeight = min(heightNextBFTParameters -1, getBFTHeights().maxHeightPrecommitted)
```
ricott1, line 526:
```suggestion
        nextHeight = min(heightNextBFTParameters, getBFTHeights().maxHeightPrecommitted)
```

ricott1: If I'm not mistaken it is possible to remove some extra floating +- 1s

proposals/lip-0061.md
```diff
@@ -564,14 +569,20 @@ def verifyAggregateCommit(blockHeader: BlockHeader)-> bool:
...
     # If there are new BFT parameters for a height h, then the chain needs to include
     # an aggregate commit for the block at height h-1. This block and the corresponding
     # certificate authenticate the BFT parameters via the validators hash property.
     # Note that the BFT parameters at height maxHeightCertified+1 are already authenticated
     # by the previous certificate.
-    heightNextBFTParameters = getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1)
-    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
-        return False
+    try:
+        heightNextBFTParameters = max(getNextHeightBFTParameters(getBFTHeights().maxHeightCertified+1), MIN_CERTIFICATE_HEIGHT + 1)
```
ricott1, line 581: Same as above, it should be possible to simplify this

ricott1 (changes requested):
I left a couple suggestions to simplify the pseudocode

proposals/lip-0063.md
```diff
@@ -279,6 +279,7 @@ Some new LIPs introduced configurable constants. The values for these constants
 | [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants) | `MAX_ASSET_DATA_SIZE_BYTES` | 18 (The Random Module is the only module adding an entry to the assets property whereby the size of the data property will not exceed 18 bytes) |
 | [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#notation-and-constants), [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md#notation-and-constants) | `OWN_CHAIN_ID` | `CHAIN_ID_MAINCHAIN` |
 | [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#constants) | `BLOCK_TIME` | 10 seconds |
+| [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#constants) | `MIN_CERTIFICATE_HEIGHT` | `HEIGHT_SNAPSHOT + 1 + (POS_INIT_ROUNDS + NUMBER_ACTIVE_VALIDATORS - 1) * ROUND_LENGTH +1 = HEIGHT_SNAPSHOT + 1 +  6239741` |
```
janhack, line 282: Why not `HEIGHT_SNAPSHOT +  6239742` in simplified form?

proposals/lip-0061.md
janhack: That is not correct as `getBFTHeights().maxHeightCertified + 1` is input to a function.

janhack (approved):
Just one minor suggestion, otherwise the PR looks good.

proposals/lip-0061.md
janhack: I think this would introduce a mistake, see above.

proposals/lip-0061.md
ricott1: ah missed it!

ricott1: can be resolved from my side

proposals/lip-0063.md
gkoumout: I left it as is to be easier for the testing, since this +1 might make the setup more confusing. It is fine for me to switch it, but we should take extra care on testing.

janhack: Okay, makes sense.

shuse2 (approved):

shuse2 (approved):

PR #341, opened 2023-04-30
LIP0049: Replace hard coded constant incorrect value with it's numerical value

sitetester:
// https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md

CCM_STATUS_CODE_OK	uint32	0	Value of status of a new CCM which is not a response due do an error.

`CCM_STATUS_CODE_OK` has numerical value of 0, but in error message, `OK` value is used instead.


ricott1 (approved):

gkoumout (approved):

janhack (approved):
Editorial approval

PR #351, opened 2023-05-08
 Improve LIP0045 and LIP 0054 Message Recovery Command

### LINKED ISSUE #331: LIP0054: Some improvements to Message Recovery Command
1. Replace `crossChainMessages` with `trsParams.crossChainMessages` in [Verification](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0054.md#verification-1)

```
def verify(trs: Transaction) -> None:
    ...
    # Process basic checks for all CCMs.
    for ccmBytes in crossChainMessages: // ***
 ```
 
 I think, here we need to replace  `crossChainMessages` with `trsParams.crossChainMessages`, since no such individual `crossChainMessages` variable exists in upper scope. Just like it's done here on further steps
```
if RMTVerify(
  [sha256(ccmBytes) for ccmBytes in trsParams.crossChainMessages], // ***
```

2.  In `applyRecovery` & `forwardRecovery `, we have some common code, which can be moved to parent `execute` method
```
ccm.status = CCM_STATUS_CODE_RECOVERED
ccm.sendingChainID, ccm.receivingChainID = ccm.receivingChainID, ccm.sendingChainID

try:
    # Modules can verify the CCM.
    # The Token module verifies the escrowed balance in the CCM sending chain for the message fee.
    for each module mdl for which verifyCrossChainMessage exists:
        mdl.verifyCrossChainMessage(trs, ccm)
except:
    emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_INVALID_CCM_VERIFY_CCM_EXCEPTION)
    return
```
=> 
```
def execute(trs: Transaction) -> None:
    ...
    ccm = decode(crossChainMessageSchema, ccmBytes) 
    ccm.status = CCM_STATUS_CODE_RECOVERED
    ccm.sendingChainID, ccm.receivingChainID = ccm.receivingChainID, ccm.sendingChainID
    
    try:
        # Modules can verify the CCM.
        # The Token module verifies the escrowed balance in the CCM sending chain for the message fee.
        for each module mdl for which verifyCrossChainMessage exists:
            mdl.verifyCrossChainMessage(trs, ccm)
    except:
        emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_INVALID_CCM_VERIFY_CCM_EXCEPTION)
    
    # If the sending chain is the mainchain, recover the CCM.
    # This function never raises an error.
    if ccm.sendingChainID == getMainchainID():
        applyRecovery(trs, ccm)
    # If the sending chain is not the mainchain, forward the CCM.
    # This function never raises an error.
    elif ccm.sendingChainID != getMainchainID():
        forwardRecovery(trs, ccm)
        ...
```
3. `crossChainCommand` needs to be `ccm. crossChainCommand`
```
def applyRecovery(trs: Transaction, ccm: CCM) -> None:
    ...
    if ccm.module is not supported:
        emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_MODULE_NOT_SUPPORTED)
        return
    elif crossChainCommand is not supported: // ***
        emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_CROSS_CHAIN_COMMAND_NOT_SUPPORTED)
        return
        ...
```
Just like it's mentioned correctly here
// https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#apply
```
def apply(trs: Transaction, ccm: CCM) -> None:
   ...
   if ccm.module is not supported:
        bounce(ccm, CCM_STATUS_CODE_MODULE_NOT_SUPPORTED, CCM_PROCESSED_CODE_MODULE_NOT_SUPPORTED)
        return
    elif ccm.crossChainCommand is not supported: // ***
        bounce(ccm, CCM_STATUS_CODE_CROSS_CHAIN_COMMAND_NOT_SUPPORTED, CCM_PROCESSED_CODE_CROSS_CHAIN_COMMAND_NOT_SUPPORTED)
        return
        ...
```

### LINKED ISSUE #338: LIP0045: Update description for `ownChainAccountSchema`
// https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#properties-4

> On a sidechain, the own chain account is initialized as part of the mainchain registration command processing. It contains the following properties:
> 
> name: The name of the sidechain registered on the mainchain with the sidechain registration command.
> chainID: The chain ID assigned to the sidechain on the mainchain after processing the sidechain registration command.
> nonce: The chain nonce, an incremental integer indicating the total number of CCMs sent from the chain. The initial value is set to 0.

> On the manchain, the own chain account is present by default, set to an object with properties:
> 
> name = CHAIN_NAME_MAINCHAIN,
> ID = getMainchainID(), // ***
> nonce = 0,
> serialized with the JSON schema ownChainAccountSchema.
> 

Here `ID` needs to be replaced with `chainID`, since 
- In `ownChainAccountSchema`, we have `chainID`
- [getMainchainID()](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#mainchain-chain-id) return type is `ChainID`
- Also it will make it look similar to `On a sidechain` (object) part

### PR DESCRIPTION
**Pull request by gkoumout**
This PR resolves #331.

This PR resolves #338.

### CODE REVIEW

File: proposals/lip-0054.md [hunk 1]
```diff
@@ -427,6 +427,18 @@ def execute(trs: Transaction) -> None:
...
+        ccm.sendingChainID, ccm.receivingChainID = ccm.receivingChainID, ccm.sendingChainID
+
+        try:
+            # Modules can verify the CCM.
+            # The Token module verifies the escrowed balance in the CCM sending chain for the message fee.
+            for each module mdl for which verifyCrossChainMessage exists:
+                mdl.verifyCrossChainMessage(trs, ccm)
+        except:
+            emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_INVALID_CCM_VERIFY_CCM_EXCEPTION)
+            return
```
ricott1, line 440: I would not extract this code from the functions. This return statement is wrong, for instance

ricott1 (commented):
You introduced a protocol change. I would not refactor if not strictly necessary

proposals/lip-0054.md
gkoumout: Ok this return should be removed. The intention is not to introduce protocol change, just to avoid using the same code twice. 

I could withdraw this change, no problem from my side.

ricott1: Maybe my point is that we should keep refactoring changes to a minimum to avoid introducing bugs inadvertently. IMHO the current version is better too, I understand there is code duplication, but on the other hand there is a nice separation in each one of those functions that make the code nicely readable.

gkoumout: I see your point. I thought it is just a shortcut refactoring, but indeed it is not as simple as that and it does not make sense to spend effort to make it equivalent.

Concretely, in the common execution of those two functions, we have the following block:

```
    ccm.status = CCM_STATUS_CODE_RECOVERED
    ccm.sendingChainID, ccm.receivingChainID = ccm.receivingChainID, ccm.sendingChainID

    try:
        # Modules can verify the CCM.
        # The Token module verifies the escrowed balance in the CCM sending chain for the message fee.
        for each module mdl for which verifyCrossChainMessage exists:
            mdl.verifyCrossChainMessage(trs, ccm)
    except:
        emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_INVALID_CCM_VERIFY_CCM_EXCEPTION)
        return
```

In case of exception, then:

i) the rest of the function is not executed and 
ii) the ccm is appended to recovered CCMS.

By moving the common part in the `execute` function, then it is not straighforward to achieve the same outcome. The current version does (ii) but not (i) (i.e., the functions `applyRecovery` or `forwardRecovery` are erroneously called). On the other hand, if we just add `continue` after emitting the event, then we would achieve (i) but not (ii), i.e., we will not append the ccm to the `recoveredCCMs`.  The suggested way to fix it would be to do the following:

```
 for ccmBytes in trsParams.crossChainMessages:
        ccmID = sha256(ccmBytes)
        # Set ccmID as default topic to all events emitted in applyRecovery and forwardRecovery
        # (instead of the transaction ID). The default topic is added automatically by emitEvent.
        eventsDefaultTopic = ccmID

        ccm = decode(crossChainMessageSchema, ccmBytes)
        ccm.status = CCM_STATUS_CODE_RECOVERED
        ccm.sendingChainID, ccm.receivingChainID = ccm.receivingChainID, ccm.sendingChainID

        try:
            # Modules can verify the CCM.
            # The Token module verifies the escrowed balance in the CCM sending chain for the message fee.
            for each module mdl for which verifyCrossChainMessage exists:
                mdl.verifyCrossChainMessage(trs, ccm)

            # If the sending chain is the mainchain, recover the CCM.
            # This function never raises an error.
            if ccm.sendingChainID == getMainchainID():
                applyRecovery(trs, ccm)
            # If the sending chain is not the mainchain, forward the CCM.
            # This function never raises an error.
            elif ccm.sendingChainID != getMainchainID():
                forwardRecovery(trs, ccm)

        except:
            emitCCMEvent(ccm, CCM_PROCESSED_RESULT_DISCARDED, CCM_PROCESSED_CODE_INVALID_CCM_VERIFY_CCM_EXCEPTION)        
       
        # Append the recovered CCM to the list of recovered CCMs.
        # Notice that the ccm has been mutated in the applyRecovery and forwardRecovery functions
        # as the status is set to CCM_STATUS_CODE_RECOVERED (so that it cannot be recovered again).
        recoveredCCMs.append(encode(crossChainMessageSchema, ccm))    

```

This I think would work and have the intended behavior.

However it makes indeed the `execute` function more complicated (so overall we don't even make specs simpler) and introduces the un-necessary extra effort to check correctness compared to the previous version.  I would therefore recommend to revert this change and keep the current specs.

gkoumout: I reverted this change, since this seemed to be the most reasonable option.

ricott1 (approved):

janhack (approved):
Editorial approval

Issue #357, opened 2023-05-15
LIP51 - Fix supportTokenID function.


gkoumout:
## Motivation

In token module, the function `supportTokenID`, in case the specified token is already supported, the expected outcome should be that there is no state change

## Current Specifications

In the current specs, we add the token in the `supportedTokenIDs` array, creating therefore multiple occurences of the same token.

## Proposed Specifications

The following code block

```
if supportedTokens(chainID) exists:
    if supportedTokens(chainID).supportedTokenIDs == []:
        return

    add tokenID to supportedTokens(chainID).supportedTokenIDs, maintaining the array in lexicographical order

```

should take also care of the case that the tokenID already exists, for example:

```
if supportedTokens(chainID) exists:
    if supportedTokens(chainID).supportedTokenIDs == []:
        return
    if there does not exist an item with value `tokenID` in supportedTokens(chainID).supportedTokenIDs:
        add tokenID to supportedTokens(chainID).supportedTokenIDs, maintaining the array in lexicographical order
```

## Affected LIPs

LIP0051

PR #358, opened 2023-05-17
LIP 0051 - Fix `supportTokenID` function

mjerkov:
Resolves [Issue 357](https://github.com/LiskHQ/lips/issues/357).

## Rationale

Instead of Proposed specification, it uses the existing `isTokenSupported` function:

```
def supportTokenID(tokenID: TokenID) -> None:

    if not isTokenSupported(tokenID):
	    if supportedTokens(chainID) exists:
		    add tokenID to supportedTokens(chainID).supportedTokenIDs, maintaining the array in lexicographical order
	    else:
        	create an entry in the supported tokens substore with
	        key = chainID
	        value = encode(
	            schema=supportedTokensSchema,
        	    object={"supportedTokenIDs": [tokenID]}
	        )

    emitEvent(
        module=MODULE_NAME_TOKEN,
        name=EVENT_NAME_TOKEN_ID_SUPPORTED,
        data={"tokenID": tokenID},
        topics=[tokenID]
    )
```

for the following reasons:

- To be consistent with the requirement to emit an event `EVENT_NAME_TOKEN_ID_SUPPORTED` in *all* relevant cases.
- To avoid repeating the existing code.

### DIFF
--- a/proposals/lip-0051.md
+++ b/proposals/lip-0051.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/define-state-and-state-transitions-o
 Status: Draft
 Type: Standards Track
 Created: 2021-05-21
-Updated: 2023-05-09
+Updated: 2023-05-22
 ```
 
 ## Abstract


proposals/lip-0051.md
```diff
@@ -2246,29 +2246,17 @@ def removeAllTokensSupportFromChainID(chainID: ChainID) -> None:
...
 
-    else:
-        create an entry in the supported tokens substore with
-        key = chainID
-        value = encode(
-            schema=supportedTokensSchema,
-            object={"supportedTokenIDs": [tokenID]}
-        )
+    if not isTokenSupported(tokenID):
+	    if supportedTokens(chainID).supportedTokenIDs exists:
```
sergeyshemyakov, line 2251: The usual check in several other similar places is `if supportedTokens(chainID) exists`, I wonder if it is possible that `supportedTokens(chainID)` exists but it does not have the `supportedTokenIDs` property. Then the code below would create another / rewrite the entry in the supported tokens substore with key `chainID`.

For rigorousness and random bug prevention I would add another clause:
```
if supportedTokens(chainID).supportedTokenIDs exists:
    ...
else if supportedTokens(chainID) exists:
    # should never happen
    supportedTokens(chainID).supportedTokenIDs = [tokenID]
else:
    ...
```

sergeyshemyakov (approved):
I approve while having one remark since I will be out for the rest of the week. You could resolve my comment and then merge the PR.

proposals/lip-0051.md
gkoumout: I think here there is a typo. It should be 

`if supportedTokens(chainID) exists:`

Since if the entry exists, it will always have a `supportedTokenIDs` property..

gkoumout: This also replies to Sergey's comment.

gkoumout (changes requested):
Just left one comment for a typo.

proposals/lip-0051.md
```diff
@@ -2246,29 +2246,17 @@ def removeAllTokensSupportFromChainID(chainID: ChainID) -> None:
...
 
-    else:
-        create an entry in the supported tokens substore with
-        key = chainID
-        value = encode(
-            schema=supportedTokensSchema,
-            object={"supportedTokenIDs": [tokenID]}
-        )
+    if not isTokenSupported(tokenID):
+	    if supportedTokens(chainID) exists:
```
gkoumout, line 2251: I think indentation is wrong here (both in the nested if-else and in the "add tokenID" part..). It should be same indentation as in the beginning of the function and the else-"create an entry..." part.

gkoumout (changes requested):

gkoumout (approved):

has5aan (approved):

janhack (approved):
Editorial approval

Issue #362, opened 2023-05-25
Specifications in LIP 58 differs from the implementation


janhack:
## Motivation

The current height and the value of `nextHeight` in the function `setBFTParameters` are computed differently in the implementation and specifications. 

## Current Specifications

In [LIP 58 ](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#execution-4) the current height in the function `getCurrentHeight` is given by `genesisHeight` if `length(bftVotes.blockBFTInfos) == 0` and the function `getCurrentHeight` is used in the function `setBFTParameters`. However, in the [implementation](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/bft/method.ts#L222) in that case `bftVotes.maxHeightPrevoted` is used. Note that due the the [initialization in the genesis block processing](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#before-application-processing), when the function `setBFTParameters` is called in the "After Application Processing" phase of the genesis block, then  `bftVotes.maxHeightPrecommitted == genesisHeight` holds. This means both approaches are equivalent.

## Proposed Specifications

As the genesis height is currently not available in the BFT logic context in the implementation, it would be easiest to update the specification to use `bftVotes.maxHeightPrecommitted` instead of  `genesisHeight` .

## Affected LIPs

- [LIP 0058](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058)

PR #363, opened 2023-05-25
Remove usage of genesis height

janhack:
Resolves https://github.com/LiskHQ/lips/issues/362


AnanyaSshri (approved):

proposals/lip-0058.md
```diff
@@ -763,8 +763,6 @@ In the before application processing stage of a genesis block `b`, the BFT store
   * `bftVotes.maxHeightCertified = b.header.height`,
   * `bftVotes.blockBFTInfos` is an empty array,
   * `bftVotes.activeValidatorsVoteInfo` is an empty array.
-* The variable `genesisHeight` is initialized as follows:
```
gkoumout, line 766: So, to make sure I completely follow the reasoning behind this change: According to old specs, this variable was initialized in the genesis block "before application processing" stage and should have been available in the context. 

However (correct me if I am wrong) this was not done in the implementation. So the reason of removing this is that is was not implemented, but somehow having this variable is not strictly necessary, since we can recover it from the BFT Votes substore after the genesis block?  

Removing this would imply that `genesisHeight` would not be available in the future (after `blockBFTInfos` does not contain the genesis block info anymore). But it seems that this is not necessary to have in any other part of the LIP.. 

Let me know if we are on the same page.

proposals/lip-0058.md
```diff
@@ -82,7 +82,6 @@ As explained in [Section "Comparison to the Lisk-BFT paper" in LIP 0056][lip-005
 | Name | Description |
 |------|-------------|
 | `uint32be(x)` | For an integer `x` with `0 <= x < 2^32`, the function returns 4 bytes representing the big endian unsigned integer serialization of `x`. |
```
gkoumout, line 84: Optional suggestion: We can even remove this function, since we kind of stopped using it from all other LIPs. Not a big deal, but since it is only used in 3 places over the whole LIP, we could just omit it (and even completely remove the table) and just replace it whenever used by `x.to_bytes(4,'big')`, see e.g., [notation in PoS LIP](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#snapshot-substore-1).

But as I said, this is a completely optional suggestion and it is okay to skip it.

proposals/lip-0058.md
```diff
@@ -82,7 +82,6 @@ As explained in [Section "Comparison to the Lisk-BFT paper" in LIP 0056][lip-005
...
-| `genesisHeight` | This variable holds the height of the genesis block of the blockchain. It is generally available in the context of the consensus domain and initialized during the genesis block processing or, afterwards, at the beginning of starting the node. |
 | `//` | This symbol represents integer division. |
```
gkoumout, line 86: Line 297: Section title "Internal Function" --> "Internal Functions"

gkoumout (commented):
Looks good from my side. Just left a question to make sure I understand correctly the introduced change and two minor syntactic comments.

proposals/lip-0058.md
```diff
@@ -763,8 +763,6 @@ In the before application processing stage of a genesis block `b`, the BFT store
   * `bftVotes.maxHeightCertified = b.header.height`,
   * `bftVotes.blockBFTInfos` is an empty array,
   * `bftVotes.activeValidatorsVoteInfo` is an empty array.
-* The variable `genesisHeight` is initialized as follows:
-  * `genesisHeight = b.header.height`.
 
 #### After Application Processing
```
gkoumout, line 768: Question about lines 779-780: Those two steps actually suggest a queue data structure where new entries are added to the one side and old entries are deleted from the other side (it would be inefficient to use a list/array since at each step need to re-assign all indexes). I was just curious about how this is actually implemented..

proposals/lip-0058.md
janhack: Yep, the implementation did not use or initialize the genesis height, but instead was using `bftVotes.maxHeightPrevoted`. So the implementation is doing exactly what the same as the updated version of this LIP.

proposals/lip-0058.md
janhack: I would not make cosmetic changes to the LIP at this stage unless something is currently unclear or confusing.

proposals/lip-0058.md
janhack: Fixed.

proposals/lip-0058.md
janhack: As it is outside the scope of this LIP update, it could be clarified on Slack? Note for the serialized data we cannot use a queue.

gkoumout: Sounds good, it's anyways more of a curiosity question.

gkoumout (approved):

PR #367, opened 2023-06-01
Update LIP 0052 - Additional check in the `transferCrossChain` command

mjerkov:
## Description

As of now, it is possible to use the `transferCrossChain` command to initiate the cross-chain NFT transfer in the way it is not intended, i.e., to use it for transferring NFT on the same chain.

## Proposed Solution

In `transferCrossChain` command, we add the following additional check, to ensure that `receivingChainID` is not the same as `OWN_CHAIN_ID` (i.e., the `sendingChainID`):

```
if receivingChainID == OWN_CHAIN_ID:
        raise Exception("Receiving chain cannot be the sending chain.")
```

### DIFF
--- a/proposals/lip-0052.md
+++ b/proposals/lip-0052.md
@@ -1828,6 +1828,9 @@ def transferCrossChain(
     includeAttributes: bool
 ) -> None:
 
+    if receivingChainID == OWN_CHAIN_ID:
+        raise Exception("Receiving chain cannot be the sending chain.")
+
     if len(data) > MAX_LENGTH_DATA:
         emitFailedCrossChainTransferEvent(senderAddress, nftID, amount, receivingChainID, recipientAddress, RESULT_DATA_TOO_LONG)
         raise Exception("Data field is too long")


sergeyshemyakov (changes requested):
I think you should also add this check to the [verification of cross-chain transfer command](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0052.md#verification-1).

Otherwise looks good!

sergeyshemyakov (approved):
Approved!

proposals/lip-0052.md
```diff
@@ -1828,6 +1831,9 @@ def transferCrossChain(
     includeAttributes: bool
 ) -> None:
 
+    if receivingChainID == OWN_CHAIN_ID:
```
gkoumout, line 1834: I just realized that in this function in all other checks we emit failure events, so perhaps we should do this here as well for consistency (same for the token module..). Not sure how important it is though, espcecially for the token module..

gkoumout (commented):
Left one comment.

proposals/lip-0052.md
mjerkov: I agree with the suggested changes and have pushed the related commit. I will also expand the Description and Proposed solution accordingly.

proposals/lip-0051.md
```diff
@@ -193,6 +193,7 @@ We further use the utility function `getMainchainID()` defined in [LIP 0037][lip
 | `TOKEN_ID_NOT_AVAILABLE` | uint32 | 11 | Used when the specified token ID is not available. |
 | `TOKEN_ID_NOT_NATIVE` | uint32 | 12 | Used when the specified token ID is not native to the chain. |
 | `INSUFFICIENT_ESCROW_BALANCE` | uint32 | 13 | Used when the escrowed account does not have sufficient balance. |
+| `INCORRECT_RECEIVING_CHAIN` | uint32 | 14 | Used when, during the cross-chain token transfer, the receiving chain is set to be equal to the sending chain. |
```
gkoumout, line 196: Name looks good, we could also use `INVALID_RECEIVING_CHAIN`

gkoumout (commented):
Left one minor comment, looks good otherwise.

sergeyshemyakov (approved):
Looks good to me

has5aan (approved):

Incede (approved):

sergeyshemyakov (approved):

gkoumout (approved):
I did not check the LIP52 changes, approving for the LIP 51 part.

proposals/lip-0051.md
```diff
@@ -2447,7 +2448,7 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
   * The `supplySubstore` array must be in ascending order of `tokenID`.
   * Across all elements of the `escrowSubstore` array, the `tokenID` must be unique (it appears at most once).
   * The `escrowSubstore` array must be in ascending order of `tokenID`.
-  * The `supportedTokensSubstore` array must be in lexicographical order of `chainID`.  For each entry of this array, the `supportedTokenIDs` array should be in lexicographic order.
+  * The `supportedTokensSubstore` array must be in lexicographical order of `chainID` and the `chainID` should be of length `CHAIN_ID_LENGTH` or `0`. For each entry of this array, the `supportedTokenIDs` array should be in lexicographic order.
```
gkoumout, line 2451:
```suggestion
  * The `supportedTokensSubstore` array must be in lexicographical order of `chainID`. For each entry of this array, the `chainID` property should be of length `CHAIN_ID_LENGTH` or `0` and the `supportedTokenIDs` array should be in lexicographic order.
```

gkoumout: Also, I think we should add the constraint that if there is an entry of length `0`, then it is the only entry of the array (since this implies that all tokens are supported, therefore does not make sense to have other entries).

proposals/lip-0051.md
```diff
@@ -2460,7 +2461,8 @@ storeValue = encode(
             "module": lockedBalance.module,
             "amount": lockedBalance.amount
         } for each lockedBalance in userEntry.lockedBalances]
-)
+    }
+)    
 ```
```
gkoumout, line 2467: Comment about line 2486: 

While creating entries in the supported tokens substore, in case `chainID = EMPTY_BYTES`, in theory the value of `supportedTokenIDs` could be an array of arbitrary long size containing tokenIDs. We should not write this in the state, instead we should use as value the empty array.

I see 2 possible ways to achieve this. 

1.  Either add an if condition in line 2488 to distinguish the case where `supportedTokensEntry.chainID = EMPTY_BYTES` or
2. Add an extra requirement in the verification above (e.g., line 2451) that the unique entry of the `genesisBlockAssetObject.supportedTokensSubstore` array with length 0 should have `entry.supportedTokenIDs = []`.

gkoumout (changes requested):
Left few comments on the new updates.

proposals/lip-0051.md
mjerkov: Agreed, thanks for this suggestion.

proposals/lip-0051.md
mjerkov: I would rather write this in the verification section. I suggest the following, please check:
```
The `supportedTokensSubstore` array must either:
    - Have a unique entry with `chainID` of length `0` and no other entries. This is the case when all tokens are supported, i.e., `chainID == ALL_SUPPORTED_TOKENS_KEY`. In that case, the the `supportedTokenIds` is an empty array.
    - Have all entries with `chainID` property of length `CHAIN_ID_LENGTH` and `supportedTokenIDs` array in lexicographic order. 
```

proposals/lip-0051.md
mjerkov: Please see my comment below.

proposals/lip-0051.md
gkoumout: Sounds good. Syntactically, we should use either ... or ...

For the second bullet point don't forget to copy/paste that the `supportedTokensSubstore` array should be sorted by lexicographic order of `chainID`.

proposals/lip-0051.md
```diff
@@ -193,6 +193,7 @@ We further use the utility function `getMainchainID()` defined in [LIP 0037][lip
 | `TOKEN_ID_NOT_AVAILABLE` | uint32 | 11 | Used when the specified token ID is not available. |
 | `TOKEN_ID_NOT_NATIVE` | uint32 | 12 | Used when the specified token ID is not native to the chain. |
 | `INSUFFICIENT_ESCROW_BALANCE` | uint32 | 13 | Used when the escrowed account does not have sufficient balance. |
+| `INVALID_RECEIVING_CHAIN` | uint32 | 14 | Used when, during the cross-chain token transfer, the receiving chain is set to be equal to the sending chain. |
 
 ### Type Definition
```
sergeyshemyakov, line 198: I leave my comment here but it concerns [supported tokens substore](https://github.com/LiskHQ/lips/blob/514f9fb029f3fb229d1ec7323417ad69ba08ade8/proposals/lip-0051.md#supported-tokens-substore).

I would modify the description of store keys like this:
"Each store key is the identifier of the chain to which the supported tokens are native: `chainID`, or `ALL_SUPPORTED_TOKENS_KEY`."

proposals/lip-0051.md
```diff
@@ -2447,7 +2448,9 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
   * The `supplySubstore` array must be in ascending order of `tokenID`.
   * Across all elements of the `escrowSubstore` array, the `tokenID` must be unique (it appears at most once).
   * The `escrowSubstore` array must be in ascending order of `tokenID`.
-  * The `supportedTokensSubstore` array must be in lexicographical order of `chainID`.  For each entry of this array, the `supportedTokenIDs` array should be in lexicographic order.
+  * The `supportedTokensSubstore` array must adhere to one of the following two mutually exclusive conditions:
+    * Have a unique entry with `chainID` of length `0` and no other entries. This indicates that all tokens are supported, i.e., `chainID == ALL_SUPPORTED_TOKENS_KEY`. In this case, the `supportedTokenIds` should be an empty array.
+    * Have all entries with `chainID` property of length `CHAIN_ID_LENGTH`. The `supportedTokenIDs` array must be in lexicographic order, and the `supportedTokensSubstore` array must be in lexicographic order of `chainID`.
```
sergeyshemyakov, line 2453: In similar cases in other LIPs we usually require that all entries are unique, you could also add it here for consistency.

sergeyshemyakov (approved):
I approve but there are still a couple of open points from my side.

proposals/lip-0051.md
mjerkov: Sounds good, I would do that (also for LIP 0052).

proposals/lip-0051.md
```diff
@@ -2448,7 +2448,9 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
   * The `supplySubstore` array must be in ascending order of `tokenID`.
   * Across all elements of the `escrowSubstore` array, the `tokenID` must be unique (it appears at most once).
   * The `escrowSubstore` array must be in ascending order of `tokenID`.
-  * The `supportedTokensSubstore` array must be in lexicographical order of `chainID` and the `chainID` should be of length `CHAIN_ID_LENGTH` or `0`. For each entry of this array, the `supportedTokenIDs` array should be in lexicographic order.
+  * The `supportedTokensSubstore` array must adhere to one of the following two mutually exclusive conditions:
```
gkoumout, line 2451: This is in case the array is non-empty. An empty array is also valid input. So I would say that if the array is non-empty, then it must satisfy those constraints. 

Minor comment, although a like the "mutually exclusive" phrase, might be confusing for devs, to it would be good to use either ... or ... . But not strong opinion here.

gkoumout (changes requested):

proposals/lip-0051.md
mjerkov: Hmm, what is your suggestion, how to write it to be more clear?

proposals/lip-0051.md
```diff
@@ -2448,9 +2448,9 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
...
-  * The `supportedTokensSubstore` array must adhere to one of the following two mutually exclusive conditions:
-    * Have a unique entry with `chainID` of length `0` and no other entries. This indicates that all tokens are supported, i.e., `chainID == ALL_SUPPORTED_TOKENS_KEY`. In this case, the `supportedTokenIds` should be an empty array.
-    * Have all entries with `chainID` property of length `CHAIN_ID_LENGTH`. The `supportedTokenIDs` array must be in lexicographic order, and the `supportedTokensSubstore` array must be in lexicographic order of `chainID`.
+  * The `supportedTokensSubstore` array, if non-empty, must adhere to one of the following two mutually exclusive conditions:
+    * Have a single entry with `chainID` of length `0` and no other entries. This indicates that all tokens are supported, i.e., `chainID == ALL_SUPPORTED_TOKENS_KEY`. In this case, the `supportedTokenIds` should be an empty array.
+    * Have all entries unique and with `chainID` property of length `CHAIN_ID_LENGTH`. The `supportedTokenIDs` array must be in lexicographic order, and the `supportedTokensSubstore` array must be in lexicographic order of `chainID`.
```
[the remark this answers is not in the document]
mjerkov: Done, check [commit](https://github.com/LiskHQ/lips/pull/367/commits/05ce094f5d8e92ec1cb0abaac28217b3fffe08e6).

proposals/lip-0051.md
gkoumout: Shouldn't this be resolved now?

proposals/lip-0051.md
```diff
@@ -2447,7 +2448,9 @@ Let `genesisBlockAssetBytes` be the `data` bytes included in the block assets fo
   * The `supplySubstore` array must be in ascending order of `tokenID`.
```
gkoumout, line 2448: small typo on line 2447, just caught it: "across all element" --> "across all elements"

gkoumout (approved):
Well done!

has5aan (approved):

PR #377, opened 2023-06-07
Update LIP 0057 - change bft capping value

### LINKED ISSUE #376: Update BFT capping constant
## Motivation

Increase the maximum BFT weight of a validator to 10%

## Current Specifications

In current PoS module, the maximum BFT weight of a validator is set to 5%,

## Affected LIPs

[LIP0057](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md)

gkoumout:
Fixes #376.


proposals/lip-0057.md
```diff
@@ -64,7 +64,7 @@ This part of the state store is used to maintain the timestamp of the last block
...
 
 Note that further we choose `floor(2/3 * aggregateBFTWeight) + 1` as precommit and certificate threshold for PoS chains, see [LIP 0058][lip-0058] for an explanation of these parameters. Here `aggregateBFTWeight` is the sum of BFT weights of the active validators. This choice of BFT parameters means that the safety and liveness of the BFT consensus protocol is ensured if `<1/3 * aggregateBFTWeight` is malicious (for the case of changing validators the guarantees are slightly weaker). The capping of BFT weights to 5 % for the Lisk mainchain implies that at least 7 validators would need to be malicious as any set of 6 validators would have `<1/3 * aggregateBFTWeight` BFT weight. Similarly, only a set of at least 14 validators could have BFT weight of `>= floor(2/3 * aggregateBFTWeight) + 1` and would therefore be able to generate any certificate. Hence, the capping ensures a minimum of decentralization in the validator set, even in the case of very unbalanced validator weight distribution. In practice, we expect the capping to only effect very few validators, in particular, as rewards are also distributed proportional to the BFT weights.
```
ricott1, line 69:
```suggestion
Note that further we choose `floor(2/3 * aggregateBFTWeight) + 1` as precommit and certificate threshold for PoS chains, see [LIP 0058][lip-0058] for an explanation of these parameters. Here `aggregateBFTWeight` is the sum of BFT weights of the active validators. This choice of BFT parameters means that the safety and liveness of the BFT consensus protocol is ensured if `<1/3 * aggregateBFTWeight` is malicious (for the case of changing validators the guarantees are slightly weaker). The capping of BFT weights to 10 % for the Lisk mainchain implies that at least 7 validators would need to be malicious as any set of 6 validators would have `<1/3 * aggregateBFTWeight` BFT weight. Similarly, only a set of at least 14 validators could have BFT weight of `>= floor(2/3 * aggregateBFTWeight) + 1` and would therefore be able to generate any certificate. Hence, the capping ensures a minimum of decentralization in the validator set, even in the case of very unbalanced validator weight distribution. In practice, we expect the capping to only effect very few validators, in particular, as rewards are also distributed proportional to the BFT weights.
```

ricott1: Also the rationale about the number of nodes must be updated.

ricott1 (changes requested):
We need to update the rationale about the capping limit on mainchain plus one leftover 5%

ricott1 (approved):

ishantiw (approved):

janhack (approved):
Looks good.

Could we still link an issue for updating the implementation?

ishantiw:
Related SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8567

janhack (approved):
Editorial approval

PR #390, opened 2023-06-26
remove call to isLive

### LINKED ISSUE #389: Remove call to `isLive` in `isChannelActive` in the sidechain case
## Motivation

In the interoperability module, the function [`isChannelActive`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#ischannelactive) informs whether a channel with another chain is open an active. 

Whevever `isChannelActive(chainID)` is called in a sidechain, in case there exists an account for the chain with id `chainID` exists, then it should return true if and only if `chainAccount(chainID).status == CHAIN_STATUS_ACTIVE`, because this is the only case when the channel is active. In all other cases, we know that the channel can not be active: if status is `CHAIN_STATUS_REGISTERED`, then the channel has not been activated yet; on the other hand, if status is `CHAIN_STATUS_TERMINATED`, then the channel is not active anymore.

## Current Specifications

Currently, there is an un-necessary call to `isLive`. In particular, for the case that we consider here, we return:

```
    # Processing on a sidechain.
    else:
        # Account may not exist on a sidechain.
            return isLive(chainID) and chainAccount(chainID).status == CHAIN_STATUS_ACTIVE`
```
The function `isLive` will return False in case status is `CHAIN_STATUS_TERMINATED`. In all other cases returns True. Therefore, the call to `isLive` is redundant, because in case status is `CHAIN_STATUS_TERMINATED`, the condition `chainAccount(chainID).status == CHAIN_STATUS_ACTIVE` does not hold

## Proposed Specifications

Remove the unnecessary call to `isLive` function, i.e., replace the pseudocode above by:

```
    # Processing on a sidechain.
    else:
        # Account may not exist on a sidechain.
            return chainAccount(chainID).status == CHAIN_STATUS_ACTIVE
```

## Affected LIPs

[LIP 0045](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md)

gkoumout:
This PR fixes #389 .

### DIFF
--- a/proposals/lip-0045.md
+++ b/proposals/lip-0045.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/introduce-interoperability-module/29
 Status: Draft
 Type: Standards Track
 Created: 2021-05-21
-Updated: 2023-06-14
+Updated: 2023-07-03
 Requires: 0043, 0049, 0053, 0054
 ```
 
--- a/proposals/lip-0045.md
+++ b/proposals/lip-0045.md
@@ -1336,7 +1336,7 @@ def isChannelActive(chainID: ChainID) -> bool:
     else:
         # Account may not exist on a sidechain.
         if chainAccount(chainID) exists:   
-            return isLive(chainID) and chainAccount(chainID).status == CHAIN_STATUS_ACTIVE
+            return chainAccount(chainID).status == CHAIN_STATUS_ACTIVE
         
         # Check that mainchain account exists.
         if chainAccount(getMainchainID()) does not exist:


ricott1 (approved):

mitsuaki-u (approved):

mitsuaki-u (changes requested):
I would also add `timestamp` as a parameter since `isLive` called in the mainchain portion still requires it

mitsuaki-u (approved):

AndreasKendziorra (approved):

janhack (approved):
Editorial approval

PR #391, opened 2023-06-26
Add event names to table in LIP 0070

### LINKED ISSUE #383: Event names not defined in LIP 70
## Current Specifications

LIP 70 defines events, but the constants defining the event name are missing from the [specification table](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0070.md#constants)

## Proposed Specifications

Add the constants for event name in the specifications table, setting the value same as the one set in the implementation.

## Affected LIPs

[LIP 70](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0070.md)

gkoumout:
This PR fixes #383

### DIFF
--- a/proposals/lip-0070.md
+++ b/proposals/lip-0070.md
@@ -6,7 +6,7 @@ Discussions-To: https://research.lisk.com/t/introduce-reward-sharing-mechanism/3
 Status: Draft
 Type: Standards Track
 Created: 2022-10-26
-Updated: 2023-06-01
+Updated: 2023-07-03
 Required: 0022, 0023, 0057
 ```
 
--- a/proposals/lip-0070.md
+++ b/proposals/lip-0070.md
@@ -49,6 +49,13 @@ Here we define the constants related to reward sharing. All other PoS module con
 | `COMMISSION_INCREASE_PERIOD` | uint32 | 260000 | Determines how frequently (after how many blocks) a validator can increase their commission. |
 | `MAX_COMMISSION_INCREASE`| uint32 | 500 | Determines the maximum allowed increase on the commission per transaction. |
 
+### Event Names
+
+| Name | Type | Value | Description |
+|------|------|-------|-------------|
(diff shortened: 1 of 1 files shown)


ricott1 (approved):

proposals/lip-0070.md
```diff
@@ -49,6 +49,13 @@ Here we define the constants related to reward sharing. All other PoS module con
 | `COMMISSION_INCREASE_PERIOD` | uint32 | 260000 | Determines how frequently (after how many blocks) a validator can increase their commission. |
 | `MAX_COMMISSION_INCREASE`| uint32 | 500 | Determines the maximum allowed increase on the commission per transaction. |
 
+### Event Names
+
+| Name | Type | Value | Description |
+|------|------|-------|-------------|
+| `EVENT_NAME_COMMISSION_CHANGE` | string | "commissionChange" | Used for events emitted whenever validators change their commission. |
+| `EVENT_NAME_REWARDS_ASSIGNED` | string | "rewardsAssigned" | Used for events emitted whenever shared rewards are assigned to stakers. |
```
bobanm, line 57: I'm not familiar if there is any convention here, but the constant name and the value of `EVENT_NAME_REWARDS_ASSIGNED` is in past tense, while `EVENT_NAME_COMMISSION_CHANGE` is in present tense.

To me, it makes more sense that both of these events use past tense, as they trigger when the action has occurred.

No matter of which tense is selected, it would make it more consistent if both events use the same tense.

bobanm (changes requested):

vjaiman (approved):

proposals/lip-0070.md
gkoumout: The intention is that "assigned" comes from the verb "assign" in past tense, while "change" is just the noun "change". 

The strings used for names are the ones used in SDK implementation and I guess we would not like to change them.

bobanm (approved):

janhack (approved):
Editorial approval

Issue #364, opened 2023-05-31
LIP 0048 has incomplete rationale section


janhack:
## Motivation

The Fee module as introduced in LIP 0048 has evolved significantly since the initial version of the LIP was published. However, the LIP does not give any rationale for the current design and does not explain how the Fee module and its public method `payFee` is supposed to be used in a blockchain created with the Lisk SDK. It is therefore hard for anyone reading the LIP to understand the design rationale for fees in the Lisk SDK and also hard for developers creating their own module to understand how to utilize the Fee module.

## Proposed solution

Expand the rationale section of the LIP, in particular, addressing the following questions:

- What is the reasoning behind the current fee design in Lisk?
- How are fees for cross-chain messages and transaction fees different and why?
- What is the idea of a fee pool and how can it be used? This point is already shortly addressed, but it could be nice to expand a bit on it with some more concrete ideas (e.g., funds could be put in a community pool that can be spend via on-chain governance or could be proportionally shared among validators per round).
- What is the purpose of the function `payFee`? How should it be used by different modules? 

## Affected LIPs

LIP 0048

PR #392, opened 2023-06-26
Rewrite LIP 0048 Rationale section

mjerkov:
Resolves Issue [#364](https://github.com/LiskHQ/lips/issues/364).

## Note

Apart from the above mentioned Issue, this PR corrects the following typo: It replaces `ccm.ccmProcessing` by `ctx.ccmProcessing` in [these lines](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0048.md?plain=1#L182-L184).


proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when the minimum fee per byte is set to zero.
+As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee, which can be set to zero. For a transaction `trs`, the minimum fee is determined by its size and the proportionality constant `MIN_FEE_PER_BYTE`: 
+
+```python
+trsSize = length(encodeTransaction(trs))
+minFee = MIN_FEE_PER_BYTE * trsSize
+```
+
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when it is set to zero.
```
ricott1, line 43: We should add a rationale for why we allow for this. Specifically, I think it has to do with using LSK as the fee token in sidechains.

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
 ### Fee Token ID
 
-Each chain can configure the token used to pay fees. On the Lisk mainchain, the token used for transaction fees is the LSK token.
+Each chain can configure the token used to pay fees, as defined by `TOKEN_ID_FEE`. On the Lisk mainchain, the token used for transaction fees is the LSK token.
 
 ### Minimum Fee per Transaction
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when the minimum fee per byte is set to zero.
+As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee, which can be set to zero. For a transaction `trs`, the minimum fee is determined by its size and the proportionality constant `MIN_FEE_PER_BYTE`:
```
ricott1, line 36:
```suggestion
As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee. For a transaction `trs`, the minimum fee is determined by its size and the proportionality constant `MIN_FEE_PER_BYTE`, which can be set to zero: 
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
 
 ## Rationale
 
+The fee structure presented here serves to allow defining sufficient cost to manage state growth and prevent transaction pool spamming. In principle, any transaction resulting in a state change should incur a fee. This reasoning is implemented in the `payFee` method, which can be invoked by other modules.
```
ricott1, line 28: I don't understand how the fee prevents transaction pool spamming, can you elaborate a bit?

ricott1, line 28:
```suggestion
The fee structure presented here serves to allow defining sufficient cost to manage state growth and prevent transaction pool spamming. In principle, any transaction resulting in a state change could pay a fee. This reasoning is implemented in the `payFee` method, which can be invoked by other modules.
```

mjerkov: Agreed. Maybe this?
```suggestion
The fee structure presented here serves to allow defining sufficient cost to manage state growth and prevent transaction pool spamming. In principle, any transaction resulting in a state increase should pay a fee. This reasoning is implemented in the `payFee` method, which can be invoked by other modules.
```

mjerkov: I'm not sure either, I believe I picked this up in some discussions related to fees. Please suggest how to rephrase it.

proposals/lip-0048.md
mjerkov: Noted.

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+```python
+trsSize = length(encodeTransaction(trs))
+minFee = MIN_FEE_PER_BYTE * trsSize
+```
+
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when it is set to zero.
+
+### Fee Pool
+
+Chains can choose to either burn the minimum fee - to discourage validators from sending zero-cost transactions in the blocks they generate - or, on the other hand, to transfer it to a dedicated pool address, `ADDRESS_FEE_POOL`.
```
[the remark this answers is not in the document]
mjerkov: Noted.

proposals/lip-0048.md
gkoumout, line 43:
```suggestion
Note that chains can freely configure their `MIN_FEE_PER_BYTE`, which can be even set to zero. The value `MIN_FEE_PER_BYTE` can not be modified, unless a hard fork occurs. For this reason, we allow sidechains the possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block for which the minimum fee does not apply. This could be useful, for example, for sidechains that want to use LSK as the fee token: initially, no LSK tokens exist in the sidechain, so the chain would not be able to have any transactions, unless `MIN_FEE_PER_BYTE = 0`. By allowing an initial exception period, users have some time window to transfer LSK tokens to this chain; after this time, `MIN_FEE_PER_BYTE` could apply with a reasonable non-zero value. 
```

gkoumout: This is just a suggested text and perhaps can be improved; I hope you get the main point of the suggestion.

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when it is set to zero.
+
+### Fee Pool
+
+Chains can choose to either burn the minimum fee - to discourage validators from sending zero-cost transactions in the blocks they generate - or, on the other hand, to transfer it to a dedicated pool address, `ADDRESS_FEE_POOL`. 
+
+The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants for various purposes like funding a community pool for on-chain governance spending, sharing the funds among validators proportionally per round, incentivizing ecosystem developers, supporting chain maintenance and improvement, rewarding user staking/voting activity, etc.
+
+In the Lisk mainchain, the minimum fee is always burned.
```
gkoumout, line 51: Perhaps it would be good to give an explanation of "why" we do so (if we have a good one..). I guess one of them was to decrease the inflation, although they way we do it does not seem to have justification.

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+minFee = MIN_FEE_PER_BYTE * trsSize
+```
+
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when it is set to zero.
+
+### Fee Pool
+
+Chains can choose to either burn the minimum fee - to discourage validators from sending zero-cost transactions in the blocks they generate - or, on the other hand, to transfer it to a dedicated pool address, `ADDRESS_FEE_POOL`. 
+
+The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants for various purposes like funding a community pool for on-chain governance spending, sharing the funds among validators proportionally per round, incentivizing ecosystem developers, supporting chain maintenance and improvement, rewarding user staking/voting activity, etc.
```
gkoumout, line 49: perhaps we could give a more concrete example, based on DEX?  Otherwise, this paragraph looks good to me.

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
 
 ## Rationale
 
+The fee structure presented here serves to allow defining sufficient cost to manage state growth and prevent transaction pool spamming. In principle, any transaction resulting in a state change should incur a fee. This reasoning is implemented in the `payFee` method, which can be invoked by other modules.
+
```
gkoumout, line 29: I would suggest the following structure:

Subsection: Transaction fees. Include the "fee token ID" and "Minimum Fee Per Transaction" sub-sub sections.

Subsetion: CCM fees.

Then the discussion on fee pool applies to both trs and ccm fees. 

Also, we could then expalin a bit more in detail the `payFee` functionality, perhaps with one example of a transaction when it is called (e.g., validator registration) and one example of a ccm processing when it is called (e.g., need to initialize a user account).

proposals/lip-0048.md
```diff
@@ -25,15 +25,34 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+In the Lisk mainchain, the minimum fee is always burned.
+
+### Cross-chain Message Fees
 
-Chains can choose that this minimum fee gets either burned (to avoid that validators can send transactions in the blocks they generate without cost) or transferred to the specified pool address `ADDRESS_FEE_POOL`. The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants. In the Lisk mainchain, the minimum fee is always burned.
+Cross-chain messages or CCMs facilitate interoperability by enabling cross-chain transfer of data. They require an associated fee, distinct from the usual transaction fee, paid by the relayer. Similarly to transaction fees, the minimum CCM fee can be either burned or transferred to a fee pool, depending on the chain configuration.
 
 ## Specification
```
gkoumout, line 58: Btw I also have a comment about the verify transaction function (lines 240-255):

In line 245 we set `MIN_FEE_PER_BYTE` to 0. While I understand the intention, strictly speaking this is wrong, since we over-write the initial value.

Instead I think that what we really want to do is:

```
    if h < MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE:
	    # Set minFee to 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
        minFee = 0
    else:
        trsSize = length(encodeTransaction(trs))
        minFee = MIN_FEE_PER_BYTE * trsSize
```

gkoumout (changes requested):
Looks good to me, just left some comments for improvements/additions. 

Moreover, in order to be complete, we might want to make more clear the message fee part, i.e., which part is managed by fee module and which not (payMessageFee and update escrows happens by token module), although it might not be so easy. Curious to hear the opinion of Ale on that as well.

proposals/lip-0048.md
mjerkov: Noted.

proposals/lip-0048.md
mjerkov: Agreed, it will be more readable that way.

proposals/lip-0048.md
mjerkov: Very good, thanks for the suggestion!

proposals/lip-0048.md
mjerkov: I think I understand now what your question implied, will try to correct it.

proposals/lip-0048.md
mjerkov: I don't have an idea how to give a more concrete example based on DEX, please instruct me. All I know is that DEX uses the funds collected in a pool to incentivize the validators, which is already generally mentioned in the paragraph above.
Or did you have something more specific in mind?

proposals/lip-0048.md
mjerkov: Yes, I don't have a good idea. Maybe I can write something on the line of:
```suggestion
In the Lisk mainchain, the minimum fee is always burned, reducing the total supply and thus introducing a deflationary pressure to counteract the inflation.
```

proposals/lip-0048.md
mjerkov: Yes, I believe you're right. Do you think the implementation will change (I believe not)?

mjerkov:
I pushed another commit, still didn't know how to write the part on hooks and ccm (handling by token module et al.). Maybe you have a suggestion...

proposals/lip-0048.md
```diff
@@ -25,32 +25,43 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
 
 ## Rationale
 
-The fee structure presented here serves to allow defining sufficient cost to manage state growth and prevent transaction pool spamming. In principle, any transaction resulting in a state change should incur a fee. This reasoning is implemented in the `payFee` method, which can be invoked by other modules.
+### Initialization Fee
```
gkoumout, line 28: In my opinion, starting with this topic is a bit confusing. Moreover, not all extra-command fees are really initialization fees, for example the sidechain registration fee. 

I would suggest to start with the transaction fee subsection. 

We could even link to the transaction schema LIP and start by saying that each transaction has a fee.

proposals/lip-0048.md
```diff
@@ -25,32 +25,43 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+#### Fee Token ID
 
 Each chain can configure the token used to pay fees, as defined by `TOKEN_ID_FEE`. On the Lisk mainchain, the token used for transaction fees is the LSK token.
 
-### Minimum Fee per Transaction
+#### Minimum Fee per Transaction
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee, which can be set to zero. For a transaction `trs`, the minimum fee is determined by its size and the proportionality constant `MIN_FEE_PER_BYTE`: 
+As introduced in [LIP 0013][lip-0013], all transactions must have a transaction fee greater or equal to a *minimum fee*, which is introduced in order to discourage validators from sending zero-cost transactions in the blocks they generate.
```
gkoumout, line 47: At this stage of the text, perhaps it is not so clear why this discourages validators. Perhaps it would be good to add a sentence explaining this. 

Example: 

As introduced in [LIP 0013][lip-0013], all transactions must have a transaction fee greater or equal to a *minimum fee*. This is always deducted from the part of the fee which is given to the block generator. This way, the minimum fee (if non zero) is designed to discourage validators from sending zero-cost transactions in the blocks they generate.

proposals/lip-0048.md
gkoumout: Suggested structure (unpolished):

Subsection: Transaction Fees.

In the lisk ecosystem, as in almost all blockchains, each transaction has an associated transaction fee (link transaction schema LIP 68). Transaction fees prevent spamming a blockchain with unlimited transactions at no cost. Moreover, part of the transaction fees is provided as reward to validators who run and maintain the security of the network, as we explain in detail below.  [this part makes clear *why* we have transaction fees].

[Then add the sub-subsections of FeeTokenID and Minimum Fee Per Transaction, which are very well written IMO. ]

Subsection: Initialization fees ( we could even call them state growth fees or something else)

The fee structure presented here allows for defining a sufficient cost to manage state growth. In principle, any transaction resulting in a state increase should imply a fee.  This reasoning is implemented in the [`payFee`](#payfee) method, which can be invoked by other modules to charge an additional fee if during the execution of a transaction, a state growth is trigerred (e.g., new validator register in PoS module, new sidechain gets registered, or initialize a new user account).

[then discuss what happens after trs execution]

We call the part of the fee used during the execution of a transaction (i.e., the sum of minFee plus the initialization fees) as the "consumed" part of the fee;  the rest of the fee is called available fee and it is given as reward to the block generator, as an additional reward for securing the network. This way, it is expected that whenever the network is congested, validators prioritize transactions with higher available fee. 

[Add here the Fee pool paragraph]

[Add the ccm paragraph]

proposals/lip-0048.md
```diff
@@ -241,10 +252,11 @@ def verify(trs: Transaction) -> None:
     b = block including trs
     h = b.header.height
     if h < MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE:
-	    # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
-        MIN_FEE_PER_BYTE = 0
-    trsSize = length(encodeTransaction(trs))
-    minFee = MIN_FEE_PER_BYTE * trsSize
+	    # Set minFee = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```
gkoumout, line 255: indentation is off.

proposals/lip-0048.md
```diff
@@ -241,10 +252,11 @@ def verify(trs: Transaction) -> None:
     b = block including trs
```
gkoumout, line 252: Comment about line 249: Perhaps add a sentence or two explaining this? In particular, that this module runs this step before *every* transaction submitted in the chain (perhaps link to LIP55 defining the steps..)

gkoumout (changes requested):
Looks good to me, I have mostly writing style comments (which can be taken in consideration at your own discretion) and an indentation typo.

proposals/lip-0048.md
mjerkov: I don't know what is happening with my VSC, I'll commit from this suggestion:
```suggestion
    # Set minFee = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```

mjerkov: Second try:
```suggestion
        # Set minFee = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```

proposals/lip-0048.md
```diff
@@ -222,10 +252,11 @@ def verify(trs: Transaction) -> None:
...
     h = b.header.height
     if h < MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE:
-	    # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```
mjerkov, line 225:
```suggestion
	            # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```

proposals/lip-0048.md
```diff
@@ -222,10 +252,11 @@ def verify(trs: Transaction) -> None:
...
-        MIN_FEE_PER_BYTE = 0
-    trsSize = length(encodeTransaction(trs))
-    minFee = MIN_FEE_PER_BYTE * trsSize
+    # Set minFee = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```
mjerkov, line 255:
```suggestion
	                    # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
```

proposals/lip-0048.md
```diff
@@ -222,10 +252,12 @@ def verify(trs: Transaction) -> None:
     b = block including trs
     h = b.header.height
     if h < MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE:
-	    # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
-        MIN_FEE_PER_BYTE = 0
-    trsSize = length(encodeTransaction(trs))
-    minFee = MIN_FEE_PER_BYTE * trsSize
+                                    # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
+	                            minFee = 0
+        minFee = 0
```
mjerkov, line 257:
```suggestion
        # Set MIN_FEE_PER_BYTE = 0 for the first MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE blocks.
        minFee = 0
```

proposals/lip-0048.md
```diff
@@ -25,32 +25,43 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 ```python
 trsSize = length(encodeTransaction(trs))
 minFee = MIN_FEE_PER_BYTE * trsSize
 ```
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, which can be even set to zero. The value `MIN_FEE_PER_BYTE` can not be modified, unless a hard fork occurs. For this reason, we allow sidechains the possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block for which the minimum fee does not apply. 
 
-Note that chains can freely configure their `MIN_FEE_PER_BYTE`, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when it is set to zero.
+This could be useful, for example, for sidechains that want to use LSK as the fee token: Initially, no LSK is present on the sidechain, so the sidechain would not be able to process any transactions unless `MIN_FEE_PER_BYTE` is set to zero. By allowing an initial exception period, users have a time window to transfer LSK tokens to this sidechain, after which the initially set value of `MIN_FEE_PER_BYTE` would start being applied.
```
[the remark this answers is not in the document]
mjerkov: Noted.

proposals/lip-0048.md
mjerkov: Agreed, this makes it more clear. The nuances...

proposals/lip-0048.md
mjerkov: Yes, I also noticed that starting with "initialization" fees (which are, I agree, not only that, so the naming is not perfect) is a bit awkward, but I had no idea where to introduce them so I thought the beginning of the Rationale section is OK. 
Thanks for the suggestions, it's better this way since transaction fees are discussed first, which is a more natural thing to do.

proposals/lip-0048.md
ricott1: Maybe I would avoid mentioning state growth altogether. Transactions should pay a fee for several reasons, e.g. state growth, computational effort from the node, bandwidth, block space, etc. What I would clearly mention is that we do not have an automatic way of calculating the fee of a transaction for each one of these costs. Instead, we only calculate the fee for the block space automatically (similarly to bitcoin if you want) with the min fee, and leave flexibility to developers to assign a fee for the other cases "by hand". For instance, in the Token module we assign by hand an initialization fee which covers for the state growth. In principle, we could also assign a fee each time an expensive calculation is required (e.g. verifying an inclusion proof).

mjerkov: I tried to summarize your idea, please check the latest commit.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
 
 ## Rationale
 
-### Fee Token ID
+Fees for blockchain transactions are necessary for a variety of reasons, including state growth, node computation, and block space utilization.
 
-Each chain can configure the token used to pay fees. On the Lisk mainchain, the token used for transaction fees is the LSK token.
+The current implementation lacks an automated method for calculating fees in all of these cases. Instead, we automate the calculation of block space fees using a *minimum fee* approach, similar to Bitcoin's mechanism, and allow developers discretion to use the [`payFee`](#payfee) method for fee assignment in other situations, e.g., when transaction implies state growth or more significant computational effort required by a node.
```
ricott1, line 30: Just to clarify: This is similar to Bitcoin in the sense that the only relevant metric for the calculation of the min fee is the transaction size, but bitcoin does not have a min fee (that I know of). For simplicity I would drop the 'similar to Bitcoin's mechanism' sentence.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
 
 ## Rationale
 
-### Fee Token ID
+Fees for blockchain transactions are necessary for a variety of reasons, including state growth, node computation, and block space utilization.
```
ricott1, line 28:
```suggestion
Fees for blockchain transactions are necessary for a variety of reasons, including paying for state growth, node computation, and block space utilization.
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 
-### Minimum Fee per Transaction
+### Transaction Fees
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when the minimum fee per byte is set to zero.
+Each transaction in the Lisk ecosystem has a *transaction fee* associated with it, as defined in [LIP 0068][lip-0068]. Transaction fees prevent spamming a blockchain with unlimited transactions at zero cost. In addition, a part of the transaction fee is provided as a reward to the validators who operate the network and maintain its security, as we explain in detail below.
```
ricott1, line 34:
```suggestion
Each transaction in the Lisk ecosystem has a *transaction fee* associated with it, the `fee` property in the transaction schema in [LIP 0068][lip-0068]. Transaction fees prevent spamming a blockchain with unlimited transactions at zero cost. Part of the transaction fee is burned or transferred to a common pool. The remaining part is given as a reward to the validator who generated the block that included the transaction. 
```

ricott1: Also, I would move this paragraph to the very top and remove this section.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+Each transaction in the Lisk ecosystem has a *transaction fee* associated with it, as defined in [LIP 0068][lip-0068]. Transaction fees prevent spamming a blockchain with unlimited transactions at zero cost. In addition, a part of the transaction fee is provided as a reward to the validators who operate the network and maintain its security, as we explain in detail below. 
 
-Chains can choose that this minimum fee gets either burned (to avoid that validators can send transactions in the blocks they generate without cost) or transferred to the specified pool address `ADDRESS_FEE_POOL`. The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants. In the Lisk mainchain, the minimum fee is always burned.
+#### Fee Token ID
```
ricott1, line 36: I would move this section down, after the minimum Fee section

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+### Method `payFee`
+
+In principle, any transaction that results in a state increase or significant computational effort could require a fee. As mentioned above, since current implementation lacks an automated way of charging fees in these cases, we introduce the [`payFee`](#payfee) method. 
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a user account is initialized in the Token module. Also, developers may decide to call the `payFee` method in transactions requiring an extensive computation, e.g., in verifying an inclusion proof.
+
+For example, this method is called during the execution of the [validator registration command][lip-0057#execution] in LIP 0057:
```
ricott1, line 59: Maybe I would cite the token account initialization example rather than the validator one. The point is that in reality the validator and chain registration pay a way higher fees because you are kinda paying for the 'name space' rather than the state space. I would just avoid this discussion

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a user account is initialized in the Token module. Also, developers may decide to call the `payFee` method in transactions requiring an extensive computation, e.g., in verifying an inclusion proof.
+
+For example, this method is called during the execution of the [validator registration command][lip-0057#execution] in LIP 0057: 
+```python
+Fee.payFee(VALIDATOR_REGISTRATION_FEE)
+```
+
+### Fee Processing
+
+A part of the fee equal to the sum of the minimum fee and, if applicable, the initialization fee, is referred to as the *consumed fee*. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address.
```
ricott1, line 66: I think here we should be more precise. This 'consumed fee' would be minimum fee plus other fees paid using the `payFee` method, which could be more general than initialization fees.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+For example, this method is called during the execution of the [validator registration command][lip-0057#execution] in LIP 0057: 
+```python
+Fee.payFee(VALIDATOR_REGISTRATION_FEE)
+```
+
+### Fee Processing
+
+A part of the fee equal to the sum of the minimum fee and, if applicable, the initialization fee, is referred to as the *consumed fee*. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address. 
+
+The remainder of the fee, referred to as the *available fee*, is given to the block generator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee  in the case of network congestion.
```
ricott1, line 68:
```suggestion
The remainder of the fee, referred to as the *available fee*, is given to the block generator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in the case of network congestion.
```

ricott1: Also, we already said this I think. Maybe we could put this sentence together with the paragraph in the current "### Transaction Fees" section and merge them together. Than we can remove this section and directly discuss the fee pool.

proposals/lip-0048.md
gkoumout: Agree, this should be merged with the very first paragraph of the rationale.

gkoumout: Not sure we should add the burn/pool topic here. The main point of this paragraph is to explain *why* we need fees, and one of the reasons is that rewards validators. I would then keep it as is and just merge it with the very first sentence. Meaning, we just motivate the use of transaction fees.

proposals/lip-0048.md
gkoumout: Indeed, this is not important; we could put it in the very end of rationale for the time being, and after getting the final structure, put it in the appropriate place.

proposals/lip-0048.md
gkoumout: Also, I would not start the paragraph by saying that we lack some feature. We could just outline the two main topics (minimum fee based on blockspace) and extra command fees and say that they are explained below.

gkoumout:
```suggestion
The fee system proposed here, automates the calculation of block space fees using a *minimum fee* approach: the transaction fee should exceed a threshold which depends on the size of the transaction, as we explain below. Fees related to other reasons (e.g., state growth, computational effort) are determined by the corresponding parts of the protocol (e.g., extra fee to register a validator in the PoS module); the fee module takes care of the handling of such fees, via the [`payFee`](#payfee) method.
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 
-Chains can choose that this minimum fee gets either burned (to avoid that validators can send transactions in the blocks they generate without cost) or transferred to the specified pool address `ADDRESS_FEE_POOL`. The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants. In the Lisk mainchain, the minimum fee is always burned.
+#### Fee Token ID
+
+Each chain can configure the token used to pay fees, as defined by `TOKEN_ID_FEE`. On the Lisk mainchain, the token used for transaction fees is the LSK token.
+
+#### Minimum Fee per Transaction
+
+As introduced in [LIP 0013][lip-0013], all transactions must have a transaction fee that is greater than or equal to a *minimum fee* that is always deducted from the part of the fee that goes to the block generator. The minimum fee, unless it is zero, is intended to discourage validators from sending zero-cost transactions in the blocks they generate.
```
gkoumout, line 42: I think also here we are confusing things. The minFee is intended to charge for blockspace. What discourages validators from sending transactions is that they don't get the minFee back in case they send a transaction in the block they generate.

gkoumout:
```suggestion
As introduced in [LIP 0013][lip-0013], all transactions must have a transaction fee that is greater than or equal to a *minimum fee*. This fee corresponds to a cost for using the blockspace and therefore depends on the size of the transaction.
```

gkoumout: Then the discussion about burning the minFee could be moved to after defining this fee.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+For a transaction `trs`, the minimum fee is determined by its size and the proportionality constant `MIN_FEE_PER_BYTE`: 
+```python
+trsSize = length(encodeTransaction(trs))
+minFee = MIN_FEE_PER_BYTE * trsSize
+```
```
gkoumout, line 48: We can add here the discussion on removing minFee from rewards.

gkoumout, line 48:
```suggestion
```
The minimum fee is always deducted from the part of the fee that goes to the block generator, as we explain below. Therefore, the minimum fee, unless it is zero, discourages validators from sending zero-cost transactions in the blocks they generate.
```

File: proposals/lip-0048.md [hunk 26]
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+Note that chains can freely configure their `MIN_FEE_PER_BYTE`, which can even be set to zero. The `MIN_FEE_PER_BYTE` value cannot be changed, unless a hard fork occurs. For this reason, we allow sidechains to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block, for which the minimum fee does not apply. 
+
+This could, for example, be useful for sidechains that want to use LSK as the fee token. Initially, there is no LSK on the sidechain, so the sidechain cannot process transactions unless `MIN_FEE_PER_BYTE` is set to zero, which is generally not desirable since it essentially disables the feature of minimum fee for the chain. By granting an initial exemption period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks, users have a window of time to transfer LSK tokens to the sidechain, after which the initially set value of `MIN_FEE_PER_BYTE` would start being applied. 
+
+### Method `payFee`
```
gkoumout, line 53: I would still try to give a different title here, such as "additional fees", or "state growth fees". The payFee is just the function that does this.

proposals/lip-0048.md
gkoumout:
```suggestion
We call the part of the fee used during the execution of a transaction (i.e., the sum of minFee plus the fees handled by payFee) as the "consumed" part of the fee.
```

gkoumout: Namely, I would not discuss the burn vs pool here since it destructs the flow. 

1) Define consumed, available
2) Say that available goes to validator
3) Discuss the burn vs pool.

proposals/lip-0048.md
gkoumout: I think it is good to keep it, although it was also said before, in order to have a concrete paragraph describing this precisely.

proposals/lip-0048.md
mjerkov: Thanks for the suggestion, I changed the parahraph as suggested.

proposals/lip-0048.md
mjerkov: I went for the Greg's suggestion, maybe we don't need to mention the burning here, since this is moved to the Rationale introduction.

proposals/lip-0048.md
mjerkov: Agreed and changed as suggested.

proposals/lip-0048.md
mjerkov: I opted for leaving it here.

proposals/lip-0048.md
mjerkov: Yes, I agree that the minFee discussion is a bit unprecise; I've rewritten it according to your suggestion.

proposals/lip-0048.md
mjerkov: I renamed the paragraph "Additional Fees", I think it's a general term which doesn't say much, but covers everything :)

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+### Additional Fees
+
+In principle, any transaction that results in a state increase or significant computational effort could require a fee. Since the current implementation does not provide an automatic way to charge fees in these cases, we introduce the [`payFee`](#payfee) method. 
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a [user account is initialized][lip-0051#initializeuseraccountinternal] in the Token module: 
+```python
+Fee.payFee(USER_ACCOUNT_INITIALIZATION_FEE)
+```
+In addition, developers may decide to call the `payFee` method for transactions that require extensive computation, such as when verifying an inclusion proof.
```
sergeyshemyakov, line 60: I would add something like: "..., or if any other resource is extensively used during transaction execution" to be slightly more general. Maybe a transaction accesses a decentralized storage or interacts with an oracle, etc.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+### Fee Processing
+
+We refer to the total of the minimum fee and the fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" fee, is given to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address. 
+
+#### Fee Pool
+
+The option to transfer the consumed fees to a dedicated pool address `ADDRESS_FEE_POOL` could be beneficial for sidechains, as fees collected this way can be used as an incentive for validators and other protocol participants for various purposes: Funding a community pool for on-chain governance spending, sharing funds proportionally among validators per round, incentivizing ecosystem developers, supporting chain maintenance and improvement, rewarding users' staking and voting activities, etc.
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that counteracts inflation.
```
sergeyshemyakov, line 70: I think "counteracts" is a strong word, it sounds like the burnt fees per block are comparable with minted rewards. I would say something like: "works against inflation".

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+### Cross-chain Message Fees
+
+Cross-chain messages, or CCMs, facilitate interoperability by enabling cross-chain transfer of data. They require an associated fee, distinct from the usual transaction fee. Similar to transaction fees, the minimum fee can either be burned or transferred to a fee pool depending on the configuration of the chain, and the available fee is rewarded to a CCM relayer.
```
sergeyshemyakov, line 74: There is no minimum fee for CCMs, only what was consumed by `payFee`, so I suggest to change "the minimum fee" to "the consumed fee".

Also I would say that the fee goes to the CCM relayer in the second sentence. It would also be good to explain that this fee is supposed to cover CCM execution and incentivize relayers.

sergeyshemyakov (approved):
Generally looks good, I have a couple of suggestions.

proposals/lip-0048.md
mjerkov: Right. Changed as suggested.

proposals/lip-0048.md
mjerkov: Changed as suggested.

proposals/lip-0048.md
mjerkov: Changed as suggested.

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 ## Rationale
 
-### Fee Token ID
+### Transaction Fees
 
-Each chain can configure the token used to pay fees. On the Lisk mainchain, the token used for transaction fees is the LSK token.
+Blockchain transaction fees are instrumental for various reasons, including: 
+- Paying for state growth, node computation, and blockspace utilization. 
+- Preventing spamming a blockchain with unlimited zero-cost transactions. 
+- Incentivizing validators who operate the network and maintain its security.
```
gkoumout, line 33:
```suggestion
- Incentivizing validators who operate the network and maintain its security, by providing them part of the collected fees.
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 
-### Minimum Fee per Transaction
+Consequently, each transaction in the Lisk ecosystem has a *transaction fee* associated with it, as defined by the `fee` property of the transaction schema in [LIP 0068][lip-0068].  
 
-As introduced in [LIP 0013][lip-0013], all transactions must have a fee greater or equal to a minimum fee (which can be zero). The minimum fee is computed from the transaction size. Chains can configure their minimum fee per byte, with a possibility to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block when the minimum fee per byte is set to zero.
+The fee system proposed here automates the calculation of blockspace fees using a *minimum fee* approach: The transaction fee should exceed a threshold that depends on the size of the transaction. Fees associated to other factors, such as state growth or computational effort, are determined by the corresponding parts of the protocol, e.g., fee for registering a validator in the PoS module. The fee module manages these *additional* fees through the [`payFee`](#payfee) method.
```
gkoumout, line 37:
```suggestion
The fee system proposed here automates the calculation of blockspace fees using a *minimum fee* approach: The transaction fee should be greater than or equal to a threshold that depends on the size of the transaction. Fees associated to other factors, such as state growth or computational effort, are determined by the corresponding parts of the protocol, e.g., fee for registering a validator in the PoS module. The fee module manages these *additional* fees through the [`payFee`](#payfee) method.
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
 
-Chains can choose that this minimum fee gets either burned (to avoid that validators can send transactions in the blocks they generate without cost) or transferred to the specified pool address `ADDRESS_FEE_POOL`. The latter choice could be beneficial for sidechains, as fees collected in this way can be used as an incentive for validators and other protocol participants. In the Lisk mainchain, the minimum fee is always burned.
+#### Minimum Fee per Transaction
+
+As introduced in [LIP 0013][lip-0013], all transactions must have a transaction fee that is at least equal to a *minimum fee*. This fee corresponds to the cost of using blockspace and is therefore dependent on the size of the transaction, with the proportionality constant of `MIN_FEE_PER_BYTE`: 
+```python
+trsSize = length(encodeTransaction(trs))
+minFee = MIN_FEE_PER_BYTE * trsSize
+```
+The minimum fee is always subtracted from the part of the fee that goes to the block generator, as we explain below. Therefore, the minimum fee discourages validators from sending zero-cost transactions in the blocks they generate.
```
gkoumout, line 46:
```suggestion
The minimum fee is always subtracted from the part of the fee that goes to the block generator, as we explain below. This way, block generators do not have the option to include zero-cost transactions in the blocks they generate, unless minimum fee is set to zero.
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a [user account is initialized][lip-0051#initializeuseraccountinternal] in the Token module: 
+```python
+Fee.payFee(USER_ACCOUNT_INITIALIZATION_FEE)
+```
+In addition, developers may decide to call the `payFee` method for transactions that require extensive computation, e.g., when verifying an inclusion proof, or if any other resource is extensively used during transaction execution.
+
+### Fee Processing
+
+We refer to the total of the minimum fee and the fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" fee, is given to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address.
```
gkoumout, line 64:
```suggestion
We refer to the sum of the minimum fee and the additional fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" fee, is given to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address. 
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+The option to transfer the consumed fees to a dedicated pool address `ADDRESS_FEE_POOL` could be beneficial for sidechains, as fees collected this way can be used as an incentive for validators and other protocol participants for various purposes: Funding a community pool for on-chain governance spending, sharing funds proportionally among validators per round, incentivizing ecosystem developers, supporting chain maintenance and improvement, rewarding users' staking and voting activities, etc.
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that works against inflation.
+
+### Cross-chain Message Fees
+
+Cross-chain messages, or CCMs, facilitate interoperability by enabling cross-chain transfer of data. They require an associated fee, distinct from the usual transaction fee. Similar to transaction fees, the consumed fee can either be burned or transferred to a fee pool depending on the configuration of the chain. The available fee is rewarded to a CCM relayer to cover the cost of CCM execution and to incentivize participation in CCM relaying.
+
+#### Fee Token ID
```
gkoumout, line 76: Now that we found the right structure, I think this could be added in the initial presentation of trs fees, for example after the link to LIP 68 (meaning adding this sentence there instead of having a separate sub-sub-section for it)

proposals/lip-0048.md
gkoumout: Minor point: I meant to add it as text between subsection title and pseudocode. THis way you could put also a link to LIP 55).

proposals/lip-0048.md
```diff
@@ -25,15 +25,57 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+We refer to the total of the minimum fee and the fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" fee, is given to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address. 
+
+#### Fee Pool
+
+The option to transfer the consumed fees to a dedicated pool address `ADDRESS_FEE_POOL` could be beneficial for sidechains, as fees collected this way can be used as an incentive for validators and other protocol participants for various purposes: Funding a community pool for on-chain governance spending, sharing funds proportionally among validators per round, incentivizing ecosystem developers, supporting chain maintenance and improvement, rewarding users' staking and voting activities, etc.
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that works against inflation.
+
+### Cross-chain Message Fees
```
gkoumout, line 72: I think now this is the only paragraph that perhaps needs to be extended content-wise. We can discuss on this.

proposals/lip-0048.md
mjerkov: Thanks. I would only change it to:
```suggestion
- Incentivizing validators who operate the network and maintain its security, by providing them with a part of collected fees.
```

proposals/lip-0048.md
mjerkov: Done.

sergeyshemyakov (approved):
I think this is a good balance of mentioning the important details and not being too precise.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+Since managing message fees requires checking and updating the escrow balances for the message fee token, this task is assigned to the Token module. In particular, paying for message fees is called for in the Interoperability module, whose [`sendInternal`][lip-0045#sendinternal] method uses the [`payMessageFee`][lip-0051#paymessagefee] of the Token module.
+
+Note that, for protocol reasons (see, e.g., [`bounce`][lip-0045#bounce] and [`terminateChain`][lip-0045#terminatechain] methods), Interoperability module allows for message fees to be zero. Consequently, setting a minimum message fee should be delegated to each module that provides its own cross-chain commands. 
+
+Similar to transaction fees, the consumed part of the message fee can either be burned or transferred to a fee pool, depending on the chain configuration. The available part of the message fee is left to the relayer in order to cover the cost of CCM execution and to incentivize its participation in CCM relaying.
```
gkoumout, line 80: Highlight that here consumed fee is only the so-called "additional fees", since there is no minFee.

gkoumout (changes requested):
I think it is in a very good shape, just made few more comments. Almost there!

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+Note that chains can freely configure `MIN_FEE_PER_BYTE`, which can even be set to zero. The `MIN_FEE_PER_BYTE` value cannot be changed, unless a hard fork occurs. For this reason, we allow sidechains to define a period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks after the genesis block, for which the minimum fee does not apply. 
+
+This could, for example, be beneficial for sidechains that want to use LSK as the fee token. Initially, there is no LSK on the sidechain, so the sidechain cannot process transactions unless `MIN_FEE_PER_BYTE` is set to zero, which is generally not desirable since it essentially disables the feature of minimum fee for the chain. By granting an initial exemption period of `MAX_BLOCK_HEIGHT_ZERO_FEE_PER_BYTE` blocks, users have a window of time to transfer LSK tokens to the sidechain, after which the initially set value of `MIN_FEE_PER_BYTE` would start being applied. 
+
+### Additional Fees
```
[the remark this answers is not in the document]
mjerkov: Agreed.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+```python
+Fee.payFee(USER_ACCOUNT_INITIALIZATION_FEE)
+```
+In addition, developers may decide to call the `payFee` method for transactions that require extensive computation, e.g., when verifying an inclusion proof, or if any other resource is extensively used during transaction execution.
+
+### Fee Processing
+
+We refer to the sum of the minimum fee and the additional fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" part of the fee, is transferred to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address. 
+
+#### Fee Pool
```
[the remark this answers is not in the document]
mjerkov: Yes, seems like a good idea to remove the title, the flow will be better.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that works against inflation.
+
+### Cross-chain Message Fees
+
+[Cross-chain messages][lip-0049#ccmschema] (CCMs) facilitate interoperability by enabling cross-chain transfer of data. Each CCM contains a *message fee*, intended to cover the cost of transaction processing on the receiving chain. 
+
+Since managing message fees requires checking and updating the escrow balances for the message fee token, this task is assigned to the Token module. In particular, paying for message fees is called for in the Interoperability module, whose [`sendInternal`][lip-0045#sendinternal] method uses the [`payMessageFee`][lip-0051#paymessagefee] of the Token module.
+
+Note that, for protocol reasons (see, e.g., [`bounce`][lip-0045#bounce] and [`terminateChain`][lip-0045#terminatechain] methods), Interoperability module allows for message fees to be zero. Consequently, setting a minimum message fee should be delegated to each module that provides its own cross-chain commands.
```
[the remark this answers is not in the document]
mjerkov: Thanks for these suggestions, I will include them.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+In principle, any transaction that results in a state increase or significant computational effort could require a fee. Since the current implementation does not provide an automatic way to charge fees in these cases, we introduce the [`payFee`](#payfee) method. 
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a [user account is initialized][lip-0051#initializeuseraccountinternal] in the Token module: 
+```python
+Fee.payFee(USER_ACCOUNT_INITIALIZATION_FEE)
+```
+In addition, developers may decide to call the `payFee` method for transactions that require extensive computation, e.g., when verifying an inclusion proof, or if any other resource is extensively used during transaction execution.
+
+### Fee Processing
```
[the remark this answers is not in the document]
gkoumout: I would still suggest to have it as a sub-sub section.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+#### Additional Fees
+
+In principle, any transaction that results in a state increase or significant computational effort could require a fee. Since the current implementation does not provide an automatic way to charge fees in these cases, we introduce the [`payFee`](#payfee) method. 
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a [user account is initialized][lip-0051#initializeuseraccountinternal] in the Token module:
```
gkoumout, line 56: minor: perhaps add links to validator registration or sidechain registration executions so that a reader can navigate easily and get the point.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+This method can be invoked by other modules to charge an additional fee when a state increase is triggered during the execution of a transaction, e.g., when a new validator is registered in the PoS module, a new sidechain is registered, or a [user account is initialized][lip-0051#initializeuseraccountinternal] in the Token module: 
+```python
+Fee.payFee(USER_ACCOUNT_INITIALIZATION_FEE)
+```
+In addition, developers may decide to call the `payFee` method for transactions that require extensive computation, e.g., when verifying an inclusion proof, or if any other resource is extensively used during transaction execution.
+
+### Fee Processing
+
+We refer to the sum of the minimum fee and the additional fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" part of the fee, is transferred to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fee is either burned or transferred to a dedicated pool address.
```
gkoumout, line 64:
```suggestion
We refer to the sum of the minimum fee and the additional fees processed by the `payFee` method as the "consumed" part of the fee. The remainder of the fee, called the "available" part of the fee, is transferred to the validator as a reward. In this way, validators are assumed to prioritize transactions with a higher available fee in case of network congestion. [After the transaction is executed](#after-command-execution), consumed fees are either burned or transferred to a dedicated pool address. 
```

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that works against inflation.
+
+### Cross-chain Message Fees
+
+Cross-chain messages (CCMs) facilitate interoperability by enabling cross-chain transfer of data. Each CCM contains a [*message fee*][lip-0049#ccmschema], intended to cover the cost of transaction processing on the receiving chain (see [LIP 0049][lip-0049#fee]). 
+
+Since managing message fees requires checking and updating the escrow balances for the message fee token, this task is assigned to the Token module. In particular, paying for message fees is called for in the Interoperability module, whose [`sendInternal`][lip-0045#sendinternal] method uses the [`payMessageFee`][lip-0051#paymessagefee] of the Token module. After the Token module has ensured that the message fee has been credited to the relayer's account, the Fee module handles the transaction fee processing in the receiving chain.
```
gkoumout, line 74: Looks very good. One small point: I would explicitly explain that the ccm fee management includes two chains. 

e.g.., "paying for message fees *in the sending chain*", or " the ccm sender has to pay the fee in the sending chain". Then in the receiving chain, the token module assigns the fee to the relayer account and updates escrows if necessary.

proposals/lip-0048.md
```diff
@@ -25,15 +25,59 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+Furthermore, the CCM sender should ensure to include any additional fees necessary for the CCM execution, e.g., to pay for the user account initialization on the receiving chain. 
+
+Similar to transaction fees, the consumed part of the message fee, which might only be additional fees (as there is no mandatory minimum fee), can either be burned or transferred to a fee pool, depending on the chain configuration. The available part of the message fee is left to the relayer to cover the cost of CCM execution and to incentivize its participation in CCM relaying. However, in contrast to including transactions in a block, a relayer cannot prioritize CCMs with higher fees as the interoperability protocol requires them to be sent in order. This also implies that users have no reason to pay higher message fees than are required for successful CCM execution.
```
gkoumout, line 80:
```suggestion
The fee processing is similar to transaction fees; note that the consumed part of the message fee contains only the so-called additional fees,  as there is no mandatory minimum fee. Similarly to transaction fees, the consumed part can either be burned or transferred to a fee pool, depending on the chain configuration. The available part of the message fee is left to the relayer to cover the cost of CCU execution (note the relayer pays transaction fees for posting the CCU in the receiving chain, since a CCU is a regular transaction) and to incentivize its participation in CCM relaying. However, in contrast to including transactions in a block, a relayer cannot prioritize CCMs with higher fees as the interoperability protocol requires them to be sent in order. This also implies that users have no reason to pay higher message fees than are required for successful CCM execution.
```

gkoumout: Also in the very end, we could even say that (because users no need to pay more fee to include their ccm) a consequence of this is that users know that the ccms would be delivered as soon as a relayer decides to post a ccm in the target chain, avoiding this way extra delays on delivering the ccm, something like that.

proposals/lip-0048.md
gkoumout: Also I am not sure that so much detail of mentioning sendInternal and payMessage fee is needed, but not strong opinion.

gkoumout (commented):
Well done! Looks very good now. 

I just have some last minor comments/suggestions. Treat them as optional. Last round!

proposals/lip-0048.md
mjerkov: Done.

proposals/lip-0048.md
mjerkov: Thanks for the suggestion, I've rewritten that part.

proposals/lip-0048.md
```diff
@@ -25,15 +25,65 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+In the Lisk mainchain, the consumed fees are always burned. This reduces the total supply, creating a deflationary pressure that works against inflation.
+
+### Cross-chain Message Fees
+
+Cross-chain messages (CCMs) facilitate interoperability by enabling cross-chain transfer of data. Each CCM contains a [*message fee*][lip-0049#ccmschema], intended to cover the cost of transaction processing on the receiving chain (see [LIP 0049][lip-0049#fee]). 
+
+Since managing message fees requires checking and updating the escrow balances for the message fee token, this task is assigned to the Token module whose [`payMessageFee`][lip-0051#paymessagefee] method charges message fees in the sending chain. In the receiving chain, the Token module assigns the message fee to the relayer account and updates escrows if necessary, and then the Fee module handles the transaction fee processing as described above. Note that the CCM sender should include any additional fees necessary for the CCM execution, e.g., to pay for the user account initialization on the receiving chain.
```
gkoumout, line 74: The last sentence of this paragraph, could be moved to the end of the next paragraph IMO

proposals/lip-0048.md
```diff
@@ -25,15 +25,65 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+#### Minimum Fee per CCM
+
+For protocol reasons (see, e.g., [`bounce`][lip-0045#bounce] and [`terminateChain`][lip-0045#terminatechain] methods in LIP 0045), the Interoperability module allows the message fee to be zero. Consequently, the choice of setting a minimum message fee is delegated to each module that provides its own cross-chain commands. Nevertheless, there is a minimum fee in case the user wants the failed CCM to be returned (see [`bounce`][lip-0045#bounce] method in LIP 0045). Thus, even though this fee is not mandatory, it is advisable to treat it as a suggested minimum fee. 
+
+#### Message Fee Processing
+
+Processing of message fees is similar to that of transaction fees. It should be noted that the consumed part of the message fee contains only the so-called additional fees, as there is no mandatory minimum fee. The consumed part can either be burned or transferred to a fee pool, depending on the chain configuration.
```
gkoumout, line 82:
```suggestion
Processing of message fees is similar to that of transaction fees. The only essential difference is that the consumed part of the message fee contains only the so-called additional fees, as there is no mandatory minimum fee. The consumed part is handled in the same way as for transactions (i.e., either be burned or transferred to a fee pool, depending on the chain configuration) . 
```

gkoumout: This just to make clear that consumed fee processing is the same for transactions and ccms, for example it is not possible to burn the transaction fee and send the ccm fee to a pool.

proposals/lip-0048.md
```diff
@@ -25,15 +25,65 @@ This LIP defines the fee system in a modular way, as currently [used in the Lisk
...
+
+#### Message Fee Processing
+
+Processing of message fees is similar to that of transaction fees. It should be noted that the consumed part of the message fee contains only the so-called additional fees, as there is no mandatory minimum fee. The consumed part can either be burned or transferred to a fee pool, depending on the chain configuration. 
+
+The available part of the message fee is left to the relayer to cover the cost of CCU execution and to incentivize its participation in CCM relaying. Also, since a CCU is a regular transaction, note that the relayer pays transaction fees for posting the CCU in the receiving chain.
```
gkoumout, line 84:
```suggestion
The available part of the message fee is left to the relayer to cover the cost of CCU execution (recall that a CCU is a regular transaction, therefore the relayer has to pay transaction fees for posting the CCU in the receiving chain) and to incentivize its participation in CCM relaying. 
```

gkoumout: Suggest this change because the last sentence seemed out of context compared to the first one.

gkoumout (commented):
Very nice! Few more suggestions, mostly for re-arranging few sentences.

proposals/lip-0048.md
mjerkov: Yes, I noticed it doesn't fit so well here. Thanks for the suggestion.

gkoumout: You forgot to remove it from here, you just pasted in the next paragraph ;-)

gkoumout, line 74:
```suggestion
Since managing message fees requires checking and updating the escrow balances for the message fee token, this task is assigned to the Token module whose [`payMessageFee`][lip-0051#paymessagefee] method charges message fees in the sending chain. In the receiving chain, the Token module assigns the message fee to the relayer account and updates escrows if necessary, and then the Fee module handles the message fee processing in a similar way as it handles transaction fees.
```

gkoumout (changes requested):
Two small typos found. Otherwise, looks very good.

proposals/lip-0048.md
mjerkov: Oh. Thanks for noticing this

gkoumout (approved):
Well done!

janhack (approved):
Editorial approval

Issue #393, opened 2023-06-27
Cross-chain transfer of LIP 0051 does not verify if the receiving chain is active


mjerkov:
## Motivation
The requirement that the receiving chain is active is a necessary prerequisite for a successful cross-chain transfer.

## Current Specifications
Currently, neither [cross chain transfer command verification](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#verification-1) and [cross chain transfer method](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#transfercrosschain-2) of Token module LIP 0051 does not check whether the receiving chain is active, i.e. if it has the status `CHAIN_STATUS_ACTIVE` as defined in LIP 0045.

## Proposed Specifications
1. In [cross chain transfer command verification](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#verification-1), introduce the following:
```python
if not Interoperability.isChannelActive(receivingChainID):
    raise Exception("Receiving chain is not active.")
```
2. In  [cross chain transfer method](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#transfercrosschain-2), introduce the following:
```python
if not Interoperability.isChannelActive(receivingChainID):
    emitFailedCrossChainTransferEvent(senderAddress, tokenID, amount, receivingChainID, recipientAddress, RECEIVING_CHAIN_NOT_ACTIVE)
    raise Exception("Receiving chain is not active.")
```
and add `RECEIVING_CHAIN_NOT_ACTIVE` to the [table](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#event-names-and-results) of event result codes.

## Affected LIPs
LIP 0051

janhack:
We close the issue and do not use the `isChannelActive` method for the following reasons:

- Avoid using more internals of the Interoperability module, in particular, avoid setting a precedent for more complicated interactions between interoperable modules and the Interoperability without that being necessary (it is okay that the send method can fail).
- Keeping transaction verification minimal and not doing stateful checks that can be detected via dry-running.
- Relying on the UI to better address that the receiving chain is active as the UI can check the Lisk Mainchain state, which cannot be checked on-chain for sidechains.

PR #394, opened 2023-06-27
Add `isChannelActive` check to cross-chain transfer of LIP 0051

mjerkov:
Resolves Issue [#393](https://github.com/LiskHQ/lips/issues/393).

### DIFF
--- a/proposals/lip-0051.md
+++ b/proposals/lip-0051.md
@@ -194,6 +194,7 @@ We further use the utility function `getMainchainID()` defined in [LIP 0037][lip
 | `TOKEN_ID_NOT_NATIVE` | uint32 | 12 | Used when the specified token ID is not native to the chain. |
 | `INSUFFICIENT_ESCROW_BALANCE` | uint32 | 13 | Used when the escrowed account does not have sufficient balance. |
 | `INVALID_RECEIVING_CHAIN` | uint32 | 14 | Used when, during the cross-chain token transfer, the receiving chain is set to be equal to the sending chain. |
+| `RECEIVING_CHAIN_NOT_ACTIVE` | uint32 | 15 | Used when, during the cross-chain token transfer, the receiving chain is not active. |
 
 ### Type Definition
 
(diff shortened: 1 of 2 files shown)


proposals/lip-0051.md
```diff
@@ -511,6 +512,9 @@ def verify(trs: Transaction) -> None:
     if receivingChainID == OWN_CHAIN_ID:
         raise Exception("Receiving chain cannot be the sending chain.")
 
+    if not Interoperability.isChannelActive(receivingChainID):
```
ricott1, line 515: The previous check is superfluous now

ricott1 (changes requested):
One check is now superfluous

proposals/lip-0051.md
```diff
@@ -1921,6 +1925,10 @@ def transferCrossChain(
         emitFailedCrossChainTransferEvent(senderAddress, tokenID, amount, receivingChainID, recipientAddress, INVALID_RECEIVING_CHAIN)
```
gkoumout, line 1925: This check should also be deleted now (and the `INVALID_RECEIVING_CHAIN` variable, since it would not be used anymore..)

gkoumout (changes requested):
Looks good to me, just one small comment.

proposals/lip-0051.md
mjerkov: As discussed, it's not wrong to keep this check for reasons of clarity.

proposals/lip-0051.md
mjerkov: As discussed, it's not wrong to keep this check for reasons of clarity.

ricott1 (approved):

gkoumout (approved):

proposals/lip-0051.md
ishantiw, line 515: Just a note, since in mainchain we need `timestamp` for `isLive` function being called from `isChannelActive` function, we pass it as an argument to it in SDK, on mainchain we call `isChannelActive(receivingChainID, timestamp)` and on sidechain `isChannelActive(receivingChainID)` so in token module we make `timestamp` arg optional similar to `interop.send()` where also we need timestamp

ishantiw (approved):
Related SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8675

proposals/lip-0051.md
mjerkov: Thanks for the clarification of the SDK implementation.

janhack:
We do not make the additional check about the channel state to keep the protocol simple and as it can be detected via dry-running a transaction whether a channel is active.

PR #399, opened 2023-07-05
Verify sparse Merkle tree inclusion proofs in command exectutions 

### LINKED ISSUE #398: Verify sparse Merkle tree inclusion proofs during command execution
## Motivation

Disallow creation of invalid transaction consuming several node resources

## Current Specifications

There are several commands in LIP 0053 and LIP 0054 that perform a validation of a (sparse) Merkle tree inclusion proof in their verification step. This could lead to abuses where users send a very expensive (computational wise) trs to a node which will fail in the transaction pool (and hence pay no fee). Notice that the update of the root is indeed done in the execution. The list of commands is the following:

- [Cross-chain Update](https://github.com/LiskHQ/lips/blob/901f91aa0460198a9c11e91bb0ff3df31fc7e028/proposals/lip-0053.md?plain=1#L294)
- [State Recovery](https://github.com/LiskHQ/lips/blob/901f91aa0460198a9c11e91bb0ff3df31fc7e028/proposals/lip-0054.md?plain=1#L252)
- [State Recovery Initialization](https://github.com/LiskHQ/lips/blob/901f91aa0460198a9c11e91bb0ff3df31fc7e028/proposals/lip-0054.md?plain=1#L680-L683)
- [Message Recovery Initialization](https://github.com/LiskHQ/lips/blob/901f91aa0460198a9c11e91bb0ff3df31fc7e028/proposals/lip-0054.md?plain=1#L790)
- [Message Recovery](https://github.com/LiskHQ/lips/blob/901f91aa0460198a9c11e91bb0ff3df31fc7e028/proposals/lip-0054.md?plain=1#L406-L410)

## Proposed Specifications

Move the inclusion proof verification of the respective commands to their execution step. 

## Affected LIPs

0053, 0054

AndreasKendziorra:
This PR does the following:
- fixes #398 
- emit events in LIP 0054 for inclusion proof failures 
- add additional cheap check in verification of message recovery command to ensure that the indices of all provides messages are in the range `[0, ..., outboxSize -1]`
- improve checks (in terms of efficiency) in message recovery command to ensure uniqueness and increasing order  
- fix a bug in state recovery initialization command (it was accepting a chain account with status "registered" which should not be the case)
- add schema validation of sidechain account in state recovery initialization command
- fix schema validation of channel in message recovery initialization command

### DIFF
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -416,6 +416,10 @@ def beforeCrossChainMessagesExecution(ccu: CCU) -> None:
     verifyCertificateSignature(ccu)
 
     if ccu.params.inboxUpdate is not empty:
+        # This check is expensive. Therefore, it is done in the execute step instead of the verify
+        # step. Otherwise, a malicious relayer could spam the transaction pool with computationally
+        # costly CCU verifications without paying fees.
+        verifyPartnerChainOutboxRoot(ccu)
         # Initialize the relayer account for the message fee token.
         # This is necessary to ensure that the relayer can receive the CCM fees
         # If the account already exists, nothing is done.
@@ -637,12 +641,6 @@ def execute(trs: Transaction) -> None:
     ccu = trs
     ccu.params = decode(crossChainUpdateTransactionParams, trs.params)
 
-    if ccu.params.inboxUpdate is not empty:
-        # This check is expensive. Therefore, it is done in the execute step instead of the verify
-        # step. Otherwise, a malicious relayer could spam the transaction pool with computationally
-        # costly CCU verifications without paying fees.
-        verifyPartnerChainOutboxRoot(ccu)
-
     try:
         beforeCrossChainMessagesExecution(ccu)
     except:
@@ -709,12 +707,6 @@ _Figure 2: A schematic of the steps of a sidechain CCU execution. Some details a
 def execute(trs: Transaction) -> None:
     ccu = trs
     ccu.params = decode(crossChainUpdateTransactionParams, trs.params)
-
-    if ccu.params.inboxUpdate is not empty:
-        # This check is expensive. Therefore, it is done in the execute step instead of the verify
-        # step. Otherwise, a malicious relayer could spam the transaction pool with computationally
-        # costly CCU verifications without paying fees.
-        verifyPartnerChainOutboxRoot(ccu)
 
     try:
         beforeCrossChainMessagesExecution(ccu)
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -246,6 +246,25 @@ def calculateNewActiveValidators(activeValidators: list[ActiveValidator], blsKey
     return [v for v in newActiveValidators if v.bftWeight > 0]
 ```
 
+#### verifyOutboxRootWitness
+
+```python
+def verifyOutboxRootWitness(ccu: CCU) -> None:
+    outboxRootWitness = ccu.params.inboxUpdate.outboxRootWitness
+    # The outbox root witness properties must be set either both to their default values
+    # or both to a non-default value.
+    if outboxRootWitness.bitmap == EMPTY_BYTES and len(outboxRootWitness.siblingHashes) > 0:
+        raise Exception("The bitmap in the outbox root witness must be non-mepty if the sibling hashes are non-empty.")
+    if outboxRootWitness.bitmap != EMPTY_BYTES and len(outboxRootWitness.siblingHashes) == 0:
+        raise Exception("The sibling hashes in the outbox root witness must be non-mepty if the bitmap is non-empty.")
+
+    # The outbox root witness is empty if and only if the certificate is empty
+    if outboxRootWitness.bitmap == EMPTY_BYTES and len(ccu.params.certificate) > 0:
+        raise Exception("The outbox root witness must be non-empty to authenticate the new partnerChainOutboxRoot.")
+    if outboxRootWitness.bitmap != EMPTY_BYTES  and len(ccu.params.certificate) == 0:
+        raise Exception("The outbox root witness can be non-empty only if the certificate is non-empty.")
+```
+
 #### verifyPartnerChainOutboxRoot
 
 ```python
@@ -262,23 +281,9 @@ def verifyPartnerChainOutboxRoot(ccu: CCU) -> None:
     # because all messages have been included in the CCU.
     newInboxRoot = calculateRootFromRightWitness(inboxTree.size, inboxTree.appendPath, ccu.params.inboxUpdate.messageWitnessHashes)
 
-
-    outboxRootWitness = ccu.params.inboxUpdate.outboxRootWitness
-    # The outbox root witness properties must be set either both to their default values
-    # or both to a non-default value.
-    if outboxRootWitness.bitmap == EMPTY_BYTES and len(outboxRootWitness.siblingHashes) > 0:
-        raise Exception("The bitmap in the outbox root witness must be non-mepty if the sibling hashes are non-empty.")
-    if outboxRootWitness.bitmap != EMPTY_BYTES and len(outboxRootWitness.siblingHashes) == 0:
-        raise Exception("The sibling hashes in the outbox root witness must be non-mepty if the bitmap is non-empty.")
-
-    # The outbox root witness is empty if and only if the certificate is empty
-    if outboxRootWitness.bitmap == EMPTY_BYTES and len(ccu.params.certificate) > 0:
-        raise Exception("The outbox root witness must be non-empty to authenticate the new partnerChainOutboxRoot.")
-    if outboxRootWitness.bitmap != EMPTY_BYTES  and len(ccu.params.certificate) == 0:
-        raise Exception("The outbox root witness can be non-empty only if the certificate is non-empty.")
-
     # Verification for non-empty certificates.
     if len(ccu.params.certificate) > 0:
+        outboxRootWitness = ccu.params.inboxUpdate.outboxRootWitness
         outboxKey = STORE_PREFIX_INTEROPERABILITY + SUBSTORE_PREFIX_OUTBOX_ROOT + sha256(OWN_CHAIN_ID)
         proof = {
             siblingHashes: outboxRootWitness.siblingHashes,
@@ -330,6 +335,9 @@ def verifyCommon(ccu: CCU) -> None:
     validatorsUpdate = ccu.params.activeValidatorsUpdate
     if len(validatorsUpdate.blsKeysUpdate) > 0 or len(validatorsUpdate.bftWeightsUpdate) > 0 or validatorsUpdate.bftWeightsUpdateBitmap != EMPTY_BYTES or ccu.params.certificateThreshold != validators(ccu.params.sendingChainID).certificateThreshold:
         verifyValidatorsUpdate(ccu)
+
+    if ccu.params.inboxUpdate is not empty:
+        verifyOutboxRootWitness(ccu)
 ```
 
 #### updateValidators
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/introduce-cross-chain-update-mechani
 Status: Draft
 Type: Standards Track
 Created: 2021-05-22
-Updated: 2023-06-08
+Updated: 2023-07-12
 Requires: 0045, 0049, 0058, 0061
 ```
 
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -330,8 +330,6 @@ def verifyCommon(ccu: CCU) -> None:
     validatorsUpdate = ccu.params.activeValidatorsUpdate
     if len(validatorsUpdate.blsKeysUpdate) > 0 or len(validatorsUpdate.bftWeightsUpdate) > 0 or validatorsUpdate.bftWeightsUpdateBitmap != EMPTY_BYTES or ccu.params.certificateThreshold != validators(ccu.params.sendingChainID).certificateThreshold:
         verifyValidatorsUpdate(ccu)
-    if ccu.params.inboxUpdate is not empty:
-        verifyPartnerChainOutboxRoot(ccu)
 ```
 
 #### updateValidators
@@ -630,6 +628,13 @@ _Figure 1: A schematic of the steps of a mainchain CCU execution. Some details a
 def execute(trs: Transaction) -> None:
     ccu = trs
     ccu.params = decode(crossChainUpdateTransactionParams, trs.params)
+
+    if ccu.params.inboxUpdate is not empty:
+        # This check is expensive. Therefore, it is done in the execute step instead of the verify
+        # step. Otherwise, a malicious relayer could spam the transaction pool with computationally
+        # costly CCU verifications without paying fees.
+        verifyPartnerChainOutboxRoot(ccu)
+
     try:
         beforeCrossChainMessagesExecution(ccu)
     except:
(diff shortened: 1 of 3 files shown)


proposals/lip-0054.md
```diff
@@ -221,10 +221,7 @@ def verify(trs: Transaction) -> None:
     if recoveryModule does not have a recover function:
         raise Exception("Module is not recoverable.")
 
-    terminatedStateRoot = terminatedStateAccount(trsParams.chainID).stateRoot
-
     queryKeys = []
-    storeQueries = []
 
     # Calculate store prefix from the module name according to LIP 0040.
     storePrefix = module_name_to_store_prefix(trsParams.module)
```
ricott1, line 227: I think now we can remove this (and update the queryKey) since we do not check the inclusion proof here

proposals/lip-0054.md
```diff
@@ -665,6 +675,13 @@ def verify(trs: Transaction) -> None:
     if sidechainAccount.status != CHAIN_STATUS_TERMINATED
```
ricott1, line 675: Unrelated to this PR, but aren't we allowing to terminate sidechains which are in the registered status and did not activate yet?

ricott1: On line 673, we should also validate the sidechainAccount object against the schema

proposals/lip-0053.md
```diff
@@ -630,6 +628,13 @@ _Figure 1: A schematic of the steps of a mainchain CCU execution. Some details a
 def execute(trs: Transaction) -> None:
     ccu = trs
     ccu.params = decode(crossChainUpdateTransactionParams, trs.params)
+
+    if ccu.params.inboxUpdate is not empty:
+        # This check is expensive. Therefore, it is done in the execute step instead of the verify
+        # step. Otherwise, a malicious relayer could spam the transaction pool with computationally
+        # costly CCU verifications without paying fees.
+        verifyPartnerChainOutboxRoot(ccu)
```
ricott1, line 636: This works, but maybe is not optimal, since the SMT inclusion proof is not necessarily performed in `verifyPartnerChainOutboxRoot`. One option would be to refactor `verifyPartnerChainOutboxRoot` to not perform the expensive check but still the cheap ones, and move only the SMT verify here. @gkoumout Opinions?

ricott1 (changes requested):
Some unrelated comments on LIP54 that we can still fix, and one question for LIP53

proposals/lip-0054.md
ricott1: for aesthetics reasons you may want to keep this tho, up to you

proposals/lip-0053.md
gkoumout: Agree that this would be better, but it is fine for me to keep it as is for simplicity.

gkoumout: Although it does not seem to be a big edit: Just remove lines 280-295 from `verifyPartnerChainOutboxRoot` and paste it here; Note that the name `verifyPartnerChainOutboxRoot` would not make much sense anymore in this case.

proposals/lip-0053.md
```diff
@@ -630,6 +628,13 @@ _Figure 1: A schematic of the steps of a mainchain CCU execution. Some details a
...
+
     try:
         beforeCrossChainMessagesExecution(ccu)
```
gkoumout, line 639: One option would be also to include lines 631-637 to `beforeCrossChainMessagesExecution`, as a first step.

proposals/lip-0054.md
gkoumout: But then you need to initialiaze queryKey as an empty list, which would be almost the same, so I would not bother modifying this.

proposals/lip-0054.md
```diff
@@ -275,9 +285,9 @@ def execute(trs: Transaction) -> None:
...
             "bitmap": entry.bitmap
         }
-        storeQueries.append(query)
+        storeQueriesUpdate.append(query)
 
-    terminatedStateAccount(trsParams.chainID).stateRoot = SMTCalculateRoot(trsParams.siblingHashes, storeQueries)
+    terminatedStateAccount(trsParams.chainID).stateRoot = SMTCalculateRoot(trsParams.siblingHashes, storeQueriesUpdate)
 ```
 
 #### Recover Function
```
gkoumout, line 293: Shouldn't we also remove the inclusion proof of Message Recovery Command verification (lines 408-421)?

ricott1: I think these are regular Merkle inclusion proofs. In principle we should also remove those if it is possible to force a node to run the verification

proposals/lip-0053.md
AndreasKendziorra: moving lines 280-295 to execute also requires to move lines 253-263 to execute. And they are also slightly expensive I guess, so it's good to do it in the execute step (then we should also move 298-300).
The only remaining lines that could stay in the verify step are 266-278. Do you agree?

proposals/lip-0054.md
AndreasKendziorra: I agree that it is not needed here. So, I removed and also renamed `queryKeys` to `keys`.
@gkoumout I'm not sure what exactly you mean with "then you need to initialiaze queryKey as an empty list" because this already the case now.

proposals/lip-0054.md
AndreasKendziorra: > Unrelated to this PR, but aren't we allowing to terminate sidechains which are in the registered status and did not activate yet?

Yes, we do. In this case, the status is `CHAIN_STATUS_TERMINATED`. Hence, the condition is not fulfilled such that NO exception is raced, right?

> On line 673, we should also validate the sidechainAccount object against the schema, similar to what we do for message recovery initialization: validateObjectSchema(channelDataSchema, trsParams.channel)

I added the schema validation. I also added the decoding step to message recovery initialization which was missing there.

proposals/lip-0054.md
AndreasKendziorra: moved to execution

proposals/lip-0053.md
AndreasKendziorra: I actullay extracted the verifycation of outboxRootWitness into a new function `verifyOutboxRootWitness` and call this one in the verify step.

proposals/lip-0053.md
AndreasKendziorra: I kept now the function `verifyPartnerChainOutboxRoot` but just extracted the cheap verification steps into a new function `verifyOutboxRootWitness` which I call in the verify step.

proposals/lip-0054.md
```diff
@@ -221,50 +221,57 @@ def verify(trs: Transaction) -> None:
...
+    for entry in trsParams.storeEntries:
+        if entry.storeValue is EMPTY_BYTES:
+            raise Exception("Recovered store value cannot be empty.")
+
+        key = entry.substorePrefix + sha256(entry.storeKey)
+        keys.append(key)
+    
+    # Check that all keys are pariwise distinct, 
+    # meaning that we are not trying to recover the same entry twice.
+    if len(keys) != len(set(keys)):
```
ricott1, line 235: If we only check this, couldn't we run the same check on the params.storeEntries directly? I see no need to hash, for instance

proposals/lip-0054.md
AndreasKendziorra: >But i think we do not want to allow to terminate registered sidechains, as the liveness requirement should not be enforced yet.

I'm not 100%ly sure what you mean. Do you think there is something wrong in our rules how/when to terminate a sidechain, or in the verification of the State Recovery Initialization Command?

ricott1 (changes requested):
I left some further comments, I think we may have a bug in LIP54. Apart from this, I think we should emit persistent events for these failed verification now that we can.

proposals/lip-0054.md
ricott1: cbr

proposals/lip-0053.md
ricott1: cbr for me

proposals/lip-0054.md
AndreasKendziorra: Right. I removed the hashing and I also construct the set directly without constructing the array before.

AndreasKendziorra:
> I left some further comments, I think we may have a bug in LIP54. Apart from this, I think we should emit persistent events for these failed verification now that we can.

Did you have a specific event in mind, or should we create a new event type for it?

proposals/lip-0054.md
AndreasKendziorra: I fixed the check for the chain status

proposals/lip-0054.md
gkoumout: Ok this comment is outdated. I meant that if if `queryKeys` disappeared, then we would have to initialize the variable `queryKey` which was used inside the for-loop. 

Anyways, does not matter anymore. Current status looks good.

proposals/lip-0054.md
```diff
@@ -368,19 +392,21 @@ def verify(trs: Transaction) -> None:
...
         raise Exception("Inclusion proof indices and cross-chain messages do not have the same length.")
 
-    # Check that the idxs are sorted in ascending order.
-    if trsParams.idxs != sorted(trsParams.idxs):
-        raise Exception("Cross-chain message indexes are not sorted in ascending order.")
-
-    # Check that the idxs are unique.
-    if len(trsParams.idxs) != len(set(trsParams.idxs)):
-        raise Exception("Cross-chain message indexes are not unique.")
+    # Check that the idxs are strictly increasing
```
gkoumout, line 395: I like this, we check at once both sorted and uniqueness in linear time.

proposals/lip-0054.md
```diff
@@ -368,19 +392,21 @@ def verify(trs: Transaction) -> None:
...
     
-    # Check that the CCMs are still pending. We can check only the first one,
+    # Check that the CCMs are still pending. We check only the first one,
     # as the idxs are sorted in ascending order.
     if trsParams.idxs[0] < terminatedOutboxAccount(trsParams.chainID).partnerChainInboxSize:
         raise Exception("Cross-chain message is not pending.")
 
+    # Check that the CCM indices do not exceed the outbox size. We check only the last one,
+    # as the idxs are sorted in ascending order.
+    if terminatedOutboxAccount(trsParams.chainID).outboxSize <= trsParams.idxs[-1]:
```
gkoumout, line 407: Why is it `<=` and not just `<` ?

proposals/lip-0053.md
gkoumout: Looks correct to me. 

However I would still suggest to move lines 639 - 645 to `beforeCrossChainMessagesExecution`. I don't see any problem with that (perhaps I am missing something) and I think it would avoid repetition of this pseudocode and make the executions shorter and nicer.

gkoumout (commented):
Looks good to me. Left some minor suggestions and one question.

proposals/lip-0054.md
ricott1: I think it should be `<=` but please double check

proposals/lip-0054.md
```diff
@@ -120,6 +120,22 @@ def emitCCMEvent(ccm: CCM, result: uint32, code: uint32) -> None:
...
+def emitInvalidSMTVerificationEvent() -> None:
+    emitEvent(
+        module = MODULE_NAME_INTEROPERABILITY,
+        name = EVENT_NAME_INVALID_SMT_VERIFICATION,
+        data = {},
+        topic = []
+    )
+
+def emitInvalidRMTVerificationEvent() -> None:
+    emitEvent(
```
ricott1, line 133:
```suggestion
    emitPersistentEvent(
```

proposals/lip-0054.md
```diff
@@ -661,10 +688,20 @@ def verify(trs: Transaction) -> None:
...
     sidechainAccount = decode(chainDataSchema, trsParams.sidechainAccount)
-    # The command fails if the sidechain is not terminated and did not violate the liveness requirement.
-    if sidechainAccount.status != CHAIN_STATUS_TERMINATED
+    validateObjectSchema(chainDataSchema, sidechainAccount)
+    # The sidechain must either be terminated or must have violated the liveness limit while being active.
+    if sidechainAccount.status == CHAIN_STATUS_REGISTERED:
+        raise Exception("Sidechain is not terminated.")
+    if sidechainAccount.status == CHAIN_STATUS_ACTIVE
         and chainAccount(getMainchainID()).lastCertificate.timestamp - sidechainAccount.lastCertificate.timestamp <= LIVENESS_LIMIT:
         raise Exception("Sidechain is not terminated.")
```
ricott1, line 697: Maybe we can improve on these error messages now and distinguish the two cases

proposals/lip-0053.md
AndreasKendziorra: I see. It's indeed a bit cleaner. I moved it.

proposals/lip-0054.md
AndreasKendziorra: Example: Say `outboxSize = 10`. So, there are 10 messages in the outbox with the indices 0 to 9. If you try to recover a message with `index = outboxSize = 10`, then you try to recover the 11th message of the outbox, which does not exist.

ricott1 (approved):

proposals/lip-0054.md
gkoumout: ah right, start from 0, sorry my bad. Ok, can be resolved from my side.

gkoumout (approved):
Looks good to me, good job! We have to decide what to do with the figures of LIP 53. Currently, they are outdated. In fact, given that execution now is much more cleaner than in earlier versions, we might even remove them completely.

ishantiw (commented):
@AndreasKendziorra I have one doubt on the key part for the state recovery command but else is looking fine
Below are the issues created for the change .

https://github.com/LiskHQ/lisk-sdk/issues/8718
https://github.com/LiskHQ/lisk-sdk/issues/8719
https://github.com/LiskHQ/lisk-sdk/issues/8720
https://github.com/LiskHQ/lisk-sdk/issues/8721
https://github.com/LiskHQ/lisk-sdk/issues/8722
https://github.com/LiskHQ/lisk-sdk/issues/8723

AndreasKendziorra:
> @AndreasKendziorra I have one doubt on the key part for the state recovery command but else is looking fine Below are the issues created for the change .
> 
> [LiskHQ/lisk-sdk#8718](https://github.com/LiskHQ/lisk-sdk/issues/8718) [LiskHQ/lisk-sdk#8719](https://github.com/LiskHQ/lisk-sdk/issues/8719) [LiskHQ/lisk-sdk#8720](https://github.com/LiskHQ/lisk-sdk/issues/8720) [LiskHQ/lisk-sdk#8721](https://github.com/LiskHQ/lisk-sdk/issues/8721) [LiskHQ/lisk-sdk#8722](https://github.com/LiskHQ/lisk-sdk/issues/8722) [LiskHQ/lisk-sdk#8723](https://github.com/LiskHQ/lisk-sdk/issues/8723)

@ishantiw for the state recovery initialization, there is no issue for

- moving `SMTVerify` to verification
- add schema validation of sidechain account in state recovery initialization command

Should it be added to https://github.com/LiskHQ/lisk-sdk/issues/8722?

proposals/lip-0054.md
```diff
@@ -221,50 +237,58 @@ def verify(trs: Transaction) -> None:
     if recoveryModule does not have a recover function:
         raise Exception("Module is not recoverable.")
 
+    keys = set()
+
+    for entry in trsParams.storeEntries:
+        if entry.storeValue is EMPTY_BYTES:
+            raise Exception("Recovered store value cannot be empty.")
+
+        key = entry.substorePrefix + entry.storeKey
```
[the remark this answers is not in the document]
AndreasKendziorra: No. In this verification step, we only want to ensure that the query keys used in the execution are pairwise distinct. And the query keys in the execution have indeed the construction
`queryKey = storePrefix + entry.substorePrefix + sha256(entry.storeKey)`
But to ensure this property, it is sufficient to check that keys `entry.substorePrefix + sha256(entry.storeKey)` are pairwise distinct. So, we skip 
- adding `storePrefix` which is the same for every query key, and
- hashing `entry.storeKey`
This way, the verification is slightly more efficient.

ishantiw:
@AndreasKendziorra 
- Regarding `add schema validation of sidechain account in state recovery initialization command` it was already fixed as part of this issue https://github.com/LiskHQ/lisk-sdk/issues/8620 on SDK.
- `moving SMTVerify to verification` You mean to execution, right? I have updated this it in the issue https://github.com/LiskHQ/lisk-sdk/issues/8722

ishantiw (approved):

AndreasKendziorra:
>  `moving SMTVerify to verification` You mean to execution, right? I have updated this it in the issue [Fix a bug in state recovery initialization command  lisk-sdk#8722](https://github.com/LiskHQ/lisk-sdk/issues/8722)

yes, I meant to execution. Thanks for updating.

janhack (approved):
Editorial approval

PR #400, opened 2023-07-05
Update ID field inconsistency and encryption mode

### LINKED ISSUE #397: LIP 0067: Update keystoreSchema and mac computation specs
## Motivation
In [LIP-0067](proposals/lip-0067.md):
1. Currently we use full 32bytes of `encryptionKey` for generating `ciphertext` while the last 16bytes of the `encryptionKey` where used again for `mac` computation. This weakens security of the key, as one can brute force only the last 16 bytes of the key (using the mac field) and then the remaining 16 bytes. 

2. [Encryption File Format](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0067.md#encryption-file-format) specifies last field name as `ID` but [examples](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0067.md#examples) use `uuid` which is inconsistent.

## Current Specifications

1. [LIP-0067](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0067.md#mac) computes `ciphertext` using `aes-256-gcm` encryption function where full 32bytes where used as `encryptionKey` while the last 16bytes of the `encryptionKey` where again used for `mac` computation. 

2. The `keystoreSchema` defined in [Encryption File Format](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0067.md#encryption-file-format) specifies last field as 

        
        "ID": {
        
                    "type": "string",

                    "format": "uuid"
       }

whereas [examples](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0067.md#examples) specifies
  
          "uuid": "fa3e4ceb-10dc-41ad-810e-17bf51ed93aa"

## Proposed Specifications

1. Be inline with [web3 keystore specs](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition#definition): Use AES-128-GCM (as it guarantee authenticity and integrity both) instead of AES-256-GCM as we just stop using 256-bit keys for the decryption function, since we don't support 256-bit security anyways. Now, the first 16 bytes of `encryptionKey` where used for generating `ciphertext`. 

2. Change `uuid` to `id` and put version outside of "crypto" as defined in ETH [specs](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition#pbkdf2-sha-256).

## Affected LIPs

LIP 0067

AnanyaSshri:
Resolves #397


mjerkov (changes requested):
First review, just some minor corrections. I suppose the examples still need to be updated.

proposals/lip-0067.md
```diff
@@ -210,13 +209,11 @@ Secret recovery phrase: `target cancel solution recipe vague faint bomb convince
...
     "mac": "a476979ca68fe90f3c96f8a5f3f0a9fe33aef8b091d1169861e44a11a680aae9",
-    "cipher": "aes-256-gcm",
+    "cipher": "aes-128-ctr",
     "cipherparams": {
-      "iv": "da7a74acbf34d20ffd3658f9",
-      "tag": "f4282899ed6cb0193e2981dca0d2ae8e"
+      "iv": "da7a74acbf34d20ffd3658f9"
     },
     "kdf": "argon2id",
     "kdfparams": {
```
ricott1, line 219: Please check if we have to change the tag length to account for different length of the input key to the aes algo

AnanyaSshri: AES-128-CTR mode doesn't have "tag".

proposals/lip-0067.md
```diff
@@ -73,8 +72,9 @@ keystoreSchema = {
     },
     "ID": {
```
ricott1, line 73: Shouldnt this be `id`?

proposals/lip-0067.md
```diff
@@ -46,12 +46,11 @@ We store secret recovery phrases and private keys in a JSON file following the f
 ```java
 keystoreSchema = {
   "type": "object",
-  "required": ["crypto", "metadata", "ID"],
+  "required": ["crypto", "metadata", "ID", "version"],
```
ricott1, line 49: Also here I think it should be `id`.

gkoumout, line 49: To be easier to parse and perform potential future updates, I suggest to use the exact same order as in [ethereum specs](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition#pbkdf2-sha-256):

order: crypto, id, version, metadata.

For crypto the order is {"cipher","cipherparams",  "ciphertext", "kdf", "kdfparams", "mac"}.

proposals/lip-0067.md
```diff
@@ -73,8 +72,9 @@ keystoreSchema = {
     },
     "ID": {
         "type": "string",
-        "format": "uuid"
-    }
+        "format": "id"
```
gkoumout, line 75: What is the difference between "uuid" and "id" here?

proposals/lip-0067.md
```diff
@@ -243,13 +241,11 @@ Key pair derived from the secret recovery phrase above, and the path `m/44'/134'
 ```json
 {
   "crypto": {
-    "version": "1",
     "ciphertext": "086a59889e0e311422eeb15bb6c753aeead210c4494eb37cf7b8f01b0ed372d64e6a08cc77e0bc8170f79f199e2ce7b4c47fe5353e97e67d53c846c029c6cd08",
```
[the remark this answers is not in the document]
gkoumout: Agree with Alessandro.

proposals/lip-0067.md
```diff
@@ -210,13 +209,11 @@ Secret recovery phrase: `target cancel solution recipe vague faint bomb convince
 ```json
 {
   "crypto": {
-    "version": "1",
     "ciphertext": "866c6f1cab3ef67514bdc54cf0143b8b824ebe7c045efb97707c158c81d313cd1a6399b7aa3002248984d39ea2604b0263fe7bdbd8cb04286a9cbd2d353fc79908daab9af04b2528bf4f06a82d79483c",
```
gkoumout, line 212: If I understand well, ciphertext is calculated using the cipher function, so since it changes, this has to change now I guess..

Btw, I realize it is never explicitly stated in the LIP, how "ciphertext" is calculated, perhaps we should add a sentence on it.

gkoumout (changes requested):
I guess the PR is in an intermediate stage. I left some comments.

proposals/lip-0067.md
AnanyaSshri: Done

proposals/lip-0067.md
```diff
@@ -46,12 +46,11 @@ We store secret recovery phrases and private keys in a JSON file following the f
 ```java
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: Updated with "type" field only and removed "format" field.

proposals/lip-0067.md
AnanyaSshri: Line 85 gives the description. Do you want to add more explicit description in "ciphertext" section also?

proposals/lip-0067.md
```diff
@@ -210,28 +205,25 @@ Secret recovery phrase: `target cancel solution recipe vague faint bomb convince
...
-    "cipher": "aes-256-gcm",
-    "cipherparams": {
-      "iv": "da7a74acbf34d20ffd3658f9",
-      "tag": "f4282899ed6cb0193e2981dca0d2ae8e"
-    },
     "kdf": "argon2id",
     "kdfparams": {
       "parallelism": 4,
       "iterations": 1,
       "memory": 2024,
```
mjerkov, line 215: I suggest checking if this is the right value for `memory`, or it should be `2048` as mentioned [here](https://github.com/LiskHQ/lips/blob/87e344e4f6f9d6362e958ab1c0fb2d070cbfe6f3/proposals/lip-0067.md?plain=1#L172).

proposals/lip-0067.md
```diff
@@ -243,31 +235,28 @@ Key pair derived from the secret recovery phrase above, and the path `m/44'/134'
...
-    "cipher": "aes-256-gcm",
-    "cipherparams": {
-      "iv": "aa97507e9f8574b2e7c7ba8b",
-      "tag": "4b3362626c82b0ba1b2de62fb84a1e73"
-    },
     "kdf": "argon2id",
     "kdfparams": {
       "parallelism": 4,
       "iterations": 1,
       "memory": 2024,
```
mjerkov, line 245: Same comment as above.

mjerkov (changes requested):
Looks good, I will do another review after the examples are updated.

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
+#### id
+
+The `id` property stores a provided uuid (version 4 UUID as specified in [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122)), this is a randomly generated ID. It is used if the keystore needs to be referred to. Implementation help: ID generation is supported by nodejs with the [`uuid`](https://github.com/uuidjs/uuid#uuidv4options-buffer-offset) package, for example.
 
-Note that when using AES-256-GCM, the tag is an output of the encryption and is needed for decryption, this is why it is stored in the `cipherparams` property.
+#### version
+
+The `version` is set to `"3"`.
 
 #### metadata
```
gkoumout, line 117: Comment about line 118: I know this text carries over from the past, but I think that what is saying can't be true. In particular, we say that the properties are not required and can be left empty; in LIP 64 however, we disallow having non-required properties. 

Again perhaps, I am missing something (e.g., might be that rules of LIP 64 do not apply for this LIP since it has to do with UI/UX related features and not protocol -- tagging @AndreasKendziorra @ricott1 who might know better) but seems to me that according to LIP 64, the JSON schema should be updated such that:

 - include "required" in "crypto", since it is of type "object".
 - I guess metadata should also be of type object? and have all properties required?

proposals/lip-0067.md
gkoumout: Yep, that would be good. Like one more sentence. 

Also, for the flow of the text, might be better to explain first the explanation of `ciphertext` and then of `cipher` and `cipherparams`. Because somehow we now first discuss how to decrypt and then what is the excrypted message (which should be decrypted..).

proposals/lip-0067.md
gkoumout: Agree with Miroslav

gkoumout (changes requested):
I am not convinced that the proposed change solves the issue, but perhaps I am missing something. 

I also left some comments on parts where text looks confusing (carrying over from previous versions).

proposals/lip-0067.md
AndreasKendziorra: As far as I know this schema is not used for Lisk Codec. If this is the case, then it's fine to have non-required properties. @ishantiw could you please confirm or disprove this?

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
-|---------------|---------------|----------------|------------|
-| "AES-256-GCM" | `aes-256-gcm` | `{iv: string, tag: string}` | [RFC 5116](https://datatracker.ietf.org/doc/html/rfc5116#section-5.2) |
+#### id
+
+The `id` property stores a provided uuid (version 4 UUID as specified in [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122)), this is a randomly generated ID. It is used if the keystore needs to be referred to. Implementation help: ID generation is supported by nodejs with the [`uuid`](https://github.com/uuidjs/uuid#uuidv4options-buffer-offset) package, for example.
 
-Note that when using AES-256-GCM, the tag is an output of the encryption and is needed for decryption, this is why it is stored in the `cipherparams` property.
+#### version
+
+The `version` is set to `"3"`.
```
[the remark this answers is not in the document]
AnanyaSshri: As we are following ETH specs and the updated specs have version 3 where they recommend to use AES-128-CTR.

proposals/lip-0067.md
AnanyaSshri: Mitsu told me there was a recent change of this value in some PR and the value 2097023kilobytes is the closest value to max 2GB that one can set for memory value. So which value should I include? @mjerkov @gkoumout

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
 The encrypted message in hexadecimal format.
 
-##### mac
-
-Computed as `SHA256(encryptionKey[-16:] + cipherText)`, where `encryptionKey` is the key derived from the user password using the key-derivation function. The `mac` can be used to verify the encryption key before starting the decryption process.
-
 ##### kdf and kdfparams
 
-The encryption key is an intermediate key derived from the user password. It is used to generate the secret key for decryption, and verify if the given password is correct. The function, and the params used to derive this key from the password are specified in `kdf`. The following values of `kdf` and `kdfparams` are allowed, depending on the key-derivation function:
+The encryption key is an intermediate key derived from the user password. It is used to generate the secret key for decryption, and verify if the given password is correct. Note that the encryption key is the leftmost 16 bytes of the derived key. The function, and the params used to derive this key from the password are specified in `kdf`. The following values of `kdf` and `kdfparams` are allowed, depending on the key-derivation function:
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
 
-##### version
+##### cipher and cipherparams
+
+The specified function encrypts the secret using the encryption key; to decrypt it, the encryption key along with `cipher` and `cipherparams` must be used. If the encryption key is longer than the key size required by the encoding function, it is truncated to the correct number of bits. The following option is supported:
 
-The `version` is set to `"1"`.
+| `cipher`      | function      | `cipherparams` | Definition |
+|---------------|---------------|----------------|------------|
+| "AES-128-CTR" | `aes-128-ctr` | `{iv: string}` | [RFC 5116](https://datatracker.ietf.org/doc/html/rfc5116#section-5.2) |
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: @ishantiw Can you please comment on Andreas's query to resolve this?

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
+The encrypted message in hexadecimal format. It is generated by inputting the message and the secret key to `aes-128-gcm` function.
 
 ##### kdf and kdfparams
 
-The encryption key is an intermediate key derived from the user password. It is used to generate the secret key for decryption, and verify if the given password is correct. The function, and the params used to derive this key from the password are specified in `kdf`. The following values of `kdf` and `kdfparams` are allowed, depending on the key-derivation function:
+The `derivedKey` is an intermediate key derived from the user password: a derivation function is applied to the user password. The derived key is used to generate the secret key for decryption, and verify if the given password is correct. The secret key is the leftmost 16 bytes of the derived key. The function, and the params used to get the derivation key from the password are specified in `kdf`. The following values of `kdf` and `kdfparams` are allowed, depending on the key-derivation function:
```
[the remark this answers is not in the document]
AndreasKendziorra: > The `derivedKey` is an ...

The term "derivedKey" was not used before in this document. So, it's bit weird to start the sentence like this. Maybe:

"A key derivation function is used to derive an intermediate key, `derivedKey`, from the user password. The derieved key ..."

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
 | `kdf`            | function   | `kdfparams` | Definition |
 |------------------|------------|-------------|------------|
 | "PBKDF2-SHA-256" | `pbkdf2`   | `{iterations: uint32, salt: string}` | [RFC 2898](https://www.ietf.org/rfc/rfc2898.txt) |
 | "argon2id"       | `argon2id` | `{parallelism: uint32, iterations: uint32, memory: uint32, salt: string}` | [RFC 9106](https://datatracker.ietf.org/doc/html/rfc9106) |
 
-##### cipher and cipherparams
+##### mac
 
-The specified function encrypts the secret using the encryption key; to decrypt it, the encryption key along with `cipher` and `cipherparams` must be used. If the encryption key is longer than the key size required by the encoding function, it is truncated to the correct number of bits. The following option is supported:
+Computed as `SHA256(derivedKey[-16:] + ciphertext)`, where `derivedKey` is the intermediate key derived from the user password using the key-derivation function. The `mac` can be used to verify the secret key before starting the decryption process.
```
AndreasKendziorra, line 106: > The `mac` can be used to verify the secret key before starting the decryption process.

Especially here it can be confusing what is meant with "secret key". Because it is used to verify secret key of the signature scheme a.k.a the decrypted message. But the reader might think it is about the encryption key.

AndreasKendziorra (changes requested):

proposals/lip-0067.md
mjerkov: I agree with @AndreasKendziorra having a point, so after all the term `encryption key` is better than `secret key`, especially since it can be also used in the context of decryption. 
Also, note that in the first sentence you explain `where derivedKey is the intermediate key derived from the user password using the key-derivation function`, which is maybe unnecessary since you already explain what `derivedKey` is in the `kdf and kdfparams` subsection.

mjerkov (approved):
Looks good, left just a small comment.

proposals/lip-0067.md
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
 
 #### crypto
 
-##### version
-
-The `version` is set to `"1"`.
+##### cipher and cipherparams
 
-##### ciphertext
+The specified function encrypts the message using the secret key; to decrypt it, the secret key along with `cipher` and `cipherparams` must be used. If the secret key is longer than the key size required by the encoding function, it is truncated to the correct number of bits. The following option is supported:
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: > > The `mac` can be used to verify the secret key before starting the decryption process.
> 
> Especially here it can be confusing what is meant with "secret key". Because it is used to verify secret key of the signature scheme a.k.a the decrypted message. But the reader might think it is about the encryption key.

@AndreasKendziorra I have updated this line in LIP but you can check once if it is correctly written.

proposals/lip-0067.md
```diff
@@ -46,20 +46,21 @@ We store secret recovery phrases and private keys in a JSON file following the f
...
-        "mac": {"type": "string"},
         "kdf": {"type": "string"},
         "kdfparams": {"type": "object"},
-        "cipher": {"type": "string"},
-        "cipherparams": {"type": "object"}
+        "mac": {"type": "string"}
       }
     },
+    "id": {"type": "string"},
+    "version": {"type": "string"},
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
```diff
@@ -46,20 +46,21 @@ We store secret recovery phrases and private keys in a JSON file following the f
 ```java
 keystoreSchema = {
   "type": "object",
-  "required": ["crypto", "metadata", "ID"],
+  "required": ["crypto", "id", "version", "metadata"],
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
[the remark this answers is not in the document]
AndreasKendziorra: I would set it to 3. The specs are not very clear about it, but I would interpret it this way that an implementation of the key storage must support CTR but may also support other block cipher modes. If I'm not wrong, version 1 was using CBC which is also different fro GCM. And overall we are closer to (or even following completely) version 3.

proposals/lip-0067.md
AndreasKendziorra: I was actually wrong. This mac is primarily used to verify that the password for the key derivation function was correct. So, when `derivedKey` is derived, it should be verified with `mac` that the value is correct. If not, either the password was wrong, or the ciphertext was altered. But altering the the ciphertext in this JSON is probably not a realistic scenario. Therefore, I would write:

"Computed as `SHA256(derivedKey[-16:] + ciphertext)`. The `mac` should be used to verify that the password for the key derivation function was correct. This is done by computing the `mac` based on the derived key and ciphertext. If this `mac` does not match with the one in the JSON file the password was not correct."

proposals/lip-0067.md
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: Done

proposals/lip-0067.md
AnanyaSshri: Removed "metadata" from required properties.

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
-The `version` is set to `"1"`.
+##### cipher and cipherparams
 
-##### ciphertext
+The specified function encrypts the message using the encryption key; to decrypt it, the encryption key along with `cipher` and `cipherparams` must be used. If the encryption key is longer than the key size required by the encoding function, it is truncated to the correct number of bits. The following option is supported:
 
-The encrypted message in hexadecimal format.
+| `cipher`      | function      | `cipherparams` | Definition |
+|---------------|---------------|----------------|------------|
+| "AES-128-GCM" | `aes-128-gcm` | `{iv: string}` | [RFC 7714](https://datatracker.ietf.org/doc/html/rfc7714) |
```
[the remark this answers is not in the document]
AnanyaSshri: I missed this one. Thanks for pointing out.

proposals/lip-0067.md
```diff
@@ -210,28 +205,28 @@ Secret recovery phrase: `target cancel solution recipe vague faint bomb convince
...
+      "memory": 2048,
+      "salt": "9cf89f89fead59b4f03a1164b47bed2f"
+    },
+    "mac": "87d989d9681446268c70aa8d22474013220363bfabab6f4d06fcf31e6d9ebfef"
   },
+  "id": "fa3e4ceb-10dc-41ad-810e-17bf51ed93aa",
+  "version": "3",
   "metadata": {
     "name": "Maxime",
     "description": "secret recovery phrase",
```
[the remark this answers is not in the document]
AnanyaSshri: Done

proposals/lip-0067.md
```diff
@@ -243,31 +238,31 @@ Key pair derived from the secret recovery phrase above, and the path `m/44'/134'
...
+      "memory": 2048,
+      "salt": "9cf89f89fead59b4f03a1164b47bed2f"
+    },
+    "mac": "87d989d9681446268c70aa8d22474013220363bfabab6f4d06fcf31e6d9ebfef"
   },
+  "id": "ef52c117-d7cc-4246-bc9d-4dd506bef82f",
+  "version": "3",
   "metadata": {
     "name": "my lisk account",
     "description": "ed25519 key pair",
```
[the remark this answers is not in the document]
AnanyaSshri: Do you mean the examples are inconsistent in the two table?

> But the table above is using Ed25519 private key and here we write ed25519 key pair.

I am not sure I get it. We are deriving the key pair from Ed25519.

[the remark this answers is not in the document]
AndreasKendziorra: I meant in the table in line 131 we specify the value "Ed25519 private key", but here we use "ed25519 key pair". I'm not sure if this property is only read by humans or also by machines. Just in case, I would use the specified value from the table.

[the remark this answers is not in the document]
AnanyaSshri: Okay. I have updated the example with "description: Ed25519 private key".

proposals/lip-0067.md
AnanyaSshri: Update the memory value with 2048.

proposals/lip-0067.md
```diff
@@ -83,36 +80,38 @@ In the following sections, we describe the uses of the properties of `keystoreSc
...
 | `kdf`            | function   | `kdfparams` | Definition |
 |------------------|------------|-------------|------------|
 | "PBKDF2-SHA-256" | `pbkdf2`   | `{iterations: uint32, salt: string}` | [RFC 2898](https://www.ietf.org/rfc/rfc2898.txt) |
 | "argon2id"       | `argon2id` | `{parallelism: uint32, iterations: uint32, memory: uint32, salt: string}` | [RFC 9106](https://datatracker.ietf.org/doc/html/rfc9106) |
 
-##### cipher and cipherparams
+##### mac
 
-The specified function encrypts the secret using the encryption key; to decrypt it, the encryption key along with `cipher` and `cipherparams` must be used. If the encryption key is longer than the key size required by the encoding function, it is truncated to the correct number of bits. The following option is supported:
+Computed as `SHA256(encryptionKey[-16:] + ciphertext)`, where `encryptionKey` is the second-leftmost 16 bytes of the key derived from the user password using the key-derivation function. The `mac` can be used to verify the encryption key before starting the decryption process.
```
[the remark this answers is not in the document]
AnanyaSshri: As discussed, we are not giving 256 bit security so we give 128 bit security and hence, the required changes in aes algo and mac computation.

janhack (approved):
Editorial approval

PR #405, opened 2023-07-06
Fix bounce issue

### LINKED ISSUE #379: Incorrect handling of `ccm.fee` when CCM is bounced
## Motivation

In the current interoperability protocol, bounced messages, i.e., CCMs that are returned to the sending chain when an error occurs 
 during their processing, incur an extra fee on the sending chain. If, for example, the target 
 chain does not support a given module, the CCU relayer receives no fee but the fee of the 
 bounced CCM is reduced by  minFee (see [bounce function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#bounce))  to penalize the chain who sent the "bad" CCM (the 
 relayer of the newly bounced CCM can only get part of the fee that had originated from the 
 original sending chain). The goal is for the originating chain to be able to recover most of 
 the fee (since no real processing occurred), but still be penalized for the "bad" CCM. 

 A problem exists when subtracting  `minFee`  because the  receiving chain does not burn it 
 nor does it update the other sidechain's escrow balance. This means that the sidechain can 
 recover the full fee amount, circumventing the penalisation for the "bad" CCM (we should 
 note that the regular fee for each byte of the CCM is still paid by the relayer when 
 submitting the CCU update transaction).
 
 This is present in the  `apply`  method when a module  is not supported or when a command 
 in a module is not supported, and in the  forward  method  when the mainchain does not 
 have a channel to the receiving chain or when this channel is only in the registered state. 
 Transactions that fail on the execute method are not affected because, in that case, the fee 
 has already been paid to the relayer.

## Current Specifications

The escrow account for the relevant chain is not updated when the ccm fee is reduced in the `bounce` function. 

## Proposed Specifications

The suggested change to the protocol is to apply the following rules:

-  if a `ccm` needs to get bounced, then all `ccm.fee` is assigned to the relayer who submitted the CCU transaction that contained the ccm.
- A `ccm` gets bounced if `ccm.fee >= minimumFee` . The bounced `ccm` will always have `ccm.fee = 0 `.
- In case a `ccm` needs to get bounced but `ccm.fee < minimumFee`, then it gets discarded. 

To achieve this, modifications in [apply](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#apply), [forward](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0049.md#forward) and [bounce](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#bounce) functions are required. 

Also, for the forwarding part (case where mainchain attempts to forward a ccm but then bounces due to unavailable channel), a minor modification in the interface with [`token.beforeCrossChainMessageForwarding` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#beforecrosschainmessageforwarding) is required. 

Analogous changes are performed in LIP 54 for the `applyRecovery` and `forwardRecovery` functions. 

## Affected LIPs

LIP0045, LIP0049, LIP0051,LIP0054

gkoumout:
Fixes #379 .


proposals/lip-0045.md
```diff
@@ -1336,7 +1322,7 @@ def isChannelActive(chainID: ChainID) -> bool:
     else:
         # Account may not exist on a sidechain.
         if chainAccount(chainID) exists:   
-            return chainAccount(chainID).status == CHAIN_STATUS_ACTIVE
+            return isLive(chainID) and chainAccount(chainID).status == CHAIN_STATUS_ACTIVE
```
ricott1, line 1325: Probably should be removed

gkoumout: Removed.

proposals/lip-0045.md
```diff
@@ -874,6 +849,15 @@ def sendInternal(sendingAddress: Address,
         data = {"ccm": ccm},
         topics = [ccm.sendingChainID, ccm.receivingChainID, ccmID]
     )
+
+def emitCCMFailedEvent(ccm: CCM, code: uint32):
```
AndreasKendziorra, line 853: the name suggests that the execution of the ccm failed. Maybe `emitCCMSentFaildEvent`? Or `emitCCMSentFaild` if it's too long.

ricott1 (approved):

proposals/lip-0045.md
gkoumout: Done. Used `emitCCMSentFailedEvent`.

proposals/lip-0054.md
```diff
@@ -113,15 +113,10 @@ Furthermore, for the rest of this section:
...
-def emitCCMEvent(ccm: CCM, result: uint32, code: uint32) -> None:
-    emitEvent(
-        module = MODULE_NAME_INTEROPERABILITY,
-        name = EVENT_NAME_CCM_PROCESSED,
-        data = {"ccm": ccm, "result": result, "code": code},
-        topics = [ccm.sendingChainID, ccm.receivingChainID]
-    )
 ```
 
+We also make use of the [`emitCCMEvent` function][[lip-0045#emitccmevent]] defined in [LIP 0045][lip-0045].
```
AndreasKendziorra, line 118:
```suggestion
We also make use of the [`emitCCMEvent` function][lip-0045#emitccmevent] defined in [LIP 0045][lip-0045].
```

AndreasKendziorra (changes requested):

proposals/lip-0053.md
```diff
@@ -404,6 +404,8 @@ def verifyRoutingRules(ccu: CCU, ccm: CCM) -> None:
 
 #### beforeCrossChainMessagesExecution
 
+Here, we make use of the [`emitCCMEvent` function][lip-0045#emitccmevent] defined in LIP 0045.
```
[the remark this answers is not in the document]
gkoumout: Good point, thanks! I added it in LIP 54 but not 53. Fixed now.

proposals/lip-0054.md
gkoumout: This was fixed in my local machine but forgot to push, should be ok now.

AndreasKendziorra (approved):

ishantiw (approved):
LGTM, https://github.com/LiskHQ/lisk-sdk/issues/8655 is the corresponding SDK issue that contains multiple issue to handle implementation on SDK

janhack (approved):
Editorial approval

PR #417, opened 2023-07-17
Update values of configurable constants defining time intervals

### LINKED ISSUE #416: Make consistent configurable constants that correspond to time intervals.
## Motivation

In the Lisk protocol, several modules expose configurable constants whose values corresponds to time intervals, such as the liveness period for sidechains (1 month) in interoperability module and the locking period for staking in PoS module. The way that the default values of those constants is defined should be consistent across all modules. 

## Current Specifications

For each of those constants, the value has been defined by deciding some duration (e.g., 1 month for PoS self-staking locking period) and then calculating manually a value, perhaps by introducing some rounding. This makes it much harder to parse the actual value and get intuition and perhaps it becomes error prone. 

As an example, the self-staking locking period in PoS module is intended to be one month.  The actual definition in the LIP is:

>  `LOCKING_PERIOD_SELF_STAKING` | uint32 | `260,000`|  The [locking period][lip-0023#explicit-unlock-mechanism] for self-staked tokens. 

Here it might be not so clear how this value has been obtained. There are multiple issues here:
 
 - It is not clearly explained that the locking period is counted in terms of "number of blocks" and not in seconds.
 - One day has exactly 24x360 = 8640 blocks. The actual value 260.000 is not a multiple of 8640. It has been obtained as `8640 x 30 = 259.200` and then applying arbitrarily some rounding to get the number 260.000

This makes quite arbitrary the way constants are defined, and makes hard to maintain a good intuition on the meaning of each value

## Proposed Specifications

To achieve consistency throughout the modules, we propose the following improvements:

- For all configurable constants that correspond to time intervals, we clearly specify the measurement unit (seconds, blocks, etc.).
- In the specs, beyond the final value of the constant, we also specify the way it is extracted. An example is the `LIVENESS_LIMIT` constant in interoperability which is defined in [LIP 0045](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md) as `30x24x3600`.
- Since the notion of a "month" is not well defined (not all months have same duration), we rather use the notion of weeks. Therefore, constants that have intended value of a month, are switched to "4 weeks", i.e., 28 days. 

## Affected LIPs

0045, 0053, 0054, 0057, 0061, 0063, 0070 .

gkoumout:
Fixes #416 .


proposals/lip-0045.md
```diff
@@ -216,7 +216,7 @@ In this section, we specify the substores that are part of the Interoperability
 | `CHAIN_REGISTRATION_FEE` | uint64 | 1000000000 | Fee to pay for a sidechain registration command in Beddows. |
 | `EMPTY_HASH` | bytes | `sha256(b"")` | SHA-256 hash of empty bytes. |
 | `EMPTY_CCM` | CCM | A CCM object partially following the [`crossChainMessageSchema`][lip-0049#ccmschema] schema with all properties set to their default values (see exact definition [below](#empty-cross-chain-message)). | The empty ccm object. |
-| `LIVENESS_LIMIT` | uint32 | 30x24x3600 | The maximum time interval for the liveness condition. |
+| `LIVENESS_LIMIT` | uint32 | 28x24x3600 | The maximum time interval (in seconds) for the liveness condition. |
```
janhack, line 219:
```suggestion
| `LIVENESS_LIMIT` | uint32 | 28 * 24 * 3600 | The maximum time interval (in seconds) for the liveness condition. |
```

janhack: I would strongly suggest to use `*` for multiplication with spacing as then it can be easily copied into a calculator or code instead of abusing the character `x` to achieve a certain visual output. Same below.

proposals/lip-0057.md
```diff
@@ -113,11 +113,11 @@ For the rest of this proposal we define the following constants:
...
 | `FAIL_SAFE_MISSED_BLOCKS` | uint32 | `50` | The number of consecutive missed blocks used in the fail safe banning mechanism. |
-| `FAIL_SAFE_INACTIVE_WINDOW` | uint32 | `130,000` | The length of the inactivity window used in the fail safe banning mechanism. |
-| `LOCKING_PERIOD_STAKING` | uint32 | `26,000` | The [locking period][lip-0023#explicit-unlock-mechanism] for regular staked tokens. |
-| `LOCKING_PERIOD_SELF_STAKING` | uint32 | `260,000` | The [locking period][lip-0023#explicit-unlock-mechanism] for self-staked tokens. |
-| `PUNISHMENT_WINDOW_STAKING` | uint32 | `260,000` | The punishment period for regular stakes, i.e., how many blocks after misbehavior report stakers can unlock tokens unstaked from a punished validator. |
-| `PUNISHMENT_WINDOW_SELF_STAKING` | uint32 | `780,000` | The punishment period for self-stakes, i.e., how many blocks after misbehavior report punished validators can unlock their self-unstaked tokens. |
+| `FAIL_SAFE_INACTIVE_WINDOW` | uint32 | 120960 = 14x24x360 | The length of the inactivity window used in the fail safe banning mechanism, measured in number of blocks. |
```
janhack, line 116:
```suggestion
| `FAIL_SAFE_INACTIVE_WINDOW` | uint32 | 120960 = 14 * 24 * 3600 // BLOCK_TIME | The length of the inactivity window used in the fail safe banning mechanism, measured in number of blocks. |
```

janhack: I think using the block time for the Lisk Mainchain in addition makes the computation more transparent and would make it easier for sidechains with a different block time to compute the analogous constant value. Similar, below.

proposals/lip-0057.md
```diff
@@ -113,11 +113,11 @@ For the rest of this proposal we define the following constants:
...
+| `LOCKING_PERIOD_STAKING` | uint32 | 25,920 = 3x24x360 | The [locking period][lip-0023#explicit-unlock-mechanism] for regular staked tokens, measured in number of blocks. |
+| `LOCKING_PERIOD_SELF_STAKING` | uint32 | 241,920 = 28x24x360 | The [locking period][lip-0023#explicit-unlock-mechanism] for self-staked tokens. |
+| `PUNISHMENT_WINDOW_STAKING` | uint32 | 241,920 = 28x24x360 | The punishment period for regular stakes, i.e., how many blocks after misbehavior report stakers can unlock tokens unstaked from a punished validator. |
+| `PUNISHMENT_WINDOW_SELF_STAKING` | uint32 | 786,240 = 13x7x24x3600 | The punishment period for self-stakes, i.e., how many blocks after misbehavior report punished validators can unlock their self-unstaked tokens. |
```
janhack, line 120: Maybe 3 * LOCKING_PERIOD_SELF_STAKING = ... That we now compute this constant using weeks and the rounding up to 13 weeks seems also quite arbitrarily.

janhack (changes requested):
Looks good overall, I mostly made cosmetic suggestions. 

Did you also check the rationale and other text of the LIPs if this needs to be updated?

sergeyshemyakov (changes requested):
In LIP 0063 there is a [parameter reference table](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0063.md#constants-1), the values should also be updated there.

Also note that LIP 0063 defined `POS_INIT_ROUNDS = 60,480`, it should be changed to `7*24*360`.

I checked other LIPs as well, I think nothing else should be updated.

proposals/lip-0045.md
gkoumout: Right, good catch. In LIP 45 for the liveness "x" was used and I just pasted, but I agree, it is much nicer in the way you suggest.

proposals/lip-0057.md
gkoumout: Committed suggestion. Although I remember there was some issue about block time in SDK and somehow we were avoiding to use it, but I might be missing something.

proposals/lip-0057.md
gkoumout: I wrote in the issue description that we use weeks instead of months as a unit of measure.

About the actual value, ok, makes more sense to just multiply by 3.

janhack:
@sergeyshemyakov I think it would be nice to also update the Governance module PR analogously.

janhack (approved):
Looks good. Nice work, in particular, on thoroughly checking all the occurrences in the rationale.

sergeyshemyakov (approved):
Looks good!

proposals/lip-0045.md
```diff
@@ -79,7 +79,7 @@ This command is used to initialize a terminated outbox account and can only be i
 
 ### Liveness Condition
 
-Active sidechains are required to prove their liveness to the mainchain at least once every 30 days. This is done by including a CCU with non-empty certificate on the mainchain. If a sidechain does not post such a CCU within 30 days, the sidechain account is terminated. This rule guarantees that users do not send funds to inactive sidechains and that users who have tokens in a sidechain which stops communicating with the ecosystem can recover their tokens.
+Active sidechains are required to prove their liveness to the mainchain at least once every 28 days. This is done by including a CCU with non-empty certificate on the mainchain. If a sidechain does not post such a CCU within 24 days, the sidechain account is terminated. This rule guarantees that users do not send funds to inactive sidechains and that users who have tokens in a sidechain which stops communicating with the ecosystem can recover their tokens.
```
ishantiw, line 82:
```suggestion
Active sidechains are required to prove their liveness to the mainchain at least once every 28 days. This is done by including a CCU with non-empty certificate on the mainchain. If a sidechain does not post such a CCU within 28 days, the sidechain account is terminated. This rule guarantees that users do not send funds to inactive sidechains and that users who have tokens in a sidechain which stops communicating with the ecosystem can recover their tokens.
```

ishantiw (commented):
Created related SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8751

proposals/lip-0045.md
gkoumout: Good catch, thanks!

ishantiw (approved):

janhack (approved):
Editorial approval

Issue #422, opened 2023-07-20
Improve LIP 0051 - endpoints, values for constants, clarify token support


mjerkov:
## Motivation
Several sections of the [Token module LIP](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md) can be improved:
1. The [Supported Tokens](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#supported-tokens) section could be improved on clarity of exposition.
2. In [Constant and Notations](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations) section, all constants should have defined values.
3. [Endpoints for Off-Chain Services](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#endpoints-for-off-chain-services) section should have non-trivial or recommended endpoints listed.

## Current Specifications
1. The following paragraph could be considered logically inconsistent, since in the opening sentence it is stated that all chains must support their native tokens, while later in the section an example of a chain with no native tokens is given, i.e. an example of a chain that only supports the LSK token:
> The choice of supported tokens must abide by two rules: _all chains must support their native tokens and all chains must support the LSK token_. The supported tokens can be specified as part of the initial configuration of the Token module at the chain creation. 
2. Multiple constants have missing values.
3. No endpoints are listed.

## Proposed Specifications
1. We suggest to rephrase the above paragraph in the following way:
> The choice of supported tokens must abide by two rules: _all chains must support the LSK token and their native tokens, if they have any_. The supported tokens can be specified as part of the initial configuration of the Token module at the chain creation. 
2. We suggest the following values:

| Constant                                           | Suggested value                      |
|----------------------------------------------------|--------------------------------------|
| `CROSS_CHAIN_COMMAND_NAME_TRANSFER`                | “crossChainTransferCommand”          |
| `EVENT_NAME_INITIALIZE_TOKEN`                      | “initializeToken”                    |
| `EVENT_NAME_INITIALIZE_USER_ACCOUNT`               | “initializeUserAccount”              |
| `EVENT_NAME_INITIALIZE_ESCROW_ACCOUNT`             | “initializeEscrowAccount”            |
| `EVENT_NAME_ALL_TOKENS_SUPPORTED`                  | “allTokensSupported”                 |
| `EVENT_NAME_ALL_TOKENS_SUPPORT_REMOVED`            | “allTokensSupportedRemoved”          |
[...]

PR #423, opened 2023-07-21
Improve LIP 0051 - endpoints, values for constants, clarify token support

mjerkov:
Resolves [Issue 422](https://github.com/LiskHQ/lips/issues/422).

### DIFF
--- a/proposals/lip-0051.md
+++ b/proposals/lip-0051.md
@@ -44,7 +44,7 @@ In particular, the LSK token is native to the Lisk mainchain which has `chainID 
 
 All chains are allowed to select which tokens their protocol supports. Supporting a token only implies that users of the chain can hold those tokens and handle them as specified in this LIP. It does not mean that the chain implements custom logic associated with those tokens.
 
-The choice of supported tokens must abide by two rules: all chains must support their native tokens and all chains must support the LSK token. The supported tokens can be specified as part of the initial configuration of the Token module at the chain creation. For example:
+The choice of supported tokens must abide by two rules: all chains support by default the LSK token and their native tokens, if they have any. Supported non-native tokens can be specified as part of the initial configuration of the Token module at the chain creation. For example:
 
 * A decentralized exchange could support all tokens.
 * A chain with a specific use case and no native token could only support the LSK token.
@@ -127,7 +127,7 @@ The following constants are used throughout the document:
 | `MODULE_NAME_TOKEN` | string | "token" |
 | `COMMAND_NAME_TRANSFER` | string | "transfer" |
 | `COMMAND_NAME_CROSS_CHAIN_TRANSFER` | string | "transferCrossChain" |
-| `CROSS_CHAIN_COMMAND_NAME_TRANSFER` | string | TBD |
+| `CROSS_CHAIN_COMMAND_NAME_TRANSFER` | string | “crossChainTransferCommand” |
 | `CCM_STATUS_OK` | uint32 | 0 |
 | `CCM_STATUS_TOKEN_NOT_SUPPORTED` | uint32 | 64 |
 | `CCM_STATUS_PROTOCOL_VIOLATION` | uint32 | 65 |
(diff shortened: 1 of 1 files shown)


proposals/lip-0051.md
```diff
@@ -127,7 +127,7 @@ The following constants are used throughout the document:
 | `MODULE_NAME_TOKEN` | string | "token" |
 | `COMMAND_NAME_TRANSFER` | string | "transfer" |
 | `COMMAND_NAME_CROSS_CHAIN_TRANSFER` | string | "transferCrossChain" |
-| `CROSS_CHAIN_COMMAND_NAME_TRANSFER` | string | TBD |
+| `CROSS_CHAIN_COMMAND_NAME_TRANSFER` | string | “crossChainTransferCommand” |
```
gkoumout, line 130: In the implementation I see they used a different value: ''transferCrossChain", see [here](https://github.com/LiskHQ/lisk-sdk/blob/0f1296c62585f26f3291f3e1a01aa757801f93f0/framework/src/modules/token/constants.ts#L15). 

We should also make sure that for commands we have the same values as in the implementation. If not, we should follow their naming conventions for simplicity.

proposals/lip-0051.md
```diff
@@ -2625,7 +2625,19 @@ def beforeCrossChainMessageForwarding(trs: Transaction, ccm: CCM, ccmFailed: boo
 
 ### Endpoints for Off-Chain Services
 
-This section specifies the non-trivial or recommended endpoints of the Token module and does not include all endpoints. TBA
+This section specifies the non-trivial or recommended endpoints of the Token module and does not include all endpoints.
+
+#### getBalances
+
+This function returns available and locked balances for all tokens associated with a given address.
```
gkoumout, line 2632: This essentially outputs all entries of the user substore where the key contains the `address`, I would treat is as trivial.

proposals/lip-0051.md
```diff
@@ -2625,7 +2625,19 @@ def beforeCrossChainMessageForwarding(trs: Transaction, ccm: CCM, ccmFailed: boo
...
+
+#### getSupportedTokens
+
+This function returns token IDs for both native and non-native tokens supported by the chain.
```
gkoumout, line 2636: I guess this function is worthwhile including..

proposals/lip-0051.md
```diff
@@ -2625,7 +2625,19 @@ def beforeCrossChainMessageForwarding(trs: Transaction, ccm: CCM, ccmFailed: boo
...
+
+#### getEscrowedAmounts
+
+This function returns the escrowed amounts and escrowed chain IDs for all escrowed tokens.
```
gkoumout, line 2640: Isn't this essentially outputing the whole escrow substore? If yes, I would also consider it trivial.

proposals/lip-0051.md
sergeyshemyakov: Note that "transferCrossChain" is also the value of  `COMMAND_NAME_CROSS_CHAIN_TRANSFER`. I suggest using the string proposed by Miroslav to avoid confusion (also update the SDK value).

mjerkov: Ah, I used that name because of [this line](https://github.com/LiskHQ/lisk-sdk/blob/0f1296c62585f26f3291f3e1a01aa757801f93f0/framework/src/modules/token/module.ts#L83). What should it be (@sergeyshemyakov)?

proposals/lip-0051.md
mjerkov: OK, will remove it. I agree it is trivial...

proposals/lip-0051.md
mjerkov: Removed.

proposals/lip-0051.md
sergeyshemyakov: I would say `CROSS_CHAIN_COMMAND_NAME_TRANSFER` should stay "crossChainTransferCommand" and the SDK should be notified so they make sure they align with the LIP.

sergeyshemyakov (approved):

proposals/lip-0051.md
mjerkov: @sitetester I wouldn't say we should change the name of the constant, we're debating about it's _value_

proposals/lip-0051.md
```diff
@@ -166,18 +166,18 @@ We further use the utility function `getMainchainID()` defined in [LIP 0037][lip
 | `EVENT_NAME_BURN` | string | "burn" | Used for events emitted during calls to the [burn](#burn) function. |
 | `EVENT_NAME_LOCK` | string | "lock" | Used for events emitted during calls to the [lock](#lock) function. |
 | `EVENT_NAME_UNLOCK` | string | "unlock" | Used for events emitted during calls to the [unlock](#unlock) function. |
-| `EVENT_NAME_INITIALIZE_TOKEN` | string | TBD | Used for events emitted during calls to the [initializeToken](#initializetoken-1) function. |
-| `EVENT_NAME_INITIALIZE_USER_ACCOUNT` | string | TBD | Used for events emitted during initialization of user substore entry.|
-| `EVENT_NAME_INITIALIZE_ESCROW_ACCOUNT` | string | TBD | Used for events emitted during initialization of escrow substore entry.|
+| `EVENT_NAME_INITIALIZE_TOKEN` | string | “initializeToken” | Used for events emitted during calls to the [initializeToken](#initializetoken-1) function. |
```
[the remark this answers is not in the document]
janhack: Changes of constant names are out-of-scope for the PR so I would keep them as is.

proposals/lip-0051.md
janhack: The name of the cross-chain command is also "transferCrossChain" in the implementation so I updated it accordingly. From the SDK team they see now issue with confusions and I think we should then follow their decision for the LIP.

janhack (approved):

sergeyshemyakov (approved):
Looks good.

PR #429, opened 2023-07-25
Add signing function for non protocol message

AndreasKendziorra:


proposals/lip-0062.md
```diff
@@ -25,6 +25,12 @@ Introducing memory efficient signing allows all objects used in the Lisk protoco
 
 With the current proposal, signatures are computed against the hash of the message to be signed and hence always computed for a byte string of small constant size.
 
+Lisk Elements currently contains another Ed25519 signature scheme [that uses double pre-hashing](https://github.com/LiskHQ/lisk-sdk/blob/0f1296c62585f26f3291f3e1a01aa757801f93f0/elements/lisk-cryptography/src/ed.ts#L99-L134). This scheme is only used for non-protocol related offchain signatures. More precisely the _sign message_ feature in Lisk Desktop is using it, and it is implemented in Lisk Commander without a concrete use case. The usage of double pre-hashing next to single pre-hashing yields a small risk for collisions. Notice that the existing double pre-hashing signature scheme prepends the tag `TAG = 0x15 + b"Lisk Signed Message:\n"` and the length of the message to the message prior to hashing. If one finds messages `m1` and `m2`, a message tag `MST` according to [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#message-tags-1) and a chain ID `chainID` such that
+```
+sha256(TAG + len(m1) + m1) = MST + chainID + m2
```
sergeyshemyakov, line 30: I would expand on the danger of the collision and explain that if `m2` is a transaction by someone, and `len(MST + chainID + m2) == 32` and an attacker finds a correct `m1` then they can pretend that they signed the message `TAG + len(m1) + m1`. I think this is the only reasonable attack scenario, or do I misunderstand something?

sergeyshemyakov (commented):
Just one suggestion for the rationale.

proposals/lip-0062.md
```diff
@@ -37,6 +43,10 @@ Introduction of another hash function to the Lisk protocol was suggested in [LIP
 
 [LIP 0008][lip-0008] removed the pre-hashing step from the Lisk signing protocol. This was mainly done as this step was deemed unnecessary and removing it could improve the theoretical security of the signing process. However this is not reflected by a practical improvement in protocol security as the SHA-256 hash function is considered collision resistant and secure. This hash function is used in several critical parts of many applications, including Lisk, and any future findings that SHA-256 is insecure would require changing the protocol throughout.
 
+### Signing Non-Protocol Messages
+
+When signing a message with the _sign message_ feature in Lisk Desktop, then the message is typically independet of any chain. Hence, no chain ID needs to be added to the message prior to hashing or signing. Therefore, we add an extra signing function for this use case that works the same as the regular Ed25519 signature scheme with pre-hashing but omits a chain ID. This function should be used in Lisk Desktop for the _sign message_ feature. The double pre-hashing scheme used in Lisk Commander should simply be removed without replacement as there is no use case for it.
```
AnanyaSshri, line 48: > ".....the regular Ed25519 signature scheme with pre-hashing...."
Just to clarify, here are we talking about single pre-hashing? 

Since we are using the terms "single pre-hashing" and "double pre-hashing", it would be good to write explicitly here also in this sentence for the sake of clarity maybe.

AnanyaSshri (approved):
Approved, just a minor suggestion.

proposals/lip-0062.md
AndreasKendziorra: I expanded a bit on it, but kept it more general and close to the attack scenario described in the audit report. The example with a transaction would also not work as the binary message would be too large.

proposals/lip-0062.md
AndreasKendziorra: good point. I added "single"

sergeyshemyakov (approved):
Looks good to me now.

shuse2 (approved):

proposals/lip-0062.md
```diff
@@ -60,6 +70,23 @@ Here, <code>[Sign](https://tools.ietf.org/html/rfc8032#section-5.1.6)</code> and
 
 The `signEd25519` and `verifyEd25519` functions defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#signing-and-verifying-with-ed25519) are superseded by the functions defined in this LIP.
 
+#### Signing Non-Protocol Messages with Ed25519
+
+A non-protocol related message `message` can be signed by the function `signNonProtocolEd25519`. The resulting signature can be verified by `verifyNonProtocolEd25519`. This scheme should be used for the _sign message_ feature in Lisk Desktop. Note that the tag `MESSAGE_TAG_NON_PROTOCOL_MESSAGE` is defined in [LIP 0037](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0037.md#message-tags-1).
+
+```python
+def signNonProtocolEd25519(sk: bytes, message: bytes) -> bytes:
```
shuse2, line 78: maybe not necessary to align, but for the SDK we have functions like
```
signMessageWithPrivateKey
verifyMessageWithPublicKey
```
so and it's used like `ed.signMessageWithPrivateKey`. I would prefer to keep those names

shuse2:
Created corresponding issue on SDK: https://github.com/LiskHQ/lisk-sdk/issues/8775

proposals/lip-0062.md
AndreasKendziorra: I see the reason to keep the names in the SDK and it's fine for me. Would be good to haven some comments then. I'm also open to new names for the LIP as I'm not satisfied with them as well. But I didn't have a better idea. But I would like to avoid `signMessageWithPrivateKey` as it is really not descriptive.

AndreasKendziorra: An alternative could be `signOffchainDataEd25519`.

eniolam1000752 (approved):

janhack (approved):
Editorial approval

Issue #415, opened 2023-07-13
Use `verifyInclusionProof` API for SMT inclusion proofs verification


ricott1:
## Motivation

For SDK modules, we should use the new API to verify an explicit inclusion proof rather than the generic one to verify a proof.

This issue is linked to https://github.com/LiskHQ/lisk-db/issues/116 and https://github.com/LiskHQ/lips/issues/293.

PR #432, opened 2023-07-31
Use the `smtVerifyInclusionProof` API from the SMT rather than `verify`

ricott1:
Closes https://github.com/LiskHQ/lips/issues/415.

### DIFF
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/introduce-cross-chain-update-mechani
 Status: Draft
 Type: Standards Track
 Created: 2021-05-22
-Updated: 2023-07-19
+Updated: 2023-08-08
 Requires: 0045, 0049, 0058, 0061
 ```
 
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -105,7 +105,7 @@ The Interoperability module supports two commands used to certify the state of a
 
 ### Notation and Constants
 
-All interoperability constants are defined in [LIP 0045][lip-0045#constants].
+All interoperability constants are defined in [LIP 0045][lip-0045#constants]. `SMTVerifyInclusionProof` is the `verify_inclusion_proof` function specified in [LIP 0039][lip-0039].
 
 ### Auxiliary Functions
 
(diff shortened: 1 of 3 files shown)


proposals/lip-0054.md
```diff
@@ -107,7 +107,7 @@ Furthermore, for the rest of this section:
 
 * Let `RMTVerify` be the `verifyDataBlock` function specified in [the appendix C of LIP 0031][lip-0031#appendixC].
 * Let `RMTCalculateRoot` be the `calculateRootFromUpdateData` function specified in [the appendix E of LIP 0031][lip-0031#appendixE].
-* Let `SMTVerify` be the `verify` function specified in [LIP 0039][lip-0039].
+* Let `SMTVerifyInclusionProof` be the `verify_inclusion_proof` function specified in [LIP 0039][lip-0039].
```
hrmhatef, line 110: There is no specified function in `lip-0039` with name verify_inclusion_proof, I guess we should change the name of the `verify` function on that file too.

hrmhatef (changes requested):
What about `verifyNonInclusionProof` function?
Should be added in this PR?

proposals/lip-0054.md
ricott1: It is specified in this PR: https://github.com/LiskHQ/lips/pull/431

ricott1:
@hrmhatef the `verify_non_inclusion_proof` API is not used in the protocol, hence there is no need to update any LIP in this sense

hrmhatef (approved):

proposals/lip-0053.md
```diff
@@ -105,7 +105,7 @@ The Interoperability module supports two commands used to certify the state of a
 
 ### Notation and Constants
 
-All interoperability constants are defined in [LIP 0045][lip-0045#constants].
+All interoperability constants are defined in [LIP 0045][lip-0045#constants]. `SMTVerifyInclusionProof` is the `verify_inclusion_proof` function specified in [LIP 0039][lip-0039].
```
janhack, line 108: We have always used `camelCase` for functions in LIPs, same as in the Lisk SDK. I don't understand why we deviate for it here and in LIP 39. Moreover, it would be nice to just use the same function name here and in LIP 39.

janhack (changes requested):
I think it would be good to follow our naming conventions.

proposals/lip-0053.md
ricott1: I think it's better to be specific on the verify function used as in this LIP we also have some functions related to the regular Merkle tree. We could rename the functions in LIP0039 directly, otherwise

janhack (approved):

shuse2:
SDK should be updated with https://github.com/LiskHQ/lisk-sdk/issues/8784

janhack (approved):
Editorial approval

Issue #436, opened 2023-08-04
Add `getMessageFeeTokenIDFromCCM` method to LIP0045


ricott1:
## Motivation

Add a method to get the message fee token ID unambiguously from a CCM.

## Current Specifications

The Interoperability method `getMessageFeeTokenID` takes a chainID as input. This leaves an ambiguity is choosing whether to use the sending or receiving chain of a CCM to get the message fee token ID.

## Proposed Specifications

Add a method to get the message fee token ID unambiguously from a CCM in LIP 0045.

## Affected LIPs
LIP0045

PR #437, opened 2023-08-04
Add `getMessageFeeTokenIDFromCCM` method to Interoperability module

ricott1:
Closes https://github.com/LiskHQ/lips/issues/436.

### DIFF
--- a/proposals/lip-0045.md
+++ b/proposals/lip-0045.md
@@ -1393,6 +1393,9 @@ def getChainAccount(chainID: ChainID) -> ChainAccount:
 
 ```python
 def getMessageFeeTokenID(chainID: ChainID) -> TokenID:
+    if chainID == ownChainAccount.chainID:
+        raise Exception('Channel with own chain account does not exist.')
+        
     # Processing on a sidechain.
     if ownChainAccount.chainID != getMainchainID():
         # Check for direct channel.
--- a/proposals/lip-0045.md
+++ b/proposals/lip-0045.md
@@ -1405,6 +1405,13 @@ def getMessageFeeTokenID(chainID: ChainID) -> TokenID:
     return channel(chainID).messageFeeTokenID
 ```
 
+#### getMessageFeeTokenIDFromCCM
+
+```python
+def getMessageFeeTokenIDFromCCM(ccm: CCM) -> TokenID:
+    return getMessageFeeTokenID(ccm.sendingChainID)
+```
+
 #### getMinReturnFeePerByte
 
 ```python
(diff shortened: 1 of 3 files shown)


AnanyaSshri (approved):

ishantiw (approved):

proposals/lip-0051.md
```diff
@@ -2551,7 +2551,7 @@ def verifyCrossChainMessage(trs: Transaction, ccm: CCM) -> None:
 def beforeCrossChainCommandExecution(trs: Transaction, ccm: CCM) -> None:
     relayerAddress = sha256(trs.senderPublicKey)[:ADDRESS_LENGTH]
 
-    messageFeeTokenID = Interoperability.getMessageFeeTokenID(ccm.receivingChainID)
```
gkoumout, line 2554: Question to make sure I understand: That means that our previous protocol was wrong because we are already in the receiving chain while running this code, i.e., `ccm.receivingChainID = OWN_CHAIN_ID`, right?

Btw, while checking this I had a look in the `getMessageFeeTokenID` function and it seems to me that it does not have the right functionality for sidechains in case it gets as input `OWN_CHAIN_ID`. See my comment in LIP 45.

proposals/lip-0045.md
```diff
@@ -1405,6 +1405,13 @@ def getMessageFeeTokenID(chainID: ChainID) -> TokenID:
     return channel(chainID).messageFeeTokenID
 ```
```
gkoumout, line 1407: Comment about the old `getMessageFeeTokenID` function:

In case the arg `chainID = OWN_CHAIN_ID` and `OWN_CHAIN_ID` is a sidechain, I guess the intended behavior is to output some error "input chain is own chain", but according to current specs it would just return the messageFeeTokenID used in the channel with mainchain.

gkoumout (changes requested):

proposals/lip-0045.md
ricott1: Yes, which is probably at least misleading. We should probably just return an error as you suggest.

proposals/lip-0051.md
ricott1: yeah. Im wondering if using the sending chain is 'secure' in this sense, or if we should try to get it from the receiving chain in case that fails

gkoumout: Yep. In the "normal case" of direct channel communication, we are always running this in the receiving chain, so we should be checking the sending chain channel. 

For the forwarding it is not that important I guess, this is why we did not pay much attention here: Since all channels between mainchain and sidechains have the same message fee token (no matter if is LSK or anything else in the future), then checking sending/receiving chain should give the same result, so sending chain is also fine. Perhaps we should document that it is very important to maintain this property in the future (i.e., "all channels with mainchain should use the same message fee token"), this is something that auditors were asking I guess..

gkoumout (approved):

ricott1:
Related PR: https://github.com/LiskHQ/lisk-sdk/pull/8806

janhack (approved):
Editorial approval

PR #450, opened 2023-08-14
Improve explanation about index representation/serialization in Merkle trees

### LINKED ISSUE #447: LIP 0031: Statement about decoding indices is misleading
In the section [Proof Serialization](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0031.md#proof-serialization) of LIP 0031, it currently says for the `idxs` property:

> At the moment of decoding, each index is converted from unsigned integer to binary, the first bit (which equals 1) is dropped and only the following bits are kept.

This is misleading, because dropping the most significant bit is not (necessarily) done when decoding an object against the `proof` schema defined in [LIP 0031](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0031.md#proof-serialization). In the SDK, the most significant bit is dropped in the function [getLocation](https://github.com/LiskHQ/lisk-sdk/blob/e990c678f6d43debfc9960e2a8f5f98ceaa9b87b/elements/lisk-tree/src/merkle_tree/utils.ts#L185-L187).

This formulation should be improved.

## Affected LIPs

0031

AndreasKendziorra:
Fixes #447

### DIFF
--- a/proposals/lip-0031.md
+++ b/proposals/lip-0031.md
@@ -84,6 +84,10 @@ The procedure for a proof-of-inclusion is as follows:
 
 In Fig. 1 we sketch a simple proof-of-inclusion for data block `data1`. The Verifier asks the Prover for a proof for `queryHashes=[leafHash(data1)]`. The Prover sends back `proof` with `size=5`, `idxs=[0001]`and `siblingHashes=[h0, h6, h4]`.
 
+### Representation of indices
+
+On an abstract level, indices in the array `idxs` are represented as bit strings. The length of such a bit string is [determined by the layer and the height of the tree](#proof-construction). For compactness, these bit strings are converted to `uint64` before being serialized such that they are [varint encoded](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0027.md#base-128-varints). As the indices can represent arbitrary nodes of the tree and not just leave nodes, different indices could represent the same integer. For example, `idxs` could contain the two indices `idx1 = "001"` and `idx2 = "0001"` (see [Fig. 2](#proof-of-inclusion)). Both would be converted to the unsigned integer `0`. To avoid this, a `1` is prepended to the bit string prior to converting to `uint64`. This way, `idx1` is converted to `0b1001 = 9` and `idx2` to `0b10001 = 17`.
+
 ## Specification
 
 ![Merkle tree with path](lip-0031/merkle_tree_with_path.png)
@@ -141,7 +145,7 @@ For an input data array of arbitrary length of the form:
 we define an array `proof` containing 3 parts:
 
 1. `size`: The total length of the data array `data`.
-2. `idxs`: An array indicating the position in the data array `data` of the elements to be verified, corresponding to `queryHashes`.
+2. `idxs`: An array indicating the position in the data array `data` of the elements to be verified, corresponding to `queryHashes`. Each entry is a bit string where the length is determined by the layer and the tree height (see [below](#proof-construction)). 
 3. `siblingHashes`: An array of `bytes`, containing the node values necessary to recompute the Merkle root starting from `queryHashes`.
 
 #### Proof Construction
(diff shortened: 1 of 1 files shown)


proposals/lip-0031.md
```diff
@@ -84,6 +84,10 @@ The procedure for a proof-of-inclusion is as follows:
 
 In Fig. 1 we sketch a simple proof-of-inclusion for data block `data1`. The Verifier asks the Prover for a proof for `queryHashes=[leafHash(data1)]`. The Prover sends back `proof` with `size=5`, `idxs=[0001]`and `siblingHashes=[h0, h6, h4]`.
 
+### Representation of indices
+
+On an abstract level, indices in the array `idxs` are represented as bit strings. The length of such a bit string is [determined by the layer and the height of the tree](#proof-construction). For compactness, these bit strings are converted to `uint64` before being serialized such that they are [varint encoded](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0027.md#base-128-varints). As the indices can represent arbitrary nodes of the tree and not just leave nodes, different indices could represent the same integer. For example, `idxs` could contain the two indices `idx1 = "001"` and `idx2 = "0001"` (see [Fig. 2](#proof-of-inclusion)). Both would be converted to the unsigned integer `1`. To avoid this, a `1` is prepended to the bit string prior to converting to `uint64`. This way, `idx1` is converted to `0b1001 = 9` and `idx2` to `0b10001 = 17`.
```
ricott1, line 89:
```suggestion
On an abstract level, indices in the array `idxs` are represented as bit strings. The length of such a bit string is [determined by the layer and the height of the tree](#proof-construction). For compactness, these bit strings are converted to `uint64` before being serialized such that they are [varint encoded](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0027.md#base-128-varints). As the indices can represent arbitrary nodes of the tree and not just leaf nodes, different indices could represent the same integer. For example, `idxs` could contain the two indices `idx1 = "001"` and `idx2 = "0001"` (see [Fig. 2](#proof-of-inclusion)). Both would be converted to the unsigned integer `1`. To avoid this, a `1` is prepended to the bit string prior to converting to `uint64`. This way, `idx1` is converted to `0b1001 = 9` and `idx2` to `0b10001 = 17`.
```

proposals/lip-0031.md
```diff
@@ -191,7 +195,7 @@ proof = {
...
+* Each index in `idxs` represented as a bit string is prepended a `1` and encoded as a `uint64` (see [above](#representation-of-indices) for the reasoning). Whenever an index needs to be converted from its `unit64` representation to a binary string, the bit string following the most significant bit must be taken. Furthermore, we use the special index `0` to flag any element which is not part of the data array, and as such can be discarded during the verification. Notice that this does not constitute a proof-of-absence for that element.
 * The elements of `siblingHashes` are encoded as an array of `bytes` of length 32.
 
 The hex-encoded proof for the example in Fig. 1 is `proof=0x0805||0x120111||0x1a20||h0||0x1a20||h6||0x1a20||h4`, with a length `L=107` bytes.
```
gkoumout, line 201: I am not sure that this example is correct, but I might be missing something. In particular I don't understand the reason why the first two values are set like that.

gkoumout (changes requested):

proposals/lip-0031.md
AndreasKendziorra: I didn't check all values, but I would say that `0x0805` is the encoding of the size, and `0x120111` of `idxs` which is an array of length one.

proposals/lip-0031.md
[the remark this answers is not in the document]
AndreasKendziorra: I change the formulation, but in a slightly different way.

ricott1 (approved):

gkoumout (approved):

janhack (approved):
Editorial approval

PR #453, opened 2023-08-15
Let getAuthAccount return default values for non-existing account

### LINKED ISSUE #452: Make getAuthAccount() return default value instead of throwing
## Motivation

LIP specs and SDK implementation are currently different. 

## Current Specifications

[LIP-0041 getAuthAccount()](https://github.com/LiskHQ/lips/blob/8e7d079eaa182d280029251e25696a5f090a91fd/proposals/lip-0041.md#getauthaccount) specifies that the function should throw when the requested entry does not exist in the Auth Account store.

## Proposed Specifications

SDK implementation differs from the LIP specification, as it returns the default value instead of throwing:

```
{
    nonce: BigInt(0),
    numberOfSignatures: 0,
    mandatoryKeys: [],
    optionalKeys: [],
};
```

This simplifies the implementation, as there is no such use cases where SDK should handle a non-existing Auth Account entry other way than returning the default value.

As agreed with @shuse2, in this case it makes more sense to keep the implementation as it is, and align the LIP.

That would align behavior of `getAuthAccount()` with behavior of [`getAvailableBalance()`](https://github.com/LiskHQ/lips/blob/8e7d079eaa182d280029251e25696a5f090a91fd/proposals/lip-0051.md#getavailablebalance) and [`getEscrowedAmount()`](https://github.com/LiskHQ/lips/blob/8e7d079eaa182d280029251e25696a5f090a91fd/proposals/lip-0051.md#getescrowedamount), where LIP-0051 specifies that for a non-existing key the system should return `0` as the default value.

Besides the code samples, the [Store Notation section](https://github.com/LiskHQ/lips/blob/be768c4fda2df6c46caefe710688314f47d8ceb8/proposals/lip-0051.md#store-notation) might also need to be updated to reflect this.

## Affected LIPs

[LIP-0041: Introduce Auth module](https://github.com/LiskHQ/lips/blob/8e7d079eaa182d280029251e25696a5f090a91fd/proposals/lip-0041.md)

AndreasKendziorra:
Fixes #452

### DIFF
--- a/proposals/lip-0041.md
+++ b/proposals/lip-0041.md
@@ -408,7 +408,7 @@ def isMultisignatureAccount(address: Address) -> bool:
 
 #### getAuthAccount
 
-This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an account object with all properties set to their default.
+This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an account object with all properties set to their default values.
 
 ```python
 def getAuthAccount(address: Address) -> AuthAccount:
--- a/proposals/lip-0041.md
+++ b/proposals/lip-0041.md
@@ -408,12 +408,18 @@ def isMultisignatureAccount(address: Address) -> bool:
 
 #### getAuthAccount
 
-This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it throws an error.
+This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an account object with all properties set to their default.
 
 ```python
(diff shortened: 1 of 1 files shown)


bobanm (approved):

proposals/lip-0041.md
```diff
@@ -408,12 +408,18 @@ def isMultisignatureAccount(address: Address) -> bool:
 
 #### getAuthAccount
 
-This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it throws an error.
+This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an account object with all properties set to their default.
```
mjerkov, line 411: Tiny suggestion, almost irrelevant:
```suggestion
This function is used to retrieve information about the auth account corresponding to the input `address`. It returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an account object with all properties set to their default values.
```

mjerkov (approved):
Approved, with a minor suggestion.

PR #456, opened 2023-08-15
Add validation for keys and values in SMT update and remove functions

janhack:
In order to ensure the correct usages of the `update` and `remove` function, we add the following validations:

- Ensure that all involved keys have the same length `KEY_LENGTH_BYTES`.
- Ensure that all values added to the SMT are always non-empty.


shuse2 (approved):
looks good to me.
However, currently in the implementation, `update` and `remove` are done in the same batch in the same function and empty value represents `deleting` key

vjaiman (approved):
Looks good.

PR #460, opened 2023-08-16
Improve Governance module functionality

sergeyshemyakov:
Changes introduced in the PR:
1. Make quorum duration modifiable.
2. Add value validation for config update proposals. In particular, for quorum duration and quorum percentage values.
3. Enable token donation to the treasury by a simple transfer: before using treasury tokens all available balance is locked together with the minted locked tokens.
4. Move the `hasEnded` check from execution to verification of the create proposal command, delete Proposal Creation Failed event as it is not needed anymore.
5. Resolve https://github.com/LiskHQ/lips/issues/463.

### DIFF
--- a/proposals/lip-0074.md
+++ b/proposals/lip-0074.md
@@ -6,7 +6,7 @@ Discussions-To: https://research.lisk.com/t/introduce-governance-module/401
 Status: Draft
 Type: Standards Track
 Created: 2023-07-26
-Updated: 2023-08-15
+Updated: 2023-08-28
 Required: LIP 0048, 0051, 0057
 ```
 
--- a/proposals/lip-0074.md
+++ b/proposals/lip-0074.md
@@ -96,7 +96,7 @@ where turnout denotes the total number of PoS staking tokens that participated i
 
 ### Treasury
 
-The treasury is a dedicated account holding tokens that can be accessed by a funding proposal (see `ADDRESS_TREASURY` in the [table of constants](#notation-and-constants)). A funding proposal can only access tokens in the treasury if they are locked by the Governance module. Tokens could be minted into the treasury each block by the Governance module, the amount is specified by `treasuryTokensPerBlock` in [config overview](#config-overview).
+The treasury is a dedicated account holding tokens that can be accessed by a funding proposal (see `ADDRESS_TREASURY` in the [table of constants](#notation-and-constants)). A funding proposal can only access tokens in the treasury if they are locked by the Governance module. Tokens could be minted into the treasury each block by the Governance module, the amount is specified by `treasuryTokensPerBlock` in [config overview](#config-overview). Users can also donate tokens by simply sending them to the address of the treasury.
 
 ### Config substore
 
@@ -127,7 +127,7 @@ This LIP uses the following terms:
 | **Name**                  |   **Type**        |       **Validation**       |       **Description**                                 |
 |---------------------------|-------------------|-----------------------|-------------------------------------------------------|
 |   Proposal                |   object          | Must satisfy `proposalSchema`.| Object containing information of a particular proposal.|
-|   Address                 |   bytes           | Must have length `LENGTH_ADDRESS`.| Account address in Lisk ecosystem.              |
+|   Address                 |   bytes           | Must have length `ADDRESS_LENGTH`.| Account address in Lisk ecosystem.              |
 
 
 ### Notation and Constants
@@ -148,9 +148,9 @@ We define the following constants:
 |`MAX_MODULE_NAME_LENGTH`|`uint32`|32|The maximum length of a string specifying the name of a module, as defined in [LIP 0068][lip68Const].|
 |`MAX_PARAM_NAME_LENGTH`|`uint32`|64|The maximum length of a string specifying the name of a modifiable parameter.|
 |`MAX_PARAM_VALUE_LENGTH`|`uint32`|64|The maximum length of a byte array specifying the new value of a modifiable parameter.|
-|`LENGTH_PROPOSAL_ID`|`uint32`|4|The number of bytes of a proposal ID.|
-|`LENGTH_TOKEN_ID`|`uint32`|8|The number of bytes of a token ID.|
-|`LENGTH_ADDRESS`|`uint32`|20|The number of bytes of an address.|
+|`PROPOSAL_ID_LENGTH`|`uint32`|4|The number of bytes of a proposal ID.|
+|`TOKEN_ID_LENGTH`|`uint32`|8|The number of bytes of a token ID.|
+|`ADDRESS_LENGTH`|`uint32`|20|The number of bytes of an address.|
 |`MODULE_NAME_GOVERNANCE`|`string`|"governance"|Name of the Governance module.|
 |`SUBSTORE_PREFIX_PROPOSALS`|`bytes`|`0x0000`|Substore prefix of the proposals substore.|
 |`SUBSTORE_PREFIX_VOTES`|`bytes`|`0x4000`|Substore prefix of the votes substore.|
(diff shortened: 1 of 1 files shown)


proposals/lip-0074.md
```diff
@@ -920,6 +925,22 @@ def applyProposalOutcome(type: int, data: bytes) -> None:
         call updateConfig(dataObj.paramName, dataObj.value) method of module dataObj.moduleName
 ```
 
+#### validateConfigUpdate
+
+The function raises an exception if the proposed config update is invalid.
+
+```python
+def validateConfigUpdate(name: string, value: bytes):
+    type = type of value with name in configSchema
```
mjerkov, line 934: To me this looks more readable:
```suggestion
    type = string indicating the type of name property in configSchema
```

Or maybe this?
```suggestion
    type = configSchema["properties"][name]["type"]
```

But probably it doesn't need to be changed, because then it would need to be changed in `updateConfig`, too... just a thinking-aloud-kind-of-comment, feel free to resolve.

proposals/lip-0074.md
```diff
@@ -920,6 +925,22 @@ def applyProposalOutcome(type: int, data: bytes) -> None:
...
+    valueDecoded = validateAndDecode(value, type)
+    if name == "quorumPercentage":
+        if valueDecoded > 1000000:
+            raise Exception("Quorum percentage must be in parts-per-million")
+    if name == "quorumDuration":
+        if valueDecoded > configStore.voteDuration:
+            raise Exception("Quorum duration can not be greater than vote duration")
```
mjerkov, line 941: Should here be some `else` indicating what happens if `name` is none of the ones stated above?

proposals/lip-0074.md
```diff
@@ -904,6 +905,10 @@ def applyProposalOutcome(type: int, data: bytes) -> None:
     if type == PROPOSAL_TYPE_FUNDING:
         # assume that the data is valid
         dataObj = decode(fundingProposalDataSchema, data)
+        # lock donated treasury tokens together with minted treasury tokens
+        treasuryDonations = Token.getAvailableBalance(ADDRESS_TREASURY, configStore.tokenIdTreasury)
+        if (treasuryDonations != 0):
+            Token.lock(ADDRESS_TREASURY, MODULE_NAME_GOVERNANCE, configStore.tokenIdTreasury, treasuryDonations)
```
mjerkov, line 911: I see you decided to lock the donated tokens, so they become undistinguishable from the minted tokens. I can't think of any reason why this shouldn't be so, and it seems like a fine solution to me. 
I suppose adding a comment to the Treasury part of the Rationale stating that users can freely donate to treasury is unnecessary, right?

mjerkov (commented):
Looks good, only some minor comments.

janhack (commented):
I just have a minor suggestion, otherwise the PR looks good. Have you aligned the approach with introducing the `validateConfigUpdate` function with the devs? Maybe it would be good to have also a developer as reviewer of this change.

proposals/lip-0074.md
sergeyshemyakov: Agree, the current version is not very readable. I have changed to your first suggestion.

proposals/lip-0074.md
sergeyshemyakov: I have added a check that name is a valid config name, so now the function should be fine.

proposals/lip-0074.md
sergeyshemyakov: Have added a sentence to the rationale.

proposals/lip-0074.md
```diff
@@ -891,6 +880,7 @@ def validateProposal(type: int, data: bytes, description: dict) -> None:
     elif type == PROPOSAL_TYPE_CONFIG_UPDATE:
         proposalDataObj = decode(configUpdateProposalDataSchema, data)
         validateObjectSchema(configUpdateProposalDataSchema, proposalDataObj)
+        proposalDataObj.moduleName.validateConfigUpdate(proposalDataObj.paramName, proposalDataObj.value)
```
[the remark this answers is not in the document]
sergeyshemyakov: Changed as suggested.

proposals/lip-0074.md
```diff
@@ -454,8 +453,8 @@ configSchema = {
...
 |`tokenIdTreasury`|   `bytes`   |     No            | `TOKEN_ID_LSK`                   |Token ID of the tokens stored in the treasury and handled by funding proposals.|
-|`voteDuration`   | `uint32`    |       No          |   240000                         |Length of the vote period in blocks. As clarified in [Rationale](#voting-on-a-proposal), `voteDuration <= LOCKING_PERIOD_STAKING`, where `LOCKING_PERIOD_STAKING` is defined in [LIP 0057][posModule].|
-|`quorumDuration` | `uint32`    |       No          |   120000                         |Length of the quorum period in blocks. After this period the quorum is checked.|
+|`voteDuration`   | `uint32`    |       No          |   `241,920 = 28 * 24 * 3600 // BLOCK_TIME`                         |Length of the vote period in blocks. As clarified in [Rationale](#voting-on-a-proposal), `voteDuration <= LOCKING_PERIOD_STAKING`, where `LOCKING_PERIOD_STAKING` is defined in [LIP 0057][posModule].|
+|`quorumDuration` | `uint32`    |       Yes          |   `120,960 = 14 * 24 * 3600 // BLOCK_TIME`                         |Length of the quorum period in blocks. After this period the quorum is checked.|
```
bobanm, line 457: If the goal is to configure these durations based on time, not on number of block passed, it would be better to simply define them as `28 * 24 * 3600 // BLOCK_TIME` and `14 * 24 * 3600 // BLOCK_TIME` and remove the calculated values `241,920` and `120,960`.

The current specification is misleading IMO, because it assumes that the `BLOCK_TIME` is `10`, while in fact it is a configurable value.

bobanm (changes requested):
It would be also good to update the point 4 of PR description, to specify that Proposal Creation Failed event is removed, because the check now happens in `verify` instead of in `execute`.

mjerkov (approved):
Should these also be changed so `LENGTH` becomes a suffix (cc @bobanm)?

`MAX_LENGTH_PROPOSAL_TITLE`
`MAX_LENGTH_PROPOSAL_AUTHOR`
`MAX_LENGTH_PROPOSAL_SUMMARY`
`MAX_LENGTH_PROPOSAL_LINK`
`MAX_LENGTH_PROPOSAL_TEXT`

bobanm:
> Should these also be changed so `LENGTH` becomes a suffix (cc @bobanm)?
> 
> `MAX_LENGTH_PROPOSAL_TITLE` `MAX_LENGTH_PROPOSAL_AUTHOR` `MAX_LENGTH_PROPOSAL_SUMMARY` `MAX_LENGTH_PROPOSAL_LINK` `MAX_LENGTH_PROPOSAL_TEXT`

@mjerkov weirdly enough, everywhere in SDK code `MAX_LENGTH` is used as a prefix and if it's `LENGTH` then is a suffix. It means that this PR is consistent with the existing naming.

If needed, we could create a new PR to also move `MAX_LENGTH` to suffix everywhere in LIPs and code.

proposals/lip-0074.md
sergeyshemyakov: I think the current version is fine. The LIP suggests initial values for the Mainchain, so it is fair to assume that `BLOCK_TIME = 10`. Moreover I mention this parameter explicitly, so if in the future we decide to change `BLOCK_TIME` value, a simple search will also indicate that the Governance module parameters should be updated accordingly.

sergeyshemyakov:
I would keep `MAX_LENGTH_...` constants as they are now.

bobanm (approved):

janhack (approved):
Looks all good!

janhack (approved):
Editorial approval

PR #466, opened 2023-08-25
LIP 0053 - fix verifyValidatorsUpdate

### LINKED ISSUE #464: LIP 0053: Fix verifyValidatorsUpdate function
## Motivation

In CCU params there is the field `activeValidatorsUpdate` which is an object containing 3 items:
 - `blsKeysUpdate`: An array containing the BLS keys of the new active validators (ordered in lexicographic order)
 - `bftWeightsUpdate`: An array containing all new BFT weights. This includes the BFT weights of the new validators and the BFT weights of old validators whose weight changed since the last certicificate. If a validator is excluded from the active validator set, then assume their new bft weight is 0. 
 - `bftWeightsUpdateBitmap`: This bitmap encodes the correspondence between the BFT weights contained in the `bftWeightsUpdate` array and the actual validator (identified by their BLS keys) that each weight corresponds to. 

The correspondence works as follows (specified in [`getActiveValidatorsUpdate` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#computing-the-validators-update)). Let 
 - `newBLSKeys` be the keys of the new validators (i.e., the keys of the `blsKeysUpdate` array).
 - `oldBLSKeys` be the keys of the validators known in the chain until the last certificate. 
 -  `allBLSKeys = newBLSKeys U oldBLSKeys`, i.e., all BLS keys of previous and current validators, sorted in lexicographic order. 
 
Then we have that the `bftWeightsUpdateBitmap` contains `|allBLSKeys|` bits where the `i`th bit (from right to left) corresponds to the `i`th entry of `allBLSKeys`; if the `i`th bit is set to 1 it means that the BFT weight of validator with BLS key `allBLSKeys[i]` is included in the BFT weights update array. Therefore the number of `1`s in the bitmap should be equal to the size of `bftWeightsUpdate` array. 

**Validation.** Two of the checks performed in [` verifyValidatorsUpdate` function](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#verifyvalidatorsupdate) (part of CCU transaction verification) is that for all new validators:
1. The corresponding bitmap value is always set to `1` 
2. The bft weight is positive.

## Current Specifications

In the current specifications, the 2 aforementioned checks are performed in the following loop

```
    # All new keys must have a positive BFT weight.
    for idx, blsKey in enumerate(allBLSKeys):
        if blsKey in blsKeysUpdate:
            # Get digit of bitmap at index idx (starting from the right).
            digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
            if digit != 1:
                raise Exception("New validators must have a BFT weight update.")
            if bftWeightsUpdate[idx] == 0:
                raise Exception("New validators must have a positive BFT weight.")
```

Check (1) is done correctly. Check (2) is not. The reason is that we check `bftWeightsUpdate[idx]`, where `idx` is the index of `blsKey` in `allBLSKeys` array. So we are mixing up indices of two different arrays.  Note that in principle the size of `allKeys` might be quite larger than the size of `bftWeightUpdate`, in which case  even checking `bftWeightsUpdate[idx]` would give an error. 

## Proposed Specifications

We should be able to answer the following question: Which indices of `bftWeightsUpdate` array correspond to newBLSkeys? Then for each of those indices `i`, we should check that `bftWeightsUpdate[i] > 0` . 

Here is a proposed way to do that (assuming that check (1) has been already performed successfully):

```
newValitatorsIndices = []                 // indices of bftWeightsUpdate array corresponding to new validators
index = 0
for idx,  blsKey in enumerate(allBLSKeys):
    # Get digit of bitmap at index idx (starting from the right).
    digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
    if digit == 1:
        if blsKey in newBLSKeys:
            newValitatorsIndices.append(index)
        index+=1

for i in newValidatorsIndices:
    if bftWeightsUpdate[i] == 0:
        raise Exception('New validators must have a positive BFT weight.')
```

It would be also possible to merge this check with check (1) (but it would make it a bit harder to parse), as follows:

```
newValitatorsIndices = []                 // indices of bftWeightsUpdate array corresponding to new validators
index = 0
for idx,  blsKey in enumerate(allBLSKeys):
    # Get digit of bitmap at index idx (starting from the right).
    digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
    if blsKey in newBLSKeys:
        if digit != 1:
            raise Exception("New validators must have a BFT weight update.")
        newValitatorsIndices.append(index)
    if digit == 1:
        index+=1

for i in newValidatorsIndices:
    if bftWeightsUpdate[i] == 0:
        raise Exception('New validators must have a positive BFT weight.')
```

## Affected LIPs

0053.

gkoumout:
Fixes #464

### DIFF
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -210,8 +210,9 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
             # New validators must have a BFT weight update            
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
-            # All new validators must have a positive BFT weight.
-            if bftWeightsUpdate[index] == 0:         # index is the position in bftWeightsUpdate array where the BFT weight of current validator is specified
+            # New validators must have a positive BFT weight.
+            # The BFT weight of current validator is specified by bftWeightsUpdate[index]
+            if bftWeightsUpdate[index] == 0:         
                 raise Exception('New validators must have a positive BFT weight.')
 
         if digit == 1:
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/introduce-cross-chain-update-mechani
 Status: Draft
 Type: Standards Track
 Created: 2021-05-22
-Updated: 2023-08-17
+Updated: 2023-08-28
 Requires: 0045, 0049, 0058, 0061
 ```
 
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -207,11 +207,11 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
         digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
 
         if blsKey in blsKeysUpdate:
-            # New validators must have a BFT weight update            
+            # New validators must have a BFT weight update.
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
             # New validators must have a positive BFT weight.
-            # The BFT weight of current validator is specified by bftWeightsUpdate[index]
+            # The BFT weight of current validator is specified by bftWeightsUpdate[index].
             if bftWeightsUpdate[index] == 0:         
                 raise Exception('New validators must have a positive BFT weight.')
 
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -215,7 +215,7 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
                 raise Exception('New validators must have a positive BFT weight.')
 
         if digit == 1:
-            index+=1
+            index += 1
 
     # Update validator list.
     newActiveValidators = calculateNewActiveValidators(validators(partnerchainID).activeValidators, blsKeysUpdate, bftWeightsUpdate, bftWeightsUpdateBitmap)
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -200,8 +200,7 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
     if bin(int.from_bytes(bftWeightsUpdateBitmap, 'big')).count("1") != len(bftWeightsUpdate):
         raise Exception("The number of 1s in the bitmap is not equal to the number of new BFT weights.")
 
-    newValidatorsIndices = []                 // indices of bftWeightsUpdate array corresponding to new validators
-    index = 0
+    index = 0                          
 
     for idx, blsKey in enumerate(allBLSKeys):
         # Get digit of bitmap at index idx (starting from the right).
@@ -211,15 +210,12 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
             # New validators must have a BFT weight update            
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
-            newValidatorsIndices.append(index)
+            # All new validators must have a positive BFT weight.
+            if bftWeightsUpdate[index] == 0:         # index is the position in bftWeightsUpdate array where the BFT weight of current validator is specified
+                raise Exception('New validators must have a positive BFT weight.')
 
         if digit == 1:
             index+=1
-    
-    # All new validators must have a positive BFT weight.
-    for i in newValidatorsIndices:
-        if bftWeightsUpdate[i] == 0:
-            raise Exception('New validators must have a positive BFT weight.')
 
     # Update validator list.
     newActiveValidators = calculateNewActiveValidators(validators(partnerchainID).activeValidators, blsKeysUpdate, bftWeightsUpdate, bftWeightsUpdateBitmap)
--- a/proposals/lip-0053.md
+++ b/proposals/lip-0053.md
@@ -200,15 +200,26 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
     if bin(int.from_bytes(bftWeightsUpdateBitmap, 'big')).count("1") != len(bftWeightsUpdate):
         raise Exception("The number of 1s in the bitmap is not equal to the number of new BFT weights.")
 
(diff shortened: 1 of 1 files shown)


proposals/lip-0053.md
```diff
@@ -200,15 +200,26 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
     if bin(int.from_bytes(bftWeightsUpdateBitmap, 'big')).count("1") != len(bftWeightsUpdate):
         raise Exception("The number of 1s in the bitmap is not equal to the number of new BFT weights.")
 
-    # All new keys must have a positive BFT weight.
+    newValidatorsIndices = []                 // indices of bftWeightsUpdate array corresponding to new validators
+    index = 0
```
AndreasKendziorra, line 204: Maybe it becomes easier to read if we rename this one to `indexBftWeightsUpdate`?

AndreasKendziorra (approved):

proposals/lip-0053.md
ricott1: also we should use python style comments

ricott1: Also update the comment to be in python style

proposals/lip-0053.md
```diff
@@ -200,15 +200,26 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
...
-            digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
+            # New validators must have a BFT weight update            
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
-            if bftWeightsUpdate[idx] == 0:
-                raise Exception("New validators must have a positive BFT weight.")
+            newValidatorsIndices.append(index)
+
+        if digit == 1:
+            index+=1
```
ricott1, line 217:
```suggestion
            index += 1
```

proposals/lip-0053.md
```diff
@@ -200,15 +200,26 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
     if bin(int.from_bytes(bftWeightsUpdateBitmap, 'big')).count("1") != len(bftWeightsUpdate):
         raise Exception("The number of 1s in the bitmap is not equal to the number of new BFT weights.")
 
-    # All new keys must have a positive BFT weight.
+    newValidatorsIndices = []                 // indices of bftWeightsUpdate array corresponding to new validators
```
ricott1, line 203: I think we can avoid this extra array and check in place directly

proposals/lip-0053.md
```diff
@@ -200,15 +200,26 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
...
+
         if blsKey in blsKeysUpdate:
-            # Get digit of bitmap at index idx (starting from the right).
-            digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
+            # New validators must have a BFT weight update            
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
-            if bftWeightsUpdate[idx] == 0:
-                raise Exception("New validators must have a positive BFT weight.")
+            newValidatorsIndices.append(index)
```
ricott1, line 214: I think we can check here directly with 

```python
if bftWeightsUpdate[index] == 0:
    raise Exception('New validators must have a positive BFT weight.')
```

ricott1 (changes requested):
Some proposal to make the code cleaner

proposals/lip-0053.md
ricott1: Also update the comment to be in python style

gkoumout: Good point. This might make it less clear, but it is more efficient of course.

proposals/lip-0053.md
gkoumout: This array is now completely removed, so no need to name it ;-)

proposals/lip-0053.md
```diff
@@ -200,15 +200,23 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
...
         if blsKey in blsKeysUpdate:
-            # Get digit of bitmap at index idx (starting from the right).
-            digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
+            # New validators must have a BFT weight update            
             if digit != 1:
                 raise Exception("New validators must have a BFT weight update.")
-            if bftWeightsUpdate[idx] == 0:
-                raise Exception("New validators must have a positive BFT weight.")
+            # New validators must have a positive BFT weight.
+            # The BFT weight of current validator is specified by bftWeightsUpdate[index]
```
ricott1, line 214:
```suggestion
            # The BFT weight of current validator is specified by bftWeightsUpdate[index].
```

proposals/lip-0053.md
```diff
@@ -200,15 +200,23 @@ def verifyValidatorsUpdate(ccu: CCU) -> None:
...
+    index = 0                          
+
     for idx, blsKey in enumerate(allBLSKeys):
+        # Get digit of bitmap at index idx (starting from the right).
+        digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
+
         if blsKey in blsKeysUpdate:
-            # Get digit of bitmap at index idx (starting from the right).
-            digit = (int.from_bytes(bftWeightsUpdateBitmap, 'big') >> idx) & 1
+            # New validators must have a BFT weight update
```
ricott1, line 210:
```suggestion
            # New validators must have a BFT weight update.
```

ricott1 (approved):

AndreasKendziorra (approved):

ishantiw (approved):
LGTM, related SDK issue https://github.com/LiskHQ/lisk-sdk/issues/8849

janhack (approved):
Editorial approval

PR #475, opened 2023-09-04
Fix typo LIP 0058

gkoumout:
In function `setBFTParameters` there is the for loop:

```
    for validator in validators:
        if validator.bftWeight > 0:
            activeValidator = object with property bftWeight and blsKey
            activeValidator.bftWeight = validator.bftWeight
            activeValidator.blsKey =  validator.blsKey
            add activeValidator to activeValidators
```

where `validators` is not defined. Instead it should be replaced by `validatorList`. This PR fixes this typo.


janhack (approved):

janhack (approved):
Editorial approval

PR #480, opened 2023-09-07
Improve PoS bootstrap and hybrid period definitions

gkoumout:


proposals/lip-0057.md
```diff
@@ -72,6 +72,12 @@ By setting the BFT weights proportional to validator weight instead of uniform B
...
 
+### Bootstrap and Hybrid Period
+
+The bootstrap period, defined in [LIP 0034][lip-0034#bootstrap-period], is the initial period when the PoS validator selection mechanism is not active and blocks are generated by a fixed set of validators (specified in genesis block), called *initial validators*. This period should be long enough to prevent long range attacks, as explained in [LIP 0034][lip-0034#bootstrap-period]. In LIP 0034, we assumed that after the bootstrap period ends, then all validators are selected according to the PoS selection mechanism. One drawback of this approach is that after the end of the bootstrap period, the change of validators might be too abrupt, possibly leading to breaking the BFT security assumptions and finalizing contradicting blocks or inability of the chain to produce certificates. To avoid those issues, in LIP 34 we just did not allow blocks to be finalized (and therefore produce certificates) during the bootstrap period. This approach has some drawbacks, for example, it does not allow sidechains to be interoperable during their bootstrap period, which is not desirable.
```
janhack, line 77: It would be easier to make suggestions if you have one sentence per line.

Note that long range attacks are not an issue for this migration any more as we do not reset the votes and make people re-vote.

> To avoid those issues, in LIP 34 we just did not allow blocks to be finalized (and therefore produce certificates) during the bootstrap period.

I would remove "just" and the part about certificates as those do not exist in Lisk Core v3.

janhack (changes requested):
Looks good overall, I just have a few minor suggestions.

proposals/lip-0057.md
gkoumout: Rephrased, have a look.

proposals/lip-0061.md
```diff
@@ -112,14 +112,14 @@ The certificate generation specified in this LIP therefore generates an aggregat
 
 Note that if possible, also aggregate commits for blocks that are not at the end of the round are created and added to blocks so that a certificate is created as soon as a block is finalized and it is not required to wait for a change of BFT parameters.
 
-## Minimum Certificate Height
+### Minimum Certificate Height
 
 During the bootstrap period for Lisk Core v4, as defined in [LIP 63][lip-0063], the 101 fixed initial validators will not have valid registered BLS keys from the beginning.
```
AndreasKendziorra, line 117: I know it was not introduced here, but since we are cleaning up w.r.t. bootstrap period, we could set the link more specifically to https://github.com/LiskHQ/lips/blob/main/proposals/lip-0063.md#bootstrap-period.

proposals/lip-0057.md
```diff
@@ -72,6 +72,12 @@ By setting the BFT weights proportional to validator weight instead of uniform B
...
 
+### Bootstrap and Hybrid Period
+
+The bootstrap period, defined in [LIP 0034][lip-0034#bootstrap-period], is the initial period when the PoS validator selection mechanism is not active and blocks are generated by a fixed set of validators (specified in genesis block), called *initial validators*. In LIP 0034, we assumed that after the bootstrap period ends, then all validators are selected according to the PoS selection mechanism. One drawback of adopting this approach here is that after the end of the bootstrap period, the change of validators might be too abrupt, possibly leading to breaking the BFT security assumptions and finalizing contradicting blocks or inability of the chain to produce certificates. To avoid those issues, a possible solution would be to not allow blocks to be finalized during the bootstrap period, as in LIP 34. This approach has some drawbacks however, for example, it does not allow sidechains to be interoperable during their bootstrap period, which is not desirable.
```
AndreasKendziorra, line 77:
```suggestion
The bootstrap period, defined in [LIP 0034][lip-0034#bootstrap-period], is the initial period when the PoS validator selection mechanism is not active and blocks are generated by a fixed set of validators (specified in genesis block), called *initial validators*. In LIP 0034, we assumed that after the bootstrap period ends, all validators are selected according to the PoS selection mechanism. One drawback of adopting this approach here is that after the end of the bootstrap period, the change of validators might be too abrupt, possibly leading to breaking the BFT security assumptions and finalizing contradicting blocks or inability of the chain to produce certificates. To avoid those issues, a possible solution would be to not allow blocks to be finalized during the bootstrap period, as in LIP 34. This approach has some drawbacks however, for example, it does not allow sidechains to be interoperable during their bootstrap period, which is not desirable.
```

AndreasKendziorra (approved):

proposals/lip-0061.md
gkoumout: Good point. Done.

AndreasKendziorra (approved):

janhack (approved):

PR #484, opened 2023-09-13
LIP0039: handle case of invalid non-inclusion proof

ricott1:
For the verification protocol of a (non-)inclusion proof, we have the following scenarios:

1. proof.key == queryKey and proof.value != empty ==> inclusion proof
2. proof.key == queryKey and proof.value == empty ==> non-inclusion proof (empty node in the path)
3. proof.key != queryKey and proof.value != empty ==> non-inclusion proof (another leaf in the path)
4. proof.key != queryKey and proof.value == empty ==> invalid proof

We currently do not check case 4 correctly. This PR fixes this by explicitly checking that a proof for which the query key and the proof key differs, must have a non empty proof value (since it must correspond to an inclusion proof for another leaf node).


proposals/lip-0039.md
```diff
@@ -585,6 +585,11 @@ def verify(queryKeys: list[bytes], proof: Proof, root: bytes) -> bool:
             continue
 
         # q is an inclusion proof for another leaf node.
```
matjazv, line 587: Should be here in the comment `non-inclusion` proof?
If proof is inclusion then `key == query.key` and in line 585 it's already skipped to the beginning of the `for` loop.

ricott1: It's an inclusion proof for _another_ leaf node, hence a non-inclusion proof for the queried key.

matjazv (approved):

janhack (approved):
To better understand the general context, I would add a comment with the 4 cases before the verify function.

janhack (approved):
Looks all good.

janhack (approved):
Editorial approval

PR #486, opened 2023-09-21
Fix block rewards for less active validators

vjaiman:
The dynamic block rewards module calculates the block rewards assuming that the number of active validators in each round is exactly NUMBER_ACTIVE_VALIDATORS (e.g., 101 in the mainchain).

**Expected behavior:**

We should take care of the case that in current round the number of eligible validators are less than NUMBER_ACTIVE_VALIDATORS (e.g., 101 in the mainchain). This might very well happen after the next hard-fork since initially old validators will be banned until they register a BLS key. So a given round, might have much fewer validators.


shuse2 (approved):

proposals/lip-0071.md
```diff
@@ -28,20 +28,20 @@ In the current [Lisk protocol][lip-0042], the rewards are the same for all activ
 
 ### Minimum Reward for Active Validators
 
-This module assigns a minimum reward i.e. `minimumRewardActiveValidators` per round to all active validators to cover the minimal operational cost to run a validator. For the Lisk mainchain, the minimum reward is 10<sup>7</sup> beddows or `0.1 LSK`.
+This module assigns a minimum reward i.e. `getMinimumRewardActiveValidators(blockHeader.height)` per round to all active validators to cover the minimal operational cost to run a validator. The minimum reward is 10% of the total reward (`getNumberActiveValidators() * getDefaultRewardAtHeight(blockHeader.height)`) in a round at the current block height. For the Lisk mainchain (assuming `101` active validators), the minimum reward is 10<sup>7</sup> beddows or `0.1 LSK` at current block height.
```
gkoumout, line 31: I would erase the parenthesis (`getNumberActiveValidators() * getDefaultRewardAtHeight(blockHeader.height)`) because I think is a bit confusing.

gkoumout, line 31:
```suggestion
This module assigns a minimum reward to all active validators that generate a block to cover the minimal operational cost to run a validator. The minimum reward is a certain fraction (specified by the configurable constant `FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS`)  of the total reward in a round. For the Lisk Mainchain, this fraction is set to 10%.  Therefore, for heights with `getDefaultRewardAtHeight(height)` set to 1LSK,  the minimum reward is 10<sup>7</sup> beddows or `0.1 LSK`.
```

proposals/lip-0071.md
```diff
@@ -28,20 +28,20 @@ In the current [Lisk protocol][lip-0042], the rewards are the same for all activ
...
+* This module distributes the remaining stake rewards i.e. `getStakeRewardActiveValidators(blockHeader.height)` proportional to the validators weight. It is given to the active validators (per round) after giving the minimum reward. For the Lisk mainchain (assuming `101` active validators), the rewards are `90.9*10^8` beddows or `90.9 LSK`. This way, the module achieves fairness (higher weight implies higher rewards) while ensuring that rewards of low-weight validators do not go below the threshold of profitability.
```
gkoumout, line 35:
```suggestion
* This module distributes the remaining stake rewards i.e. `getStakeRewardActiveValidators(blockHeader.height)` proportional to the validators weight. It is given to the active validators that generate a block, on top of the minimum reward. For the Lisk mainchain, the rewards distributed according to stake are 90% of total rewards, therefore `90.9*10^8` beddows or `90.9 LSK` per round (assuming `101` active validators). This way, the module achieves fairness (higher weight implies higher rewards) while ensuring that rewards of low-weight validators do not go below the threshold of profitability.
```

proposals/lip-0071.md
```diff
@@ -28,20 +28,20 @@ In the current [Lisk protocol][lip-0042], the rewards are the same for all activ
...
...validator misses a block then a validator forging the block second time will get the reward proportional to their weight. This results in imbalancing the total reward in a round. In terms of inflation, currently in the Lisk mainchain, there is a maximum minting of `6*60*24*365 = 3153600` LSK per year. Every time a validator misses their block, 1 LSK less is minted per year. So the important aspect that we want to guarantee here is that in the dynamic rewards also never more than `3153600` LSK are minted per year so that there is a reliable upper bound on the inflation. In terms of incentives, it is also important that it should never be profitable for a validator if someone misses a block. Therefore, for missed blocks, validators forging blocks for the second time in the round, gets `getMinimumRewardActiveValidators(height)`. For the Lisk mainchain (assuming `101` active validators), the reward for missed blocks is 10<sup>7</sup> beddows or `0.1 LSK` at current block height. Since validators are already running full nodes and there is no additional effort for them to forge a missed block. There is a possibility that the minting of the tokens will be less if there are missed blocks.
```
gkoumout, line 44: I find the first three sentences a bit confusing and also with some phrasal typos (e.g., "in a round it if a validator misses a block". Perhaps we could improve a bit here as well, since we touched the text anyways.

gkoumout: Another example:

> 
If a validator misses a block then a validator forging the block second time will get the reward proportional to their weight. This results in imbalancing the total reward in a round.

If I understand well this does not actually happen. What we want to say here I guess is that "if we were granting dynamic block rewards even in the missed block case, then we would have this issue. But we don't right?

gkoumout: Also, perhaps I am missing something, but by reading the text I don't understand why we give them 0.1 and not 1. Perhaps we should explain the reasoning. I don't remember actually why we decided to do so.

proposals/lip-0071.md
```diff
@@ -70,6 +70,7 @@ We define the following constants:
 | **Configurable Constants** | | **Mainchain Value** | |
 | `BLOCK_TIME` | integer | 10 | Block time (in seconds) set in the chain configuration. |
 | `FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS` | uint32 | `1000` | It determines the percentage of rewards assigned equally to all active validators. The percentage can be obtained by dividing the value by `100`, i.e., a value of `1000` corresponds to `10%` .|
+| `DECIMAL_PERCENT_FACTOR` | uint32 | `10000` | The value is used to divide the `FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS` to obtain the minimum reward.|
```
gkoumout, line 73: Do we need this?

vjaiman: Yes. It is the value used in to calculate minimum reward.

proposals/lip-0071.md
vjaiman: Here it is the reasoning that if we give reward according to the weight for the missed blocks then it would imbalance the reward in a round.  
This line specifies this :  "If a validator misses a block then a validator forging the block second time will get the reward proportional to their weight. This results in imbalancing the total reward in a round." but actually we don't want this in the protocol and give minimum reward for the missed blocks.

vjaiman: The reasoning why we give 0.1 LSK is, we don't want the total reward to exceed and validators are already running full nodes, there is no additional effort for them to forge a missed block.

proposals/lip-0071.md
gkoumout: I mean, do we need a config constant? Couldn't we just use 10000 fixed? (meaning the factor would have two decimal point precision)

proposals/lip-0071.md
```diff
@@ -239,19 +260,20 @@ This function is used to retrieve the block reward for a validator.
 
 ```python
 def getValidatorBlockReward(blockHeader: BlockHeader) -> uint64:
+    height = blockHeader.height
```
gkoumout, line 263: Suggestion after looking into this more carefully: We could define here `minimumRewardActiveValidators = getMinimumRewardActiveValidators(height)` and then use the constant `minimumRewardActiveValidators` whenever needed (that was my bad to completely remove it).

proposals/lip-0071.md
vjaiman: Yes, we can but I would keep it to have an alignment with sdk https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/dynamic_rewards/constants.ts. Plus, it is always good to have it configurable rather than fixed constant value.

gkoumout: I think there is some confusion here. 

1. In the implementation it is not a config constant, it is just a constant, fixed (you can see that is not part of defaultConfig or [moduleConfig](https://github.com/LiskHQ/lisk-sdk/blob/2d04e23b5b172a7b84cdad2669b985a51156e671/framework/src/modules/dynamic_rewards/types.ts#L20C18-L20C30)). So they just hard-coded the value 10000 to a non-config constant. Recall that we recently aligned with them on config constants, so it would be scary if we had a mis-alignment again.
2. In general, I think it would not be good to have the denominator as a configurable constant, since you then need extra checks that it should be at least the value of the numerator and potentially create errors from the user side. So it is fine I guess to just fix it to 10000, meaning percentage with double decimal precision, it's very clear.

vjaiman: Updated.

janhack (changes requested):
I added some remarks regarding the rationale.

proposals/lip-0071.md
```diff
@@ -281,18 +303,18 @@ This function is used to estimate the expected rewards for a validator assuming
...
     let validator be the item of validatorParams.validators with item.address == validatorAddress
     if not validator:
         return (0, 0, 0, 0) 
 
-    if validator.bftWeight > 0:
-        blockReward = minimumRewardActiveValidators + (validator.bftWeight * stakeRewardActiveValidators)//totalBftWeight 
-    else:
-        blockReward = defaultRewardStandByValidators
+    if validator.bftWeight > 0: #active validator
+        blockReward = getMinimumRewardActiveValidators(height) + (validator.bftWeight * getStakeRewardActiveValidators(height))//totalBftWeight
```
sergeyshemyakov, line 315: This reward payout does not guarantee that no more than 101 LSK are paid as rewards in a round. I think we have to strictly cap the inflation, so the reward payout should be changed.
I propose: the first time an active validator forges a block they receive a value proportional to their weight, all other blocks forged in a round give only minimal reward. We should discuss it further.

sergeyshemyakov (changes requested):
I have not commented on the rationale since we have not aligned on the technical solution yet. I will do it in the later stages of review.

proposals/lip-0071.md
```diff
@@ -281,18 +303,18 @@ This function is used to estimate the expected rewards for a validator assuming
 ##### Execution
 
 ```python
-def getExpectedValidatorRewards(validatorAddress: Address) -> tuple[uint64, uint64, uint64, uint64]:
+def getExpectedValidatorRewards(validatorAddress: Address, height: uint32) -> tuple[uint64, uint64, uint64, uint64]:
```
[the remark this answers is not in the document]
gkoumout: This function is defined so that it can be used in the wallet, to give an estimation on rewards. Therefore, we need an interface for Lisk Desktop to request the expected rewards.  Since `getDefaultRewardAtHeight` depends on the height, it should be correctly called by the wallet depending on the current height.

proposals/lip-0071.md
sergeyshemyakov: After discussions we decided to not change anything for implementational simplicity.

proposals/lip-0071.md
```diff
@@ -28,20 +28,20 @@ In the current [Lisk protocol][lip-0042], the rewards are the same for all activ
...
+* In principle, if percentage of rewards shared equally among active validators is set to `0%` i.e.`FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS == 0`, then it will be a proof-of-stake system where rewards are proportional to the validator weight. On the other hand, if the percentage of rewards shared equally among active validators is set to `100%` i.e. `FACTOR_MINIMUM_REWARD_ACTIVE_VALIDATORS == 10000`, it will be the previous proof-of-stake (PoS) system where each validator received the same reward for forging a block irrespective of their weights. Therefore, this reward system can be seen as a generalization of the previous one and allows flexibility to developers to use the variant of Proof-of-Stake they find more appropriate for their use case.
```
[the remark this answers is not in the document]
janhack: The comma is not fixed. In general, it would make the review process easier if you accept suggestions on GitHub (can also be done in one batch) instead of doing the same change again yourself.

proposals/lip-0071.md
```diff
@@ -20,28 +20,29 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
+Currently, in the Lisk protocol if some validator misses a block and another validator generates the block, then the generator gets a minimal reward at current block height (`0.1 LSK` for the mainchain). This is done to limit the total rewards in a round if an active validator misses a block. Also it should never be profitable for a validator if someone misses a block.
```
janhack, line 45:
```suggestion
In the new protocol proposed in this LIP, if some validator misses a block and another validator generates the block, then the generator gets a minimal reward at current block height (`0.1 LSK` for the mainchain). This is done to limit the total rewards in a round if an active validator misses a block. Also it should never be profitable for a validator if someone misses a block.
```

proposals/lip-0071.md
```diff
@@ -303,21 +304,22 @@ This function is used to estimate the expected rewards for a validator assuming
 ##### Execution
 
 ```python
-def getExpectedValidatorRewards(validatorAddress: Address, height: uint32) -> tuple[uint64, uint64, uint64, uint64]:
+def getExpectedValidatorRewards(validatorAddress: Address) -> tuple[uint64, uint64, uint64, uint64]:
     validatorParams = Validators.getValidatorParams()
+    height = current height of the last block
```
janhack, line 309: Why do you remove the height here? It seems to be used in line 317 below.

janhack (changes requested):
I just have a few minor points left, one also from the previous review.

proposals/lip-0071.md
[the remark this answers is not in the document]
sergeyshemyakov: Fixed.
Maybe I should try this gh tool, I just thought that gathering changes from the web UI and locally would be more complicated than doing all in a single commit.

proposals/lip-0071.md
sergeyshemyakov: I add it here. In the previous version of PR the height was a parameter of the function, I have moved it into the function.

janhack (approved):
Looks good.

gkoumout (changes requested):

proposals/lip-0071.md
```diff
@@ -20,28 +20,29 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
+In the current [Lisk protocol][lip-0042], the rewards are the same for all active and standby validators (selected according to the selection mechanism described in [LIP 0022][lip-0022] and [LIP 0023][lip-0023]) who generate a block in a round, irrespective of their validator weight. In the current scenario, the rewards for generating a block are fixed and therefore a validator with a large validator weight does not have incentives to lock further tokens via self-stakes unless they run multiple validators, which is not desirable for them due to operational costs and maintenance efforts. Thus, the LIP aims to simplify PoS participation for validators to lock the large number of tokens. By more participation, we expect to increase the Total Value Locked (TVL) and therefore, the security of the Lisk Mainchain. To achieve this, we assign block rewards to also depend on the weight of the validator generating the block.
```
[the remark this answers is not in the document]
sergeyshemyakov: Changed as suggested

proposals/lip-0071.md
```diff
@@ -20,28 +20,29 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
 
 ## Rationale
 
-​​The main purpose of Dynamic Reward module is to compute the block rewards according to the validator weight for active validators and induces specific logic after the application of a block:
+​​The main purpose of the Dynamic Reward module is to compute the block rewards according to the validator weight for active validators and induces specific logic after the application of a block:
```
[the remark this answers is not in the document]
sergeyshemyakov: Changed as suggested

proposals/lip-0071.md
```diff
@@ -20,28 +20,29 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
+* Technically, each time an active validator generates a block they get a fraction of stake reward (depending on the current number of active validators), proportionally to their BFT weight. As a consequence, if active validators generate several blocks in a round then the total reward paid out in a round could slightly exceed the estimate of `defaultReward * roundLength`. This could happen if there are less than 101 active validators on the Lisk mainchain and validators with higher than average weight generate more blocks, however this should not be a persistent problem in practice since new active validators will appear.
```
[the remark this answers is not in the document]
sergeyshemyakov: I put it to mention explicitly that we do not guarantee inflation smaller than 103 LSK / round and to mention the technical details of the block rewards computation.

proposals/lip-0071.md
sergeyshemyakov: I keep the endpoint format consistent with how I defined endpoints for DEX. Many of them also required current height as an argument, see e.g. https://github.com/LiskHQ/lisk-dex-specs/blob/main/specifications/0001.md?plain=1#L1168.
I think this function signature is better since we never explicitly mention any of the data that is passed in the context. We just always assume that we can access it in the function itself. This is why I leave the function as is.

proposals/lip-0071.md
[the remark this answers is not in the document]
sergeyshemyakov: Good suggestion, changed as you propose.

proposals/lip-0071.md
gkoumout: Ok, I will not insist, fine for me if @janhack agrees as well..

proposals/lip-0071.md
```diff
@@ -20,28 +20,29 @@ This LIP is licensed under the [Creative Commons Zero 1.0 Universal](https://cre
...
+* Technically, each time an active validator generates a block they get a fraction of stake reward (depending on the current number of active validators), proportionally to their BFT weight. Note that an active validator might generate multiple blocks in a round (assuming no missed blocks): this could happen in case the total number of validators of a round are less than the expected one (which is `roundLength`). In such cases, the total reward paid out in a round could slightly exceed the estimate of `defaultReward * roundLength`, in case validators with higher than average weight generate more blocks than the others. However, since in our PoS systems the validators are randomly shuffled, the expected value of the total reward given in a round is not higher than `defaultReward * roundLength`.
```
gkoumout, line 36: To be 100% precise, perhaps we should replace `defaultReward` by `getDefaultRewardAtHeight()` and `roundLength` by `Pos.getRoundLength()`. But since we are in the rationale, I am not sure this is 100% necessary.

gkoumout (approved):
Approved with a minor cosmetic comment on the rationale.

janhack (approved):
Editorial approval

PR #489, opened 2023-09-25
Introduce function to set thresholds in PoS module

gkoumout:
The purpose of this PR is to refactor the PoS module so that there exists a single function `computeThresholds` to determine the precommit and certificate thresholds required for BFT and interoperability respectively. 

This way, at all stages when those thresholds need to get updated, this would be done via calling `computeThresholds`. As a consequence, sidechain developers could modify the logic on computing those thresholds by just modifying this function instead of having to check all places where those thresholds are updated.


ricott1 (approved):

proposals/lip-0057.md
```diff
@@ -1464,6 +1466,22 @@ x <= (capValue / 10000) ( i * x + sum( [validatorList[k].bftWeight for k in rang
 Here `x` is the BFT weight at which the elements `validatorList[0].bftWeight, ..., validatorList[i-1].bftWeight` are capped. Solving the above equation for `x` yields the value of `cappedWeightRemainingElements` in the code. Note that if
 `cappedWeightRemainingElements > validatorList[i-1].bftWeight`, then you do not need to cap `validatorList[i-1].bftWeight` and can reduce the index `i` by 1.
 
+#### computeThresholds
+
+This function sets the precommit and certificate thresholds that are forwarded to the consensus layer, see [LIP 0058][lip-0058] for details.
```
janhack, line 1471:
```suggestion
This function computes the precommit and certificate thresholds that are forwarded to the consensus layer, see [LIP 0058][lip-0058] for details. 
```

proposals/lip-0057.md
```diff
@@ -66,7 +66,9 @@ This part of the state store is used to maintain the timestamp of the last block
...
+
+Note that further we choose `floor(2/3 * aggregateBFTWeight) + 1` as precommit and certificate threshold for PoS chains, see [LIP 0058][lip-0058] for an explanation of these parameters. Here `aggregateBFTWeight` is the sum of BFT weights of the active validators. This choice of BFT parameters means that the safety and liveness of the BFT consensus protocol is ensured if `<1/3 * aggregateBFTWeight` is malicious (for the case of changing validators the guarantees are slightly weaker). Developers might choose to modify the threshold values for their sidechains by updating the [`computeThresholds` function](#computethresholds). We crucially note however, that modifying the chosen parameters has consequences for the security guarantees of the chain according to the security-liveness tradeoff proven in Theorem 4.4 of the [Lisk-BFT paper][lisk-bft-paper]. Therefore, we highly recommend not updating those parameters, unless the tradeoff is well-understood and there is some very clear reason to do so for a specific chain.
```
janhack, line 71:
```suggestion
Note that further we choose `floor(2/3 * aggregateBFTWeight) + 1` as precommit and certificate threshold for PoS chains, see [LIP 0058][lip-0058] for an explanation of these parameters. Here `aggregateBFTWeight` is the sum of BFT weights of the active validators. This choice of BFT parameters means that the safety and liveness of the BFT consensus protocol is ensured if `<1/3 * aggregateBFTWeight` is malicious (for the case of changing validators the guarantees are slightly weaker). Developers might choose to modify the threshold values for their sidechains by updating the [`computeThresholds` function](#computethresholds). We crucially note, however, that modifying the chosen parameters has consequences for the security guarantees of the chain according to the security-liveness tradeoff proven in Theorem 4.4 of the [Lisk-BFT paper][lisk-bft-paper]. Therefore, we highly recommend not updating those parameters, unless the tradeoff is well-understood and there is some very clear reason to do so for a specific chain.
```

janhack (changes requested):
Looks good. I just have two cosmetic remarks.

janhack (approved):
Editorial approval

PR #492, opened 2023-09-27
Adjust aggregateCommit.height in genesis block and getMaxRemovalHeight() function

### LINKED ISSUE #491: Performance issue in single commit creation
## Motivation

If a validator node starts for the first time or loses all previously created single commits due to a restart, the node should automatically re-create all recent single commits from height `h1 = getMaxRemovalHeight()` up to height `h2 = getBFTHeights().maxHeightPrecommitted`, see [LIP 0061](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#initial-single-commit-creation). As for the genesis block, the property `b.header.aggregateCommit.height` is `0` (see [LIP 0060](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0060.md#validation-2)), but after the regenesis, the height of the genesis block is a large number >>0,  the note attempts to create millions of single commits for blocks before the genesis height.

## Current Specifications

- The property `b.header.aggregateCommit.height` of the genesis block is `0`.
- The function `getMaxRemovalHeight` can return heights smaller than the genesis height.

## Proposed Specifications

- The property `b.header.aggregateCommit.height` of the genesis block is set to the genesis height.
- The function `getMaxRemovalHeight` never returns a height smaller than the genesis height.

## Affected LIPs

0060, 0061

janhack:
Fixes #491

### DIFF
--- a/proposals/lip-0060.md
+++ b/proposals/lip-0060.md
@@ -7,7 +7,7 @@ Discussions-To: https://research.lisk.com/t/update-genesis-block-schema-and-proc
 Status: Draft
 Type: Standards Track
 Created: 2021-10-08
-Updated: 2023-05-22
+Updated: 2023-09-28
 Requires: 0040, 0058, 0055
 ```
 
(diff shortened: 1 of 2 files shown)


ricott1 (approved):

shuse2:
https://github.com/LiskHQ/lisk-sdk/pull/9048

PR #493, opened 2023-09-29
Create single commits only from MIN_CERTIFICATE_HEIGHT height onwards

janhack:
Certificates are only be generated for blocks of height at least `MIN_CERTIFICATE_HEIGHT`. However, in the current certificate generation mechanism single commits are created from `genesisHeight + 1` onwards leading to uncessary storage and memory usage if `genesisHeight + 1 << MIN_CERTIFICATE_HEIGHT`. This PR changes the single commit creation in LIP 61 to only create them from height `MIN_CERTIFICATE_HEIGHT` onwards. 

Related Lisk SDK issue: https://github.com/LiskHQ/lisk-sdk/issues/9057

### DIFF
--- a/proposals/lip-0061.md
+++ b/proposals/lip-0061.md
@@ -405,7 +405,7 @@ The height up to which single commits can be removed.
 ```python
 def getMaxRemovalHeight() -> uint32:
     b = block at height maxHeightFinalized
-    return max(b.header.aggregateCommit.height, genesisHeight)
+    return max(b.header.aggregateCommit.height, MIN_CERTIFICATE_HEIGHT - 1)
 ```
 
 In the function above, `maxHeightFinalized` is the maximum height up to which the current chain is considered final, as introduced in [LIP 0056](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0056.md#finalized-height), and `genesisHeight` is the height of the genesis block.
@@ -418,8 +418,8 @@ If a validator node starts for the first time or loses all previously created si
 
 If for a validator `v`, after applying a block the value returned by the function `getBFTHeights().maxHeightPrecommitted` increases from `h1` to `h2`, then the following commit messages are created by the validator `v` and gossiped as described further below:
 
-* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters), validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
-* If `existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters), `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.
+* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters) and `h >= MIN_CERTIFICATE_HEIGHT`, validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
+* If `existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters) and `h >= MIN_CERTIFICATE_HEIGHT`, `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.
 
 #### Single Commit P2P Gossip
 
--- a/proposals/lip-0061.md
+++ b/proposals/lip-0061.md
@@ -6,7 +6,7 @@ Discussions-To: https://research.lisk.com/t/introduce-certificate-generation-mec
 Status: Draft
 Type: Standards Track
 Created: 2021-05-22
-Updated: 2023-09-28
+Updated: 2023-10-04
 Requires: 0044, 0055, 0058
 ```
 
--- a/proposals/lip-0061.md
+++ b/proposals/lip-0061.md
@@ -408,18 +408,18 @@ def getMaxRemovalHeight() -> uint32:
     return max(b.header.aggregateCommit.height, MIN_CERTIFICATE_HEIGHT - 1)
 ```
 
-In the function above, `maxHeightFinalized` is the maximum height up to which the current chain is considered final, as introduced in [LIP 0056](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0056.md#finalized-height), and `genesisHeight` is the height of the genesis block.
+In the function above, `maxHeightFinalized` is the maximum height up to which the current chain is considered final, as introduced in [LIP 0056](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0056.md#finalized-height).
 
 #### Initial Single Commit Creation
 
-If a validator node starts for the first time or loses all previously created single commits due to a restart, the node should automatically re-create all recent single commits, which may be essential for the network to create aggregate commits. Let `h1 = getMaxRemovalHeight()` and `h2 = getBFTHeights().maxHeightPrecommitted`. Then the single commits are created as described in the next section [Creation After Block Processing](#creation-after-block-processing) for these values of `h1` and `h2`.
+If a validator node starts for the first time or loses all previously created single commits due to a restart, the node should automatically re-create all recent single commits, which may be essential for the network to create aggregate commits. Let `h1 = getMaxRemovalHeight()` and `h2 = getBFTHeights().maxHeightPrecommitted`. If `h2 > h1`, then the single commits are created as described in the next section [Creation After Block Processing](#creation-after-block-processing) for these values of `h1` and `h2`.
 
 #### Creation After Block Processing
 
 If for a validator `v`, after applying a block the value returned by the function `getBFTHeights().maxHeightPrecommitted` increases from `h1` to `h2`, then the following commit messages are created by the validator `v` and gossiped as described further below:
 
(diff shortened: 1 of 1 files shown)


AnanyaSshri (approved):

proposals/lip-0061.md
```diff
@@ -418,8 +418,8 @@ If a validator node starts for the first time or loses all previously created si
 
 If for a validator `v`, after applying a block the value returned by the function `getBFTHeights().maxHeightPrecommitted` increases from `h1` to `h2`, then the following commit messages are created by the validator `v` and gossiped as described further below:
 
-* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters), validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
-* If `existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters), `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.
+* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters) and `h >= MIN_CERTIFICATE_HEIGHT`, validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
```
gkoumout, line 421: I guess that adding the condition `h >= MIN_CERTIFICATE_HEIGHT` is redundant, since 

`h >= h1+1 = getMaxRemovalHeight()+1 >=  MIN_CERTIFICATE_HEIGHT - 1 +1 = MIN_CERTIFICATE_HEIGHT `

proposals/lip-0061.md
```diff
@@ -405,7 +405,7 @@ The height up to which single commits can be removed.
 ```python
 def getMaxRemovalHeight() -> uint32:
     b = block at height maxHeightFinalized
-    return max(b.header.aggregateCommit.height, genesisHeight)
+    return max(b.header.aggregateCommit.height, MIN_CERTIFICATE_HEIGHT - 1)
 ```
 
 In the function above, `maxHeightFinalized` is the maximum height up to which the current chain is considered final, as introduced in [LIP 0056](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0056.md#finalized-height), and `genesisHeight` is the height of the genesis block.
```
gkoumout, line 411: Comment about line 415: We implicitly assume that `h2 > = h1`. This was true in the previous setting but not anymore. We could just say that if `h1 > h2` we just don't create single commits at all I guess..

gkoumout (commented):
Left some comments.

proposals/lip-0061.md
AndreasKendziorra: > and `genesisHeight` is the height of the genesis block.

This part should be removed or replaced by a comment for `MIN_CERTIFICATE_HEIGHT`

proposals/lip-0061.md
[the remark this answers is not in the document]
AndreasKendziorra: I think this is not true in general. The definition `h1 := getMaxRemovalHeight()` holds for _Initial Single Commit Creation_ but not for _Creation After Block Processing_.

AndreasKendziorra (changes requested):

proposals/lip-0061.md
gkoumout: Good point. I guess it should be removed, since `MIN_CERTIFICATE_HEIGHT` is well-defined

proposals/lip-0061.md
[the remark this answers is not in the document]
gkoumout: Ok I got confused here. I agree now, can be resolved.

proposals/lip-0061.md
```diff
@@ -418,8 +418,8 @@ If a validator node starts for the first time or loses all previously created si
 
 If for a validator `v`, after applying a block the value returned by the function `getBFTHeights().maxHeightPrecommitted` increases from `h1` to `h2`, then the following commit messages are created by the validator `v` and gossiped as described further below:
```
gkoumout, line 419: Here I would define `h1 = max(getMaxRemovalHeight(), previous maxHeightPrecommitted)` for the following reasons:

1) Efficiency: Since we don't want to allow commits for height smaller than `MIN_CERTIFICATE_HEIGHT`, it would be really waste of resources to loop through all blocks of init rounds (roughly 60k blocks) with no reason. 
2) Clarity: It will make the whole part (together with initial single commit creation) much cleaner and easier to parse.

proposals/lip-0061.md
```diff
@@ -418,8 +418,8 @@ If a validator node starts for the first time or loses all previously created si
...
 
-* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters), validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
-* If `existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters), `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.
+* For every height `h` in `{h1+1, h1+2, ..., h2}` such that `existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters) and `h >= MIN_CERTIFICATE_HEIGHT`, validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
+* If `existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters) and `h >= MIN_CERTIFICATE_HEIGHT`, `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.
```
gkoumout, line 423: Here I guess the intended behavior is (assuming `h1 = max(getMaxRemovalHeight(), previous maxHeightPrecommitted)`):

 - For `h` in `[h1+1, h2-1]` , create single commits if `existBFTParameters(h+1)` returns `True` and validator `v` was active at height `h`.
 - For height `h = h2`, create single commit if validator `v` was active at height `h` (no matter the value of `existBFTParameters(h+1)`).

If that's the case I would highly recommend rephrasing like this. As it is now it is quite harder to parse and to verify correctness of the implementation.

gkoumout: Note that if we indeed define `h1 = max(getMaxRemovalHeight(), previous maxHeightPrecommitted)` , then the condition  `h >= MIN_CERTIFICATE_HEIGHT` becomes redundant.

gkoumout (changes requested):

proposals/lip-0061.md
janhack: - I removed the part "and genesisHeight is the height of the genesis block.".
- I added the requirement that `h2 > h1` (note that this is different to what you proposed @gkoumout )
- I did not swap the sections to keep the diff simple and to follow the philosophy now to only change what is strictly required.

proposals/lip-0061.md
janhack: Note that you can equivalently do `h1 = max(MIN_CERTIFICATE_HEIGHT - 1, previous maxHeightPrecommitted)`. This avoid calling `getMaxRemovalHeight()` and additional state reads. I updated the specs in this direction.

proposals/lip-0061.md
AndreasKendziorra: Can be resolved

AndreasKendziorra (approved):

proposals/lip-0061.md
gkoumout: Fine. Just to make sure the reasoning, is this because `previous maxHeightPrecommitted >= b.header.aggregateCommit.height` should always hold, right?

proposals/lip-0061.md
gkoumout: @janhack Could you clarify the difference between what I proposed and what you added? I fail to see it. 

Also, swapping of the sections was proposed to avoid mis-conceptions of the paragraph, since you essentially run the same "algorithm" for a different interval [h1,h2], so it would be easier to parse if we first see the description and then the modification. At least I got confused while reviewing the implementation due to this. 

But ok, since it is ok now, I agree it is an optional change.

janhack: You proposed `h2 > = h1` or (alternatively `not h1 > h2`), while I now wrote `h2 > h1`, i.e., strict inequality.

As mentioned above, I would keep the order of sections as is.

proposals/lip-0061.md
janhack: In general, for every block we have `maxHeightPrecommited (after processing block) >= b.header.aggregateCommit.height`. Further, we have that the finalized height is strictly lower than the finalized height. Hence,
```
maxHeightPrecommitted at currentHeight -1 >= b.header.aggregateCommit.height at currentHeight -1 >= b.header.aggregateCommit.height at maxHeightFinalized
```
Therefore,
```
 max(MIN_CERTIFICATE_HEIGHT - 1, maxHeightPrecommitted at currentHeight -1)  >= getMaxRemovalHeight()
```

ishantiw (approved):

proposals/lip-0061.md
gkoumout: Ok, this is exactly what I meant by `previous maxHeightPrecommitted >= b.header.aggregateCommit.height` (meaning b at maxHeightFinalized as in the `getMaxRemovalHeight` function).

Thanks for the clarification!

gkoumout (approved):

PR #494, opened 2023-10-06
Update LIP 0052 - Swap checks in verify transfer functions

mjerkov:
Currently, in [`verifyTransferInteral`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0052.md#verifytransferinternal) and [`verifyTransferCrossChainInternal`](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0052.md#verifytransferinternal) functions of LIP 0052, the check `owner != senderAddress` happens before the check `isNFTEscrowed`:

```
if owner != senderAddress:
    if emitEvent:
        emitFailedTransferEvent(senderAddress, recipientAddress, nftID, RESULT_INITIATED_BY_NONOWNER)
    raise Exception("Transfer not initiated by the NFT owner")

if isNFTEscrowed(nft):
    if emitEvent:
        emitFailedTransferEvent(senderAddress, recipientAddress, nftID, RESULT_NFT_ESCROWED)
    raise Exception("NFT is escrowed to another chain")
```
However, in case when NFT is escrowed to another chain, the first check would always fail because in that case the NFT owner is `chainID`, so it cannot possibly be `senderAddress`. Therefore, it makes sense to swap these checks, i.e. to check `isNFTEscrowed` before `owner != senderAddress`.

### DIFF
--- a/proposals/lip-0052.md
+++ b/proposals/lip-0052.md
@@ -1304,15 +1304,15 @@ def verifyTransferInternal(
 
     owner = nft["owner"]
 
-    if owner != senderAddress:
-        if emitEvent:
-            emitFailedTransferEvent(senderAddress, recipientAddress, nftID, RESULT_INITIATED_BY_NONOWNER)
-        raise Exception("Transfer not initiated by the NFT owner")
-
     if isNFTEscrowed(nft):
         if emitEvent:
             emitFailedTransferEvent(senderAddress, recipientAddress, nftID, RESULT_NFT_ESCROWED)
         raise Exception("NFT is escrowed to another chain")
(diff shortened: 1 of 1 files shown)


sergeyshemyakov (changes requested):
There is a similar behavior in the [destroy](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0052.md#destroy-2) function that should also be changed.

Incede (approved):

sergeyshemyakov (approved):

janhack (approved):
Editorial approval

PR #496, opened 2023-10-11
Set aggregate commit height to 0 for genesis block

### LINKED ISSUE #495: Revert aggregateCommitHeight in genesis block to 0
## Motivation

In a recent issue, we addressed [performance in single commit creation](https://github.com/LiskHQ/lips/issues/491). The goal was to not create single commits for heights smaller than the (re)genesis height. One of the actions taken was to set `b.header.aggregateCommit.height = b.header.height` for the genesis block; the previous value was `b.header.aggregateCommit.height = 0`. 

Since this change was proposed after the testnet migration, this makes the testnet genesis block invalid according to the new protocol.

## Current Specifications

At the time of migration, for the genesis block `b`, `b.header.aggregateCommit.height = 0`. By setting in the specs that `b.header.aggregateCommit.height = b.header.height` for the genesis block, makes the current genesis block invalid. 

## Proposed Specifications

Revert this change and reset `b.header.aggregateCommit.height = 0` for the genesis block `b`. Note that this will not affect the single commit creation, since (after a [subsequent edit](https://github.com/LiskHQ/lips/pull/493)) for any block `b`, single commits are created after `getMaxRemovalHeight() = max(b.header.aggregateCommit.height, MIN_CERTIFICATE_HEIGHT - 1) >= genesisHeight`, since `MIN_CERTIFICATE_HEIGHT >= genesisHeight +1`.   

## Affected LIPs

LIP 0060.

AndreasKendziorra:
Fixes #495

### DIFF
--- a/proposals/lip-0060.md
+++ b/proposals/lip-0060.md
@@ -138,7 +138,7 @@ The block header is processed as follows:
   * The value of `b.header.maxHeightPrevoted` is equal to `b.header.height`.
   * The value of `b.header.impliesMaxPrevotes` is equal to `True`.
   * The value `b.header.maxHeightGenerated`  is equal to 0.
-  * The value of `b.header.aggregateCommit.height` is `b.header.height`.
+  * The value of `b.header.aggregateCommit.height` is 0.
   * The value of `b.header.aggregateCommit.signature` is empty bytes.
   * The value of `b.header.aggregateCommit.aggregationBits` is empty bytes.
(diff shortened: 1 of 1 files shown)


gkoumout (approved):

janhack:
As we reset the Testnet and re-create the new genesis block for Testnet due to other reasons, we do not need to revert the change of b.header.aggregateCommit.height. Therefore, the PR is closed.

